PR #49, opened 2020-03-23
Bugfix/832/fix non vsam non sms existence checking zos data set

blakeinate:
##### SUMMARY

Fixes #832 in Jira. This bug was originally caused because the method that was used to check for data set existence (checking VTOC) does not apply to non-VSAM, non-SMS managed data sets. This was causing issues when a data set meeting these criteria is uncataloged.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- plugins/modules/zos_data_set.py

### DIFF
--- a/plugins/modules/zos_data_set.py
+++ b/plugins/modules/zos_data_set.py
@@ -813,13 +813,14 @@ class DataSetHandler(object):
         Returns:
             bool -- If changes were made.
         """
-        if not self._is_in_vtoc(name, volume):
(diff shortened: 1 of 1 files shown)


blakeinate:
Exception: Credentials 'cicd-key' is of type 'Secret text' where 'com.cloudbees.jenkins.plugins.sshcredentials.SSHUserPrivateKey' was expected

ddimatos:
@blakeinate i did not delete your branch.

PR #73, opened 2020-04-29
Content/81/zos tso command docreview

RadhaV2020:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/zos_fetch.py
+++ b/plugins/modules/zos_fetch.py
@@ -35,7 +35,8 @@ author: "Asif Mahmud (@asifmahmud)"
 options:
   src:
     description:
-      - Name of PDS, PDSE members, VSAM data set, USS file path.
+      - Name of a Unix System Services (USS) file, PS(sequential data set), PDS,
+        PDSE, member of a PDS, PDSE or KSDS(VSAM data set).
       - USS file paths should be absolute paths.
     required: true
     type: str
@@ -84,35 +85,40 @@ options:
   encoding:
     description:
       - Specifies which encodings the fetched data set should be converted from
-        and to. If this parameter is not provided, no encoding conversions will
-        take place.
+        and to. If this parameter is not provided, encoding conversions will
+        not take place.
     required: false
     type: dict
     suboptions:
       from:
         description:
-            - The encoding to be converted from.
+            - The character set of the source I(src).
+            - Supported character sets rely on the target version; the most
+            common character sets are supported.
         required: true
         type: str
       to:
         description:
-            - The encoding to be converted to.
+            - The destination I(dest) character set for the output to be written
+              as.
+            - Supported character sets rely on the target version; the most
+              common character sets are supported.
         required: true
         type: str
 notes:
     - When fetching PDSE and VSAM data sets, temporary storage will be used
       on the remote z/OS system. After the PDSE or VSAM data set is
-      successfully transferred, the temporary data set will be deleted. The size
+      successfully transferred, the temporary storage will be deleted. The size
       of the temporary storage will correspond to the size of PDSE or VSAM
       data set being fetched. If module execution fails, the temporary
-      storage will be cleaned.
-    - To prevent redundancy, additional checksum validation will not be done
-      when fetching PDSE because data integrity checks are done through the
-      transfer methods used. As a result, the module response will not include
-      C(checksum) parameter.
+      storage will be deleted.
+    - To ensure optimal performance, data integrity checks for PDS, PDSE, and
+      members of PDS or PDSE are done through the transfer methods used.
+      As a result, the module response will not include
+      the C(checksum) parameter.
     - All data sets are always assumed to be cataloged. If an uncataloged
       data set needs to be fetched, it should be cataloged first.
-    - Fetching HFS or ZFS is currently not supported.
+    - Fetching HFS or ZFS type data sets is currently not supported.
 seealso:
 - module: fetch
 - module: zos_copy
@@ -185,7 +191,8 @@ is_binary:
     type: bool
     sample: True
 checksum:
-    description: The SHA256 checksum of the fetched file.
+    description: The SHA256 checksum of the fetched file or data set. checksum
+    validation is performed for all USS files and sequential data sets.
     returned: success and src is a non-partitioned data set
     type: str
     sample: 8d320d5f68b048fc97559d771ede68b37a71e8374d1d678d96dcfa2b2da7a64e
@@ -199,10 +206,6 @@ note:
     returned: failure and fail_on_missing=false
     type: str
     sample: The data set USER.PROCLIB does not exist. No data was fetched.
-changed:
-    description: Indicates if any changes were made during the module operation.
-    returned: always
-    type: bool
 msg:
     description: Message returned on failure.
     returned: failure
--- a/plugins/modules/zos_fetch.py
+++ b/plugins/modules/zos_fetch.py
@@ -20,7 +20,11 @@ module: zos_fetch
 version_added: "2.9"
 short_description: Fetch data from z/OS
 description:
-  - This module fetches a file or data set from a remote z/OS system. 
+  - This module fetches a Unix System Services (USS) file,
+    PS(sequential data set), PDS, PDSE, member of a PDS or PDSE, or
+    KSDS(VSAM data set) from a remote z/OS system.
+  - In the case of a PDS or PDSE member, the destination will be the same as
+    the dest parameter. When fetching a member, destination will be a file.
   - When fetching a sequential data set, the destination file name will be the
     same as the data set name.
   - When fetching a PDS or PDSE, the destination will be a directory with the
@@ -60,8 +64,7 @@ options:
       - Override the default behavior of appending hostname/path/to/file to the
         destination. If set to "true", the file or data set will be fetched to
         the destination directory without appending remote hostname to the
-        destination. Refer to the M(fetch) module for a more detailed
-        description of this parameter.
+        destination.
     required: false
     default: "true"
     type: bool
@@ -97,17 +100,17 @@ options:
         required: true
         type: str
 notes:
-    - When fetching PDS(E) and VSAM data sets, temporary storage will be used
-      on the remote z/OS system. After the PDS(E) or VSAM data set is
+    - When fetching PDSE and VSAM data sets, temporary storage will be used
+      on the remote z/OS system. After the PDSE or VSAM data set is
       successfully transferred, the temporary data set will be deleted. The size
-      of the temporary storage will correspond to the size of PDS(E) or VSAM
+      of the temporary storage will correspond to the size of PDSE or VSAM
       data set being fetched. If module execution fails, the temporary
       storage will be cleaned.
     - To prevent redundancy, additional checksum validation will not be done
-      when fetching PDS(E) because data integrity checks are done through the
+      when fetching PDSE because data integrity checks are done through the
       transfer methods used. As a result, the module response will not include
       C(checksum) parameter.
-    - All data sets are always assumed to be cataloged. If an uncataloged 
+    - All data sets are always assumed to be cataloged. If an uncataloged
       data set needs to be fetched, it should be cataloged first.
     - Fetching HFS or ZFS is currently not supported.
 seealso:
--- a/plugins/modules/zos_fetch.py
+++ b/plugins/modules/zos_fetch.py
@@ -26,8 +26,9 @@ description:
     module to copy files from local machine to the remote z/OS system.
   - When fetching a sequential data set, the destination file name will be the
     same as the data set name.
-  - When fetching a PDS/PDS(E), the destination will be a directory with the
-    same name as the PDS/PDS(E).
+  - When fetching a PDS or PDS(E), the destination will be a directory with the
+    same name as the PDS or PDS(E).
+    # why are we using different methods to refer to PDS(E) - elsewhere I see PDSE. It will be helpful if we stick to either one.
   - Files that already exist at dest will be overwritten if they are different
     than the src.
 author: "Asif Mahmud (@asifmahmud)"
@@ -91,18 +92,18 @@ options:
     suboptions:
       from:
         description:
-            - The encoding to be converted from
+            - The encoding to be converted from.
         required: true
         type: str
       to:
         description:
-            - The encoding to be converted to
+            - The encoding to be converted to.
         required: true
         type: str
 notes:
     - When fetching PDS(E) and VSAM data sets, temporary storage will be used
       on the remote z/OS system. After the PDS(E) or VSAM data set is
-      successfully transferred, the temporary data set will deleted. The size
+      successfully transferred, the temporary data set will be deleted. The size
       of the temporary storage will correspond to the size of PDS(E) or VSAM
       data set being fetched. If module execution fails, the temporary
       storage will be cleaned.
@@ -110,7 +111,7 @@ notes:
       when fetching PDS(E) because data integrity checks are done through the
       transfer methods used. As a result, the module response will not include
       C(checksum) parameter.
-    - All data sets are always assumed to be in catalog. If an uncataloged data
+    - All data sets are always assumed to be in the catalog. If an uncataloged data
       set needs to be fetched, it should be cataloged first.
     - Fetching HFS or ZFS is currently not supported.
 seealso:
@@ -139,7 +140,7 @@ EXAMPLES = r"""
     flat: true
     is_binary: true
 
-- name: Fetch a unix file and don't validate its checksum
+- name: Fetch a Unix file and don't validate its checksum
   zos_fetch:
     src: /tmp/somefile
     dest: /tmp/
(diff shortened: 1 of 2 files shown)


plugins/modules/zos_fetch.py
```diff
@@ -26,8 +26,9 @@
     module to copy files from local machine to the remote z/OS system.
   - When fetching a sequential data set, the destination file name will be the
     same as the data set name.
-  - When fetching a PDS/PDS(E), the destination will be a directory with the
-    same name as the PDS/PDS(E).
+  - When fetching a PDS or PDS(E), the destination will be a directory with the
+    same name as the PDS or PDS(E).
+    # why are we using different methods to refer to PDS(E) - elsewhere I see PDSE. It will be helpful if we stick to either one.
```
ddimatos, line 31: It should be PDSE

plugins/modules/zos_fetch.py
```diff
@@ -91,26 +92,26 @@
...
+      successfully transferred, the temporary data set will be deleted. The size
       of the temporary storage will correspond to the size of PDS(E) or VSAM
       data set being fetched. If module execution fails, the temporary
       storage will be cleaned.
     - To prevent redundancy, additional checksum validation will not be done
       when fetching PDS(E) because data integrity checks are done through the
       transfer methods used. As a result, the module response will not include
       C(checksum) parameter.
-    - All data sets are always assumed to be in catalog. If an uncataloged data
+    - All data sets are always assumed to be in the catalog. If an uncataloged data
```
ddimatos, line 114: I am good with either if you see this in the future:
All data sets are always assumed to be in the catalog.
All data sets are always assumed to be cataloged.

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +171,66 @@
...
     sample: SOME.DATA.SET
 dest:
-    description: The destination file path on controlling machine
+    description: The destination file path on the controlling machine.
     returned: success
     type: str
     sample: /tmp/SOME.DATA.SET
 is_binary:
-    description: Indicates which transfer mode was used to fetch the file
+    description: Indicates which transfer mode was used to fetch the file.
```
ddimatos, line 184: @asifmahmud - I believe is_binary is used for both USS and Data sets, so i think its more accurate to truncate this sentence to "Indicates which transfer mode was used to fetch."

@RadhaV2020

plugins/modules/zos_fetch.py
ddimatos: I believe people think of PDS(members) and augment this in their mind to differentiate between PDS and PDSE where PDS(E) maps to members, but should be PDSE.

plugins/modules/zos_fetch.py
```diff
@@ -91,26 +92,26 @@
...
         required: true
         type: str
       to:
         description:
-            - The encoding to be converted to
+            - The encoding to be converted to.
         required: true
         type: str
 notes:
     - When fetching PDS(E) and VSAM data sets, temporary storage will be used
```
ddimatos, line 104: PDS(E) > PDSE

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +171,66 @@
 
 RETURN = r"""
 file:
-    description: The source file path on remote machine
+    description: The source file path on the remote machine.
```
ddimatos, line 174: I believe this should read:
"The source file path  or data set on the remote machine."

`src` according to the options can be a UNIX file or data set.

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +171,66 @@
...
     sample: True
 checksum:
-    description: The SHA256 checksum of the fetched file
+    description: The SHA256 checksum of the fetched file.
     returned: success and src is a non-partitioned data set
     type: str
     sample: 8d320d5f68b048fc97559d771ede68b37a71e8374d1d678d96dcfa2b2da7a64e
 data_set_type:
-    description: Indicates the fetched file's data set type
+    description: Indicates the fetched file's data set type.
```
ddimatos, line 194: @asifmahmud do we want to say " file's data set type" , trying to see were 'file's' fit into a data set , maybe just remove 'files' so it reads "Indicates the fetched data set type."?

ddimatos:
@asifmahmud - @RadhaV2020 will take a pass at my comments, hope you can shed some insight to them if you agree, disagree.

plugins/modules/zos_fetch.py
RadhaV2020: Replaced the description to Indicates the fetched data set type

RadhaV2020:
Updated the PR after incorporating Demetri's review comments

blakeinate:
 ERROR: Found 2 pep8 issue(s) which need to be resolved:
ERROR: plugins/modules/zos_fetch.py:23:70: W291: trailing whitespace
ERROR: plugins/modules/zos_fetch.py:110:74: W291: trailing whitespace

plugins/modules/zos_fetch.py
```diff
@@ -20,21 +20,18 @@
 version_added: "2.9"
 short_description: Fetch data from z/OS
 description:
-  - This module copies a file or data set from a remote z/OS system to the
-    local machine. Use the
-    L(fetch,https://docs.ansible.com/ansible/latest/modules/fetch_module.html)
-    module to copy files from local machine to the remote z/OS system.
+  - This module fetches a file or data set from a remote z/OS system.
```
ddimatos, line 23: I think we should be a bit more verbose and give credit to the details:

This module fetches a Unix System Services (USS) file,
PS(sequential data set), PDS, PDSE, member of a PDS/PDSE or
KSDS(VSAM data set).

plugins/modules/zos_fetch.py
```diff
@@ -20,21 +20,18 @@
...
-    L(fetch,https://docs.ansible.com/ansible/latest/modules/fetch_module.html)
-    module to copy files from local machine to the remote z/OS system.
+  - This module fetches a file or data set from a remote z/OS system. 
   - When fetching a sequential data set, the destination file name will be the
     same as the data set name.
-  - When fetching a PDS/PDS(E), the destination will be a directory with the
-    same name as the PDS/PDS(E).
+  - When fetching a PDS or PDSE, the destination will be a directory with the
+    same name as the PDS or PDSE.
   - Files that already exist at dest will be overwritten if they are different
```
ddimatos, line 28: When fetching files that present on the destination I(dest), they will be overwritten if they are different than the source I(src).

ddimatos: @asifmahmud - I rewrote the above a bit, but what is the behavior if the source and destination are the same, it might be further down in this doc but noting it now.

plugins/modules/zos_fetch.py
```diff
@@ -20,21 +20,18 @@
...
     than the src.
 author: "Asif Mahmud (@asifmahmud)"
 options:
   src:
     description:
-      - Name of PDS, PDS(E) members, VSAM data set, USS file path.
+      - Name of PDS, PDSE members, VSAM data set, USS file path.
```
ddimatos, line 34: Name of a Unix System Services (USS) file,
PS(sequential data set), PDS, PDSE, member of a PDS/PDSE or
KSDS(VSAM data set).

ddimatos: @asifmahmud  do you agree with the above change?

plugins/modules/zos_fetch.py
```diff
@@ -91,27 +88,27 @@
     suboptions:
```
ddimatos, line 88: I realize we were not consistent with our module options for encode between encoding and fetch modules, ill open a jira to normalize the encodng options to match the fetch module style (i think that is nicer). This is more for an FYI to @asifmahmud

ddimatos: @RadhaV2020 
In the encode module we have more details for these options, under the covers they share some similar code , although the options are differently named, i think you can pull from the encoding module and update this one, note this module fetch does not have defaults. 
```
options:
  from_encoding:
    description:
      - The charset of the source I(src).
      - Supported character sets rely on the target version; the most common
        charsets are supported.
    required: false
    type: str
    default: IBM-1047
  to_encoding:
    description:
      - The destination I(dest) charset for the output to be written as.
      - Supported character sets rely on the target version; the most common
        character sets are supported.
    required: false
    type: str
    default: ISO8859-1
```

plugins/modules/zos_fetch.py
```diff
@@ -91,27 +88,27 @@
...
-            - The encoding to be converted to
+            - The encoding to be converted to.
         required: true
         type: str
 notes:
     - When fetching PDS(E) and VSAM data sets, temporary storage will be used
       on the remote z/OS system. After the PDS(E) or VSAM data set is
-      successfully transferred, the temporary data set will deleted. The size
+      successfully transferred, the temporary data set will be deleted. The size
       of the temporary storage will correspond to the size of PDS(E) or VSAM
```
ddimatos, line 103: ...to the size of the the PS(sequential data set), PDS, PDSE, member of a PDS/PDSE or
KSDS(VSAM data set).

@RadhaV2020 - I think there is other places PDS(E) is used not showing up in the diff, I think you have a good idea how to work around that....

plugins/modules/zos_fetch.py
```diff
@@ -91,27 +88,27 @@
...
       data set being fetched. If module execution fails, the temporary
       storage will be cleaned.
```
ddimatos, line 105: "cleaned", instead use removed.

plugins/modules/zos_fetch.py
```diff
@@ -91,27 +88,27 @@
...
     - To prevent redundancy, additional checksum validation will not be done
       when fetching PDS(E) because data integrity checks are done through the
```
ddimatos, line 107: when fetching a PS(sequential data set), PDS, PDSE, member of a PDS/PDSE or
KSDS(VSAM data set). 

@asifmahmud  ^^^ is the above true.

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos Yes, I agree.

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos The above is true only when fetching partitioned data sets (PDS/PDSE) and VSAM.

plugins/modules/zos_fetch.py
```diff
@@ -91,27 +88,27 @@
...
       transfer methods used. As a result, the module response will not include
       C(checksum) parameter.
-    - All data sets are always assumed to be in catalog. If an uncataloged data
-      set needs to be fetched, it should be cataloged first.
+    - All data sets are always assumed to be cataloged. If an uncataloged 
+      data set needs to be fetched, it should be cataloged first.
     - Fetching HFS or ZFS is currently not supported.
```
ddimatos, line 112: - Fetching HFS and ZFS type data sets is currently not supported.

plugins/modules/zos_fetch.py
asifmahmud: One thing to note here also, is that fetch has a default encoding option, while we decided not have a default for `zos_copy` for example. We might want to have a consistent behavior across these modules

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +167,66 @@
 
 RETURN = r"""
 file:
-    description: The source file path on remote machine
+    description: The source file path or data set on the remote machine.
```
ddimatos, line 170: @asifmahmud - is this returning the value used as `src` , ie a USS file or Data Set ?

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos Actually, the behavior is that `dest` will be overwritten regardless, consistent with Ansible's fetch behavior

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +167,66 @@
...
     sample: /tmp/SOME.DATA.SET
 is_binary:
-    description: Indicates which transfer mode was used to fetch the file
+    description: Indicates the transfer mode that was used to fetch.
     returned: success
     type: bool
     sample: True
 checksum:
-    description: The SHA256 checksum of the fetched file
+    description: The SHA256 checksum of the fetched file.
```
ddimatos, line 185: I am not really clear on when this is returned, i see it says file , i see if its a non-partitioned data set, i also see that the checksum description says:
```  validate_checksum:
    description:
      - Verify that the source and destination checksums match after the files
        are fetched.``` but it says files does this work data sets ? Which ones? 

I also see
``` - To prevent redundancy, additional checksum validation will not be done
      when fetching PDS(E) because data integrity checks are done through the
      transfer methods used. As a result, the module response will not include
      C(checksum) parameter.``` how about PDSE members, PDS, etc, i think we need some clarity around this.

File: plugins/modules/zos_fetch.py [hunk 17]
```diff
@@ -170,66 +167,66 @@
...
     sample: PDSE
 note:
-    description: Notice of module failure when C(fail_on_missing) is false
+    description: Notice of module failure when C(fail_on_missing) is false.
     returned: failure and fail_on_missing=false
     type: str
     sample: The data set USER.PROCLIB does not exist. No data was fetched.
 changed:
-    description: Indicates if any changes were made during the module operation
+    description: Indicates if any changes were made during the module operation.
```
ddimatos, line 200: @asifmahmud 
I think we can remove changed and only document change when we differ outside the norm, ie something like command execution where we really don't know if commands are successful but only that they are successfully submitted there for changed would always be true, that is a good example. 

In this modules description it seems fetching would never make a change and conforms to the same theme of how Ansible documents it :https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#id11

plugins/modules/zos_fetch.py
```diff
@@ -170,66 +167,66 @@
...
     returned: always
     type: bool
 msg:
-    description: Message returned on failure
+    description: Message returned on failure.
```
ddimatos, line 204: This is just and FYI @asifmahmud , but i think you did a good job giving samples for msg, stdout, stderr outside the standard ansible doc that it adds value enough to leave these in the ansible doc.

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos - Yes

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos @RadhaV2020 - I should have been clearer about this. Just to clear things up:
1. `checksum` is returned when fetching USS files or sequential data sets.
2. `checksum` is not calculated for PDS/PDSE or PDS/PDSE members.

plugins/modules/zos_fetch.py
asifmahmud: @ddimatos - Yeah, I think that would be better. We don't actually have to document this parameter, since it gets reported to the user anyway

plugins/modules/zos_fetch.py
ddimatos: @asifmahmud - i took note of your comment, plan to open a JIRA up so we don't let it slip, we can decide what the default should be and then ensure zos_copy adheres to it and if need be change the other modules.

ddimatos: @RadhaV2020  - maybe something like:
from:
   description: 
      - The charset of the source I(src).
      - Supported character sets rely on the target version; the most common
        charsets are supported.
  to:
        description:
      - The destination I(dest) charset for the output to be written as.
      - Supported character sets rely on the target version; the most common
        character sets are supported.

I thought it would be reworded a bit differently but as I started to look closer , its really the same for both modules, I don't hink @asifmahmud  would disagree, but he is free to comment. 

The only issue i have with all of this is "the most common character sets are supported."  and i don't have a problem with how its written, this is what i wanted, what i have an issue with is there is over 600 supported encodings, i was thinking to output the list of encodings from `iconv` into an RST file, host it in our web-docs and then link it from the module doc to the web-doc, not sure i can get to it in this beta release, maybe ill get the doc doc and we can link it for our Hub release in the near future.

blakeinate:
 ERROR: Command "ansible-doc -t module ibm.ibm_zos_core.zos_data_set ibm.ibm_zos_core.zos_encode ibm.ibm_zos_core.zos_fetch ibm.ibm_zos_core.zos_job_output ibm.ibm_zos_core.zos_job_query ibm.ibm_zos_core.zos_job_submit ibm.ibm_zos_core.zos_operator ibm.ibm_zos_core.zos_operator_action_query ibm.ibm_zos_core.zos_tso_command" returned exit status 1.
ERROR: Found 13 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/zos_fetch.py:0:0: doc-default-does-not-match-spec: Argument 'fail_on_missing' in argument_spec defines default as (True) but documentation defines default as (False)
ERROR: plugins/modules/zos_fetch.py:0:0: doc-default-does-not-match-spec: Argument 'flat' in argument_spec defines default as (True) but documentation defines default as (False)
ERROR: plugins/modules/zos_fetch.py:0:0: doc-default-does-not-match-spec: Argument 'validate_checksum' in argument_spec defines default as (True) but documentation defines default as (False)
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'dest' in argument_spec defines type as 'path' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'encoding' in argument_spec defines type as 'dict' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'fail_on_missing' in argument_spec defines type as 'bool' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'flat' in argument_spec defines type as 'bool' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'is_binary' in argument_spec defines type as 'bool' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'src' in argument_spec defines type as 'str' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'use_qualifier' in argument_spec defines type as 'bool' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:0:0: parameter-type-not-in-doc: Argument 'validate_checksum' in argument_spec defines type as 'bool' but documentation doesn't define type
ERROR: plugins/modules/zos_fetch.py:98:9: documentation-syntax-error: DOCUMENTATION is not valid YAML
ERROR: plugins/modules/zos_fetch.py:196:5: return-syntax-error: RETURN is not valid YAML
ERROR: Found 2 yamllint issue(s) which need to be resolved:
ERROR: plugins/modules/zos_fetch.py:98:9: error: DOCUMENTATION: syntax error: could not find expected ':' (syntax)
ERROR: plugins/modules/zos_fetch.py:196:5: error: RETURN: syntax error: could not find expected ':' (syntax)
ERROR: The 3 sanity test(s) listed below (out of 41) failed. See error output above for details.

PR #107, opened 2020-05-21
Update/1979/default encoding value

asifmahmud:
##### SUMMARY

+ Change zos_job_submit default local encoding from UTF-8 to ISO8859-1
+ Add an option to override remote encoding (currently hard-coded to IBM-1047)
+ Improve doc for zos_job_submit for the encoding parameter. Better explanation of the default encoding behavior.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

zos_job_submit

### DIFF
--- a/plugins/modules/zos_job_submit.py
+++ b/plugins/modules/zos_job_submit.py
@@ -111,6 +111,9 @@ options:
         required: false
         type: str
         default: IBM-1047
+notes:
+  - For supported character sets used to encode data, refer to
+    U(https://ansible-collections.github.io/ibm_zos_core/supplementary.html#encode)
 """
 
 RETURN = r"""
@@ -435,10 +438,6 @@ jobs:
               "subsystem": "STL1"
           }
      ]
-changed:
-  description: Indicates if any changes were made during module operation.
-  type: bool
-  returned: success
 message:
   description: The output message that the sample module generates.
   returned: success
@@ -647,14 +646,21 @@ def run_module():
         location=dict(
             arg_type="str", default="DATA_SET", choices=["DATA_SET", "USS", "LOCAL"],
         ),
+        from_encoding=dict(arg_type="encoding", default=DEFAULT_ASCII_CHARSET),
+        to_encoding=dict(arg_type="encoding", default=DEFAULT_EBCDIC_CHARSET),
         volume=dict(arg_type="volume", required=False),
         return_output=dict(arg_type="bool", default=True),
         wait_time_s=dict(arg_type="int", required=False, default=60),
         max_rc=dict(arg_type="int", required=False),
-        temp_file=dict(arg_type="path", required=False)
     )
 
     result = dict(changed=False)
+    module.params.update(
+        dict(
+            from_encoding=module.params.get('encoding').get('from'),
+            to_encoding=module.params.get('encoding').get('to')
+        )
+    )
     try:
         parser = BetterArgParser(arg_defs)
         parsed_args = parser.parse_args(module.params)
@@ -668,7 +674,7 @@ def run_module():
     return_output = parsed_args.get("return_output")
     wait_time_s = parsed_args.get("wait_time_s")
     max_rc = parsed_args.get("max_rc")
-    # get temporary file names for copied files
+    # get temporary file names for copied filesß
     temp_file = parsed_args.get("temp_file")
     if temp_file:
         temp_file_2 = NamedTemporaryFile(delete=True)
--- a/plugins/modules/zos_job_submit.py
+++ b/plugins/modules/zos_job_submit.py
@@ -87,18 +87,18 @@ options:
         Ignored for USS and LOCAL.
   encoding:
     required: false
-    default: UTF-8
+    default: ISO8859-1
     type: str
     choices:
       - UTF-8
       - ASCII
-      - ISO-8859-1
+      - ISO8859-1
       - EBCDIC
       - IBM-037
       - IBM-1047
     description:
       - The encoding of the local JCL file on the ansible control node.
-      - If it is UTF-8, ASCII, ISO-8859-1, the file will be converted to EBCDIC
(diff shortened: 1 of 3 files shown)


blakeinate:
 ERROR: Found 1 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/zos_job_submit.py:0:0: doc-default-does-not-match-spec: Argument 'encoding' in argument_spec defines default as ({'from': 'ISO8859-1', 'to': 'IBM-1047'}) but documentation defines default as (None)

plugins/modules/zos_job_submit.py
```diff
@@ -628,25 +633,33 @@ def run_module():
...
+    if encoding.get('to') is None:
+        encoding['to'] = DEFAULT_EBCDIC_CHARSET
 
     arg_defs = dict(
         src=dict(arg_type="data_set_or_path", required=True),
         wait=dict(arg_type="bool", required=False),
         location=dict(
             arg_type="str", default="DATA_SET", choices=["DATA_SET", "USS", "LOCAL"],
         ),
-        encoding=dict(arg_type="encoding", default="UTF-8"),
```
blakeinate, line 638: I understand why the old arg type would not work due to your new encoding parm structure, but would it be possible to add back in some sort of string checking? I know `quote()` should make things "safe" but having some sort of input sanitization/validation would probably be good

blakeinate: For the BetterArgParser parsing, something like this could work (little hacky though):
```python
    arg_defs = dict(
        src=dict(arg_type="data_set_or_path", required=True),
        wait=dict(arg_type="bool", required=False),
        location=dict(
            arg_type="str", default="DATA_SET", choices=["DATA_SET", "USS", "LOCAL"],
        ),
        from_encoding=dict(arg_type="encoding", default=DEFAULT_ASCII_CHARSET),
        to_encoding=dict(arg_type="encoding", default=DEFAULT_EBCDIC_CHARSET),
        volume=dict(arg_type="volume", required=False),
        return_output=dict(arg_type="bool", default=True),
        wait_time_s=dict(arg_type="int", required=False, default=60),
        max_rc=dict(arg_type="int", required=False),
    )
   try:
        parser = BetterArgParser(arg_defs)
        parsed_args = parser.parse_args(module.params.update({"from_encoding": encoding.get('from'), "to_encoding": encoding.get("to"))
    except ValueError as err:
        module.fail_json(msg=str(err), **result)
```

plugins/modules/zos_job_submit.py
```diff
@@ -686,35 +699,20 @@ def run_module():
...
-                    jobId = submit_uss_jcl(temp_file_2.name, module)
-                else:
-                    module.fail_json(
-                        msg="The Local file encoding conversion failed. Please check the source file."
-                        + stderr,
-                        **result
-                    )
+            from_encoding = encoding.get('from')
+            to_encoding = encoding.get('to')
+            (conv_rc, stdout, stderr) = module.run_command(
```
blakeinate, line 704: does there need to be a check here to ensure `from_encoding` and `to_encoding` are different before running `iconv`? Not sure if it would cause any issues or not

plugins/modules/zos_job_submit.py
```diff
@@ -727,9 +725,9 @@ def run_module():
         )
 
     result["job_id"] = jobId
+    duration = 0
```
blakeinate, line 728: I actually have a PR open for this right now, will go ahead and close mine :)

plugins/modules/zos_job_submit.py
ddimatos: @blakeinate are you suggesting we sanitize all the charsets? Not sure i followed your suggestion, if so, it is something i thought we could do problem is if `iconv` changes the chartsets it supports we have to version track all these changes. I ended up adding this doc to support the charsets and added to the [encoding module doc here](https://github.com/ansible-collections/ibm_zos_core/blob/v1.1.0-beta1/plugins/modules/zos_encode.py#L108) and here is the doc site [link](https://ansible-collections.github.io/ibm_zos_core/supplementary.html#encode)

plugins/modules/zos_job_submit.py
ddimatos: I am pretty sure no error happens, but not sure if `iconv` does a NOOP or actually consumes MIPS, so better to have check and do nothing if they are the same. 

```
# iconv -f IBM-1047 -t IBM-1047 hi.txt
Hello World!
````

plugins/modules/zos_job_submit.py
asifmahmud: @blakeinate Yeah, that is probably a good idea. My initial intent was that if the provided charsets are not valid, `iconv` would error out, which then can be relayed back to the user. But I agree that it can probably be handled earlier in the control flow.

blakeinate: @ddimatos what was being done previously was a basic character check, essentially a regex containing all characters and min-max lengths of any existing encoding I could find, this ensures no characters that might be used for attempted shell injection could get through

plugins/modules/zos_job_submit.py
asifmahmud: @blakeinate @ddimatos Yes, if both charsets are the same, no error happens. Not sure if additional MIPS are consumed though.

asifmahmud:
@ddimatos Updated the doc as you requested and added encoding validation.

plugins/modules/zos_job_submit.py
blakeinate: I tried finding more information on if the GNU iconv implementation would check for matching character sets, was not able to find much. Tried testing on USS using `time iconv` with a large file, but in both cases (matching charsets and different charsets) the response was about the same. Wasn't much help :)

asifmahmud: To be fair, iconv is quite fast even for extremely large files. Hard to gauge it that way

PR #127, opened 2020-06-29
Add system info to job query

### LINKED ISSUE #125: Add system to job query response
##### SUMMARY
When querying jobs with `zos_job_query`, the following four properties are returned per job: `job_id`, `job_name`, `owner`, `ret_code`. This Feature Request asks that a fifth property `system` be returned, containing the system the job ran on / is running on.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`zos_job_query`

##### ADDITIONAL INFORMATION
Having the system returned would be valuable for numerous reasons. One such reason could be taking action against a job returned by `zos_job_query`. A cancel or modify action would require a command routed to the system the job is running on. Currently, there is not a good way to know that system.

asifmahmud:
##### SUMMARY

Add subsystem information to job query

Fixes #125

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_job_query

### DIFF
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -328,7 +328,7 @@ do ix=1 to isfrows
     Say '"'||'job_id'||'":"'||value('JOBID'||"."||ix)||'",'
     Say '"'||'job_name'||'":"'||value('JNAME'||"."||ix)||'",'
     Say '"'||'subsystem'||'":"'||value('ESYSID'||"."||ix)||'",'
-    Say '"'||'sysname'||'":"'||value('SYSNAME'||"."||ix)||'",'
+    Say '"'||'system'||'":"'||value('SYSNAME'||"."||ix)||'",'
     Say '"'||'owner'||'":"'||value('OWNERID'||"."||ix)||'",'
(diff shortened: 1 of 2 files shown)


plugins/module_utils/job.py
```diff
@@ -328,6 +328,7 @@ def _get_job_status_str(job_id="*", owner="*", job_name="*"):
     Say '"'||'job_id'||'":"'||value('JOBID'||"."||ix)||'",'
     Say '"'||'job_name'||'":"'||value('JNAME'||"."||ix)||'",'
     Say '"'||'subsystem'||'":"'||value('ESYSID'||"."||ix)||'",'
+    Say '"'||'sysname'||'":"'||value('SYSNAME'||"."||ix)||'",'
```
blakeinate, line 331: Would it be possible to switch `sysname` in output to `system`? Some other modules accept the output JSON without modification, so it may be worth outputting `system` for consistency between module responses

asifmahmud: Yes, that would be better for consistency.

PR #142, opened 2020-07-22
Feature/dependency discovery

blakeinate:
##### SUMMARY

This PR adds the playbook *host-setup.yaml* and its dependencies. *host-vars.yaml* will attempt to locate Ansible dependencies on a z/OS system and use them to build a valid inventory and host_vars file for the host. On a native LPAR, the playbook takes about 1-5 minutes to run, or about 5-15 minutes on a z/VM depending on max search depth.

Features:
- Attempts to select the best Python found on the system. For Python, IBM Python is the first choice, then followed by Anaconda installation then base Rocket Python installation.
- Attempts to find the most recent ZOAU build available on the system, and cross checks with minimum required version for the collection.
- Attempts to run modules with the discovered dependencies to ensure they are in working order.
- Will test `ping` module with each discovered python install until `ping` succeeds or options are exhausted.
- Will test each ZOAU install using `zos_data_set` module and the chosen python interpreter until `zos_data_set` succeeds or options are exhausted. 
- Makes backups of existing inventory files and host_vars files if there are name conflicts
- Is able to run without using any inventory file or environment variable configuration.
- Adjustable search depth for speed optimizations. Default max search depth is 15. 
- Outputs all discovered python and ZOAU installs in case user wants to build their own config using discovered installations.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

To run,
`ansible-playbook -i "<system_name>," -u "<username>" host-setup.yml --ask-pass`
where `<system_name>` is the host name of the z/OS system and `<username>`  is the user name used to connect to the z/OS system.


playbooks/host-setup.yaml
```diff
@@ -0,0 +1,104 @@
+###############################################################################
```
asifmahmud, line 1: This is an extremely nice playbook. However, to a new user of our collection, it may not be apparent what exactly it is doing. I suggest a an extended introduction/documentation block at the beginning of this playbook to give a brief outline of the steps it is performing.

playbooks/tasks/choose-python.yaml
```diff
@@ -0,0 +1,55 @@
+###############################################################################
```
asifmahmud, line 1: Same goes for this playbook as well.

playbooks/host-setup.yaml
blakeinate: Great point @asifmahmud , definitely would be good to have some in-file documentation

blakeinate: I have added documentation to the host-setup playbook, thank you for your feedback

playbooks/tasks/choose-python.yaml
blakeinate: I have added some basic flow documentation to the two task files choose-python.yaml and choose-zoau.yaml. Since these files can not be run directly, I placed the bulk of the documentation in host-setup.yaml

PR #152, opened 2020-08-06
Update file extensions in host-setup.yaml

blakeinate:
##### SUMMARY

This PR fixes a small issue with file extensions in host-setup.yaml, I had updated the file extensions of task files to `.yaml` from `.yml` but forgot to update the paths in host-setup.yaml.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/playbooks/host-setup.yaml
+++ b/playbooks/host-setup.yaml
@@ -97,7 +97,7 @@
 
     - name: Find valid python installation
       include_tasks:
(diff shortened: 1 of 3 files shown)


blakeinate:
Also went ahead and added `/usr/sbin` to the PATH variables generated by the host-setup.yaml playbook so mount/unmount will work without full path

PR #167, opened 2020-08-14
add FQCN to action plugins

blakeinate:
##### SUMMARY

This PR fixes a bug which causes action plugins to fail when collections are referenced using fully qualified collection names instead of playbook level imports. Since playbook level imports do not work in roles, this causes any of our modules that use action plugins to fail when used inside roles.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/action/zos_copy.py
+++ b/plugins/action/zos_copy.py
@@ -1,7 +1,7 @@
 # Copyright (c) IBM Corporation 2019, 2020
 # Apache License, Version 2.0 (see https://opensource.org/licenses/Apache-2.0)
 
-from __future__ import (absolute_import, division, print_function)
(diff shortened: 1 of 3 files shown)


blakeinate:
Sorry, think black formatter took over and replaced all the `'` with `"`, so diffs look intimidating. Only lines changed are lines where `_execute_module()` is used.

IBMAnsibleHelper:
 ERROR: Found 1 pep8 issue(s) which need to be resolved:
ERROR: plugins/action/zos_fetch.py:136:48: E203: whitespace before ':'

ddimatos:
Good catch @asifmahmud and @balkajbaf , nice work!!!

PR #214, opened 2020-11-18
Update template to support complex list with elements

ddimatos:

##### SUMMARY
This was encountered with when we were on-boarding zHMC where they had a module option that spanned a range. I saw nothing technically incorrect in this option spec and found that our Jinja template would error on a default case when it encountered a default as such below.

```
    type: list
    elements: int
    required: false
    default: [0,-1]
```
##### ISSUE TYPE
- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/templates/module.rst.j2
+++ b/docs/templates/module.rst.j2
@@ -49,16 +49,6 @@ Synopsis
 
 {% macro option_generation(opts, level) %}
 {%   for name, spec in opts | dictsort recursive %}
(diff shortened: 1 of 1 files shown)


IBMAnsibleHelper:
Exception: script returned exit code 1

IBMAnsibleHelper:
Exception: script returned exit code 1

asifmahmud (approved):
Looks good to me, thanks @ddimatos for the fix!

PR #221, opened 2020-11-24
Fixes #217: Support for optional type in return parm properties in module template

### LINKED ISSUE #217: Docs: Module template shows omitted type
##### SUMMARY

In the [zhmc collection](https://github.com/zhmcclient/zhmc-ansible-modules), the return values have pretty long lists of resource properties that we are not documenting one by one because they are documented in the Web Services API book of the HMC. In our descriptions of the return values, we have "generic" properties named `{property}` whose description references the section in the WS-API book that describes the properties. For such generic properties it does not make sense of course to call out a specific type.

Our current solution is to omit the `type` field in the return value docstring. Omitting the type field causes the module.rst.j2 template to generate a "type:" text in the output without a value.

We would like to have the "type:" text in the output to be conditional on the presence of the type field in the return value docstring.

I can bring a PR for this forward.

Note: We open this issue here because this repo has the master version of the module.rst.j2 file, across the IBM collections.

##### ISSUE TYPE
- Documentation Bug Report

##### COMPONENT NAME

* docs/templates/module.rst.j2

##### ANSIBLE VERSION

```paste below
N/A
```

##### CONFIGURATION

```paste below
N/A
```

##### OS / ENVIRONMENT

N/A

##### STEPS TO REPRODUCE

* Build documentation of [zhmc collection](https://github.com/zhmcclient/zhmc-ansible-modules) which can be seen at https://zhmcclient.github.io/zhmc-ansible-modules/.
* Look at this return value to see the effect (2 occurrences): https://zhmcclient.github.io/zhmc-ansible-modules/modules/zhmc_adapter.html#return-values

##### EXPECTED RESULTS

We would like to have the "type:" text in the output to be conditional on the presence of the type field in the return value docstring.

##### ACTUAL RESULTS

See above

andy-maier:
This PR addresses issue #217.
For details, see the commit message.

### DIFF
--- a/docs/templates/module.rst.j2
+++ b/docs/templates/module.rst.j2
@@ -151,7 +151,9 @@ See Also
       {% if _returned %}
 {{ "  " * level }}| **returned**: {{ _returned }}
       {% endif %}
+      {% if _type %}
 {{ "  " * level }}| **type**: {{ _type  }}
+      {% endif %}
 
       {%- if _sample -%}
       {% if _type != 'str' and _type != 'int' %}
@@ -184,4 +186,3 @@ Return Values
 
 {{ result_generation(returndocs,1) }}
 {% endif %}
-


ddimatos:
@andy-maier i see the need for this but have some reservations, ill give it more thought and maybe we can discuss further. 

From what i can see you want to point users at PDFs of already existing doc because its redundant to duplicate and maybe its excessive doc and thus want to point users at  a PDF via a custom `{property}` tag and **OMIT** the `type` in the **RETURN** block.

zHMC doc snippet:
image

Here are my concerns:
If in fact the Ansible modules don't have a `type` defined I have a feeling the sanity tests will fail and we will need to add it to the `ignore-*.txt` which would work to pass the tests but leave. You may not know this yet because we did not get that far to discuss Hub, but Ansible generated doc based on all the Ansible doc we include in the modules when it goes to Hub. It does NOT do this YET for Galaxy, but believe sometime in the very near future they will do the same (as I was told a while ago) which is why we invested in our own doc and now its more than doc. 

So...what happens when the `type` is left out for a collection on HUB? I don't know, will the ignore.txt be enough to keep the Hub release from failing or not? Do we think users of Hub will prefer the doc that is Hub over our doc, I don't think so either, I think our doc provides a richer experience. 

I have some suggestions though. 
- Keep they `type` definition in the RETURN block map it to what your API returns which is likely  `dict` if not `str` , then use `raw` this will ensure we don't choke ansible doc generation for Hub and mess with ansible-sanity tests. Then in the description section of the modules return block, speak briefly to the return as you sort of do now but maybe include the link such that "[For links outside of your collection, use R() if available. Otherwise, use U() or L() with full URLs (not relative links).](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#ansible-metadata-block)" 

For example:
```
contains:
                name:
                  description: "Additional L(properties,https://zhmcclient.github.io/zhmc-ansible-modules/bibliography.html#term-HMC-API) of the adapter, as described in the data model of the ‘Adapter’ object in the HMC API book. The property names have hyphens (-) as described in that book."
                  type: raw   
```

- Second idea, and its really a round about way and maybe something for the later releases if you like it.So you have a PDF of the doc, I expect it was generated from DITA, so how about using the tooling (if it exists) to DITA > MD > RST such that then you can leverage that RST inline or maybe just as a hosted version so people don't have to link out to a pdf and download it and such. I know we can go from MD to DITA, just be ways to go the other way (I have some contacts if you are interested).

Last comment , i tried to download the PDF and had multiple issues, i signed in as well still failed to access the pdf. 
image
image

Here is no sound overview of the doc in Automation Hub (so you get an idea of the doc they generate):
[automation-hub-doc-overview.m4v.zip](https://github.com/ansible-collections/ibm_zos_core/files/5625692/automation-hub-doc-overview.m4v.zip)

ddimatos:
@andy-maier - looks like you closed the parent [issue](https://github.com/zhmcclient/zhmc-ansible-modules/issues/275) thus have a work around for this and no need to disable 'type' , thinking you chose type:raw. Can we consider this completed?

PR #223, opened 2020-11-25
Fixes #222: Fixed return value formatting issues in module template

### LINKED ISSUE #222: Docs: Module template has formatting issues in return values
##### SUMMARY

The module.rst.j2 template has several formatting issues in how it generates return values:

* If the description is a list with multiple items, only the first item is shown and the others are just dropped.
* The description is not rst_ify'ed.
* It does not indent nested properties correctly, causing for example the description to be rendered in a browser with one indentation level more sometimes (see the adapter.ports property in the example). When looking at the generated rst, property names are also not always indented by the desired two blanks compared to the parent level, but the browser rendering seems to recover that in the cases I have seen.
* The 'type:' item is indented more than the other attributes of a property.
* For samples that are not str or int, the code_block statement and the following sample data item are not indented variably by level, but by a fixed number of blanks in the generated rst. This causes them to be rendered inconsistent as soon as the nesting level of the properties is more than 4 or so (but I did not try it out).
* The top level of return properties are already indented one level. Not a real problem, but inconsistent with how it is done for input parameters, and when looking at the generates rst file with a browser, grey vertical bars are shown due to that.
* There are many trailing blanks. Not a real problem, but somehow inconvenient in git with editors that strip them.

I can provide a PR that fixes that. Also, the zosmf version of the template did not have these issues. That was the version we initially used for the zhmc collection, but if the zos_core version is considered the master version, we'd like to use that one now, and get it fixed.

##### ISSUE TYPE
- Documentation Bug Report

##### COMPONENT NAME

* docs/templates/module.rst.j2

##### ANSIBLE VERSION

N/A

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### STEPS TO REPRODUCE

Have a return value description with at least three nested levels of properties, multiple items in the description, and formatting strings like C(xxx) in the description.

For example (from the zhmc_adapter module):

```
RETURN = """
adapter:
  description:
    - "For C(state=absent), an empty dictionary."
    - "For C(state=set|present|facts), the adapter and its ports."
  returned: success
  type: dict
  contains:
    name:
      description: "Adapter name"
      type: str
    "{property}":
      description: "Additional properties of the adapter, as described in the
        data model of the 'Adapter' object in the :term:`HMC API` book.
        The property names have hyphens (-) as described in that book."
    ports:
      description: "Artificial property for the ports of the adapter."
      type: list
      elements: dict
      contains:
        name:
          description: "Port name"
          type: str
        "{property}":
          description: "Additional properties of the port, as described in the
            data model of the 'Network Port' or 'Storage Port' element object
            of the 'Adapter' object in the :term:`HMC API` book.
            The property names have hyphens (-) as described in that book."
```

Generated rst output without the fix:

```
       adapter
        | For C(state=absent), an empty dictionary.
      
        | **returned**: success
              | **type**: dict
      
        name
          | Adapter name
      
                | **type**: str
      
        {property}
          | Additional properties of the adapter, as described in the data model of the 'Adapter' object in the :term:`HMC API` book. The property names have hyphens (-) as described in that book.
      
        ports
          | Artificial property for the ports of the adapter.
      
                | **type**: list
      
         name
            | Port name
      
                  | **type**: str
      
         {property}
            | Additional properties of the port, as described in the data model of the 'Network Port' or 'Storage Port' element object of the 'Adapter' object in the :term:`HMC API` book. The property names have hyphens (-) as described in that book.

                  | **type**: 
```

Intended rst output with the fix:

```
adapter
  For ``state=absent``, an empty dictionary.

  For ``state=set|present|facts``, the adapter and its ports.

  | **returned**: success
  | **type**: dict

  name
    Adapter name

    | **type**: str

  {property}
    Additional properties of the adapter, as described in the data model of the 'Adapter' object in the :term:`HMC API` book. The property names have hyphens (-) as described in that book.

    | **type**: 

  ports
    Artificial property for the ports of the adapter.

    | **type**: list

    name
      Port name

      | **type**: str

    {property}
      Additional properties of the port, as described in the data model of the 'Network Port' or 'Storage Port' element object of the 'Adapter' object in the :term:`HMC API` book. The property names have hyphens (-) as described in that book.

      | **type**: 
```

##### EXPECTED RESULTS

See above

##### ACTUAL RESULTS

See above

andy-maier:
This PR addresses all issues described in #222.
See the commit message for details.

### DIFF
--- a/docs/templates/module.rst.j2
+++ b/docs/templates/module.rst.j2
@@ -55,7 +55,6 @@ Synopsis
   {{ "  " * level }}{{ para | rst_ify }}
 
 {%     endfor %}
-
   {{ "  " * level }}| **required**: {{ spec.required | default("False") }}
   {{ "  " * level }}| **type**: {{ spec.type | default("str") }}
 {%     if spec.elements %}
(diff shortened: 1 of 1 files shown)


ddimatos (approved):
@andy-maier these are good changes, thanks for helping improve the template, it saved me some time and I appreciate that.

PR #230, opened 2020-12-02
Add support to template to skip options that begin with an underscore

ddimatos:

##### SUMMARY
This update allows for documentation on options to be skipped such that no documentation is generated for a module. This can come in handy in cases where we want to avoid displaying an option in our HTML docs. Although, this does nothing in terms of ansible_doc as the option will still show up with `ansible-doc` but we predict that most if not all of our users are using our online docs and not the command line. 

##### ISSUE TYPE
- Docs Pull Request


asifmahmud (approved):
Great idea! Approved

PR #247, opened 2021-02-03
Enhancement/3729

richp405:
##### SUMMARY
Rexx script to add delay interval and (WAIT) for operator commands
Addresses Jira-3729 and reviewer comments

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_operator (changed to temporary rexx script)

```

### DIFF
--- a/plugins/modules/zos_operator.py
+++ b/plugins/modules/zos_operator.py
@@ -191,8 +191,6 @@ def parse_params(params):
 
 
 def run_operator_command(params):
-    ## remove reset sequence... the value resets when the script ends.
-
     # Usage: (rexfile) delay reset command [parameters] [-v] [-d] [-s]
     #       -v: print out verbose security information
     #       -d: print out debug messages
--- a/plugins/modules/zos_operator.py
+++ b/plugins/modules/zos_operator.py
@@ -53,12 +53,6 @@ options:
     type: int
     required: false
     default: 0
-  reset:
-    description:
-      - Reset maximum time in seconds after action is called.
-    type: int
-    required: false
-    default: 0
 """
 
 EXAMPLES = r"""
@@ -90,12 +84,6 @@ EXAMPLES = r"""
   zos_operator:
     cmd: 'd u,all'
     delay: 5
-
-- name: Execute operator command to show jobs. Wait for 7 seconds, then reset delay to 2 seconds
-  zos_operator:
-    cmd: 'd u,all'
-    delay: 7
-    reset: 2
 """
 
 RETURN = r"""
@@ -116,7 +104,19 @@ content:
           "MV2C      2020039  04:29:57.59             IEE457I 04.29.57 UNIT STATUS 948  ",
           "         UNIT TYPE STATUS        VOLSER     VOLSTATE      SS                 ",
           "          0100 3277 OFFLINE                                 0                ",
-          "          0101 3277 OFFLINE                                 0                "
+          "          0101 3277 OFFLINE                                 0                ",
+          "---- if verbose security is true, additional output will resemble:",
+          "ISF050I USER=OMVSADM GROUP= PROC=REXX TERMINAL=09A3233B",
+          "ISF051I SAF Access allowed SAFRC=0 ACCESS=READ CLASS=SDSF RESOURCE=GROUP.ISFSPROG.SDSF",
+          "ISF051I SAF Access allowed SAFRC=0 ACCESS=READ CLASS=SDSF RESOURCE=ISFCMD.FILTER.PREFIX",
+          "ISF055I ACTION=D Access allowed USERLEVEL=7 REQLEVEL=1",
+          "ISF051I SAF Access allowed SAFRC=0 ACCESS=READ CLASS=SDSF RESOURCE=ISFCMD.ODSP.ULOG.JES2",
+          "--- if security is true, output on will resemble:",
+          "ISF147I REXX variable ISFTIMEOUT fetched, return code 00000001 value is ''.",
+          "ISF754I Command 'SET DELAY 5' generated from associated variable ISFDELAY.",
+          "ISF769I System command issued, command text: D U,ALL -S.",
+          "ISF146I REXX variable ISFDIAG set, return code 00000001 value is '00000000 00000000 00000000 00000000 00000000'.",
+          "ISF766I Request completed, status: COMMAND ISSUED."
         ]
 changed:
     description:
@@ -155,7 +155,6 @@ def run_module():
         debug=dict(type="bool", required=False, default=False),
         security=dict(type="bool", required=False, default=False),
         delay=dict(type="int", required=False, default=0),
-        reset=dict(type="int", required=False, default=0),
     )
 
     result = dict(changed=False)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: plugins/modules/zos_operator.py [hunk 1]
```diff
@@ -130,38 +170,161 @@ def run_module():
         rc_message = run_operator_command(new_params)
         result["rc"] = rc_message.get("rc")
         result["content"] = rc_message.get("message").split("\n")
+        if result["rc"] == 0:
+            if len(result["content"]) > 3:
```
ddimatos, line 174: @richp405 I don't know the answer and positional values make me nervous so my question is, what is special about length 3? Are we certain index 2 will always be populated (see line 175). Just asking for clarity, this is a good place for a comment for others to understand who might work on the code.

plugins/modules/zos_operator.py
```diff
@@ -32,16 +32,37 @@
...
+        operator commands.
+      - Return security trace messages that help you understand and diagnose the
+        execution of the operator commands
+      - Return trace instructions displaying how the the command's operation is
+        read, evaluated and executed.
     type: bool
     required: false
     default: false
-  debug:
+  wait_time_s:
```
ddimatos, line 44: @richp405 this is not a show stopper, maybe a max check here of 2147483647 or something, what is going to happen if someone enters 21474836471234 beyond the 2^32?

ddimatos (commented):
Hey @richp405 nice job, looks really good, glad we took the extra time to make it right and even better. I had two minor comments about a max int for time and one for a index check of 3, otherwise really nothing holding this back. 

Line 122 is longer than 80 chars but that is not your doing, its just to hard to break that sample into two lines and have it still be meaningful; given sanity tests did not flag it, just be fine because its quoted.

plugins/modules/zos_operator.py
ddimatos: @blakeinate thanks for reminding in python 3 int's are unbounded :) , so @richp405 we can consider this comment moot.

plugins/modules/zos_operator.py
ddimatos: @blakeinate thanks for reminding in python 3 int's are unbounded :) , so @richp405 we can consider this comment moot.

plugins/modules/zos_operator.py
```diff
@@ -130,38 +170,161 @@ def run_module():
...
+
+    delete_on_close = True
+    tmp_file = NamedTemporaryFile(delete=delete_on_close)
+    with open(tmp_file.name, "w") as f:
+        f.write(script)
+    chmod(tmp_file.name, S_IEXEC | S_IREAD | S_IWRITE)
+
+    rc, stdout, stderr = module.run_command(tmp_file.name + fulline)
+
+    message = stdout + stderr + "\n" + "Ran " + fulline + "\n"
```
blakeinate, line 323: I think you have an edge case bug here

In the event that stdout and stderr are both empty strings, the `split("\n")` in `run_module()` will separate the message string as follows
```
[ "", "Ran " + fulltime, ""]
```
Similarly, if stdout and stderr are both strings but do not contain a newline character:
```
[stdout + stderr, "Ran " + fulltime, ""]
```
In `run_module()`, you are expecting the response line to always be at index 2, which in these cases would be an empty string, causing issues in the remainder of the module. 

In addition, if your stdout and stderr strings contain multiple newlines, index 2 would contain part of the content of stdout or stderr, not the response line you are expecting.

My recommendation:

- Add additional keys to your response in `run_operator_command()` to account for the stdout, stderr, and fulline fields instead of combining them and then trying to parse them in `run_module()`, this way, you are guaranteed to always be accessing the correct data.

blakeinate (changes requested):
Updates look good Rich! I think I found a small bug you will want to address (given I interpreted the code correctly)

plugins/modules/zos_operator.py
richp405: len(result of split) = #/lines split.  The line just means "if there are at least 4 lines of output".  I can add that as inline comment.

plugins/modules/zos_operator.py
richp405: While I didn't think stdout or stderr could return without a newline (except for None), I definitely see the potential... good eyes!  My prior code was just checking to make sure there were > 3 lines.  I'll get a bit more elegant on this today.

plugins/modules/zos_operator.py
richp405: Looking back at the code, I'm realizing my thinking, that invalid or error always appeared on the third line...the test was to make sure there wasn't a bounds error.  Blake found an edge-case I hadn't considered, so the change I'll propose is to return the 3 pieces separately, then check the first few lines of stderr and stdout with mini-comprehensions, for the trigger text.

plugins/modules/zos_operator.py
```diff
@@ -129,40 +169,189 @@ def run_module():
...
-        result["content"] = rc_message.get("message").split("\n")
+
+        result["content"] = []
+        short_str = []
+        ssctr = 0
+        tstr = rc_message.get("stdout")
+        if tstr is not None:
+            for s in tstr.split("\n"):
+                result["content"].append(s)
+                if ssctr < 5:
```
ddimatos, line 180: We discussed this a few min ago but if you could add comments here as to why '5' is chosen, will help future reviews/developers

plugins/modules/zos_operator.py
```diff
@@ -129,40 +169,189 @@ def run_module():
...
+                    short_str.append(s)
+                    ssctr += 1
+        ssctr = 0
+        tstr = rc_message.get("stderr")
+        if tstr is not None:
+            for s in tstr.split("\n"):
+                result["content"].append(s)
+                if ssctr < 5:
```
ddimatos, line 188: We discussed this a few min ago but if you could add comments here as to why '5' is chosen, will help future reviews/developers

plugins/modules/zos_operator.py
```diff
@@ -129,40 +169,189 @@ def run_module():
...
+                    short_str.append(s)
+                    ssctr += 1
+
+        result["content"].append("Ran" + rc_message.get("call"))
+        result["changed"] = False
+
+        if int(result["rc"]) == 0:
+            if len(short_str) > 2:
```
ddimatos, line 196: We discussed this a few min ago but if you could add comments here as to why '2' is chosen, will help future reviews/developers

ddimatos (commented):
Hey @richp405 looks much better, just add a few comments as to what significance 5 & 2 have on the code, see the comments.

plugins/modules/zos_operator.py
richp405: comments added

plugins/modules/zos_operator.py
richp405: numbers explained in comments

plugins/modules/zos_operator.py
richp405: agreed.

ddimatos (approved):
Thanks @richp405 - these were great updates, thank you.

PR #256, opened 2021-03-15
Enhancement/3723

richp405:
##### SUMMARY

Enhancement 3723: Adding step-wise result codes for each step (assuming jes2 returned the values)
This looks at the DD returns from job_output for 'step was executed'... for each one found, a new entry is added to job.retcode[stepwise][nameofstep] set to the cc code returned

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
This adds a new sub-function to module_util/job.py job_output function, adding values where discovered in the DD sections

##### ADDITIONAL INFORMATION
zos_job_submit, and others, call a routine called job_output, which creates a result with job data and DD information.  
This enhancement adds a subsection under job[ret_code] called job[ret_code][stepwise], with entries for each step name discovered, along with the return code.


ddimatos (commented):
Wondering if we should advance one of the zos_job_output tests such that we drill down and check for the job step. For a known job like HELLLO it should be easy to do since we know the job steps ie `STEP0001`. Can discuss this more.

plugins/modules/zos_job_query.py
```diff
@@ -107,14 +107,51 @@
             Return code or abend resulting from the job submission.
           type: str
           sample: CC 0000
+        msg_code:
```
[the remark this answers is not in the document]
ddimatos: Never mind, I overlooked that you are calling the shared util jobs.py thus the routine actually does return the values. My bad :(.

plugins/modules/zos_job_output.py
```diff
@@ -318,7 +341,12 @@
           "code": 0,
           "msg": "CC 0000",
           "msg_code": "0000",
-          "msg_txt": ""
+          "msg_txt": "",
+          "steps": [
+            { "stepid": "STEP0001",
```
ddimatos, line 346: If you can update stepid to stepname it corresponds to SDSF JS panel names

richp405: Changed in code, and in documentation (zos_job_output, zos_job_query)

ddimatos: @richp405 some of the samples are not updated but the code is, I think I will wait to review this because [PR 260](https://github.com/ansible-collections/ibm_zos_core/pull/260/files) has a number of comments and both this PR and 260 share `module_utils/job.py`

plugins/modules/zos_job_output.py
```diff
@@ -195,11 +195,34 @@
...
+          elements: dict
+          contains:
+            step_name:
+              description:
+                Name of the step shown as "was executed" in the DD section.
+              type: str
+              sample: "STEP0001"
+            step_cc:
+              description:
+                Actual CC returned for this step in the DD section.
```
ddimatos, line 211: Real minor, using "Actual" might imply that elsewhere in the code we are deducing RC's. I would  replace "Actual" with "The" so it reads "The condition code (cc) returned for......"

plugins/modules/zos_job_query.py
```diff
@@ -110,14 +110,51 @@
...
+          elements: dict
+          contains:
+            step_name:
+              description:
+                Name of the step shown as "was executed" in the DD section.
+              type: str
+              sample: "STEP0001"
+            step_cc:
+              description:
+                Actual CC returned for this step in the DD section.
```
ddimatos, line 142: Very minor, see similar comment about use of "Actual"

plugins/modules/zos_job_submit.py
```diff
@@ -230,12 +230,36 @@
...
+          elements: dict
+          contains:
+            stepid:
+              description:
+                Name of the step shown as "was executed" in the DD section.
+              type: str
+              sample: "STEP0001"
+            stepcc:
+              description:
+                Actual CC returned for this step in the DD section.
```
ddimatos, line 246: See minor comment about the use of "Actual" in other comments

ddimatos (changes requested):
Looking good @richp405 , few really minor comments and I like the added printf's in the test cases, will help us pinpoint failures.

plugins/modules/zos_job_output.py
richp405: Hadn't thought about that.  Done

plugins/modules/zos_job_query.py
richp405: Done

plugins/modules/zos_job_submit.py
richp405: Done

plugins/modules/zos_job_output.py
richp405: I found 3 instances in zos_job_submit, and grepped the modules directory, but did not find others.

ddimatos (approved):
Looks good @richp405 - thanks for working through the updates with me on this, much desired feature from customers.

PR #260, opened 2021-03-26
Bugfix/5950

richp405:
##### SUMMARY
bugfix for jira 5950
The changes behavior of job_info to return a 'notfound' record, instead of a runtime exception, when a requested job is not found.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sub-features of job.py

##### ADDITIONAL INFORMATION
get_job_output  and get_job_info were both firing a runtime exception when a specified job was not found, which would interrupt a playbook, potentially unnecessarily.


plugins/module_utils/job.py
```diff
@@ -76,9 +76,47 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+
+
+def _job_not_found(job_id, owner, job_name, dd_name):
+    jobs = []
+
+    job = {}
+
+    job["job_id"] = job_id
+    job["job_name"] = job_name
+    job["subsystem"] = "NOTFOUND"
```
ddimatos, line 95: When the job string is passed in for a job that does not exist , I wonder if any of these `NOTFOUND` could be populated. Its likely you won't have that info but I am unsure without trying it out. 
Ie subsystem & system

plugins/module_utils/job.py
```diff
@@ -76,9 +76,47 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    job["system"] = "NOTFOUND"
```
ddimatos, line 96: Using I don't mind the use of `NOTFOUND` but we need to normalize/standardize on either use of spaces, case and verbosity. 

Let me further explain, some of the other properties have more verbose messages like `Job not found` where others are just `NOTFOUND`, `NOTFND`, so this is the verbosity difference I see and think case and use of spaces is pretty self explanatory.

plugins/module_utils/job.py
```diff
@@ -76,9 +76,47 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+
+    job["class"] = ""
+    job["content_type"] = ""
+
+    job["ddnames"] = []
+    dd = {}
+    dd["ddname"] = dd_name
+    dd["record_count"] = "0"
+    dd["id"] = ""
+    dd["stepname"] = "NOTFND"
```
ddimatos, line 113: Seeing inconsistencies where NOTFND vs NOTFOUND used.

plugins/module_utils/job.py
```diff
@@ -76,9 +76,47 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    job = {}
+
+    job["job_id"] = job_id
+    job["job_name"] = job_name
+    job["subsystem"] = "NOTFOUND"
+    job["system"] = "NOTFOUND"
+    job["owner"] = owner
+
+    job["ret_code"] = {}
+    job["ret_code"]["msg"] = "Job not found"
```
ddimatos, line 100: I question what we should have here for a job that does not exist in terms of ret_code, 

By the definition of the property:

- msg - would would be the job error such as: "JCL ERROR" | "ABEND(S|U)xxxx" | "CONV ABEND" | "SEC ERROR" | "SYS FAIL"| CANCELED"| "CC xxxx"

- msg_code - would the string value of the job error reduced down to something more consumable such as "00" ("00" | "04" | "08" | "S0C4"
- msg_txt  - is a place holder for us such that if we can deduce from the msg_code some message that is of value and known to us but does not come from z/OS or ZOAU we can inject it here to help users. For example:
"The target is unreachable"
"Error code xxxx received, please check yyyy is installed and configured"
- code - if and only if we can convert msg_code into an integer value we return it otherwise we return None
Something like msg_code :"00" can be converted to type int: code : 00
Something like msg_code :"ABEND", msg_code :"ABEND(S1234)", msg_code :"CANCELED", can not convert thus return None

JSON mockup for ret_code:
          "ret_code":{    
"msg":" CANCELED",  // "JCL ERROR" | "ABEND(S|U)xxxx" | "CONV ABEND" | "SEC ERROR" | "SYS FAIL"| CANCELED"| "CC xxxx"     
"msg_code":"00",        // "00" | "04" | "08" | "S0C4"      
"msg_txt":"Some message we can deduce from the return code",      
"code":04   
          }
          
 We can discuss, my thoughts are to keep it aligned to the above definitions so maybe:
 msg = 'JOB NOT FOUND'
 msg_txt = 'Unable to find a job matching {job_name} and {job_id}'
 code = None
 msg_code = "None" //unsure of this idea but can't think of a better suggestion.

plugins/modules/zos_job_output.py
```diff
@@ -80,7 +80,7 @@
 RETURN = r"""
 jobs:
   description:
-      List of jobs output.
+      List of jobs output, or a single NOTFOUND record.
```
ddimatos, line 83: Maybe something like this:
description:
     List of the jobs output or a single NOTFOUND record when the job has not been run. 
     
     
`NOTFOUND` could change depending on my earlier comments

plugins/modules/zos_job_query.py
```diff
@@ -77,7 +77,7 @@
   type: bool
 jobs:
   description:
-     The list of z/OS job(s) and status.
+     The list of z/OS job(s) and status, or a single NOTFOUND record.
```
ddimatos, line 80: Similar to other comment maybe slightly more verbose

The list of the z/OS job(s) and status  or a single NOTFOUND record when the job has not been run.

plugins/modules/zos_job_submit.py
```diff
@@ -124,7 +124,7 @@
 RETURN = r"""
 jobs:
   description:
-     List of jobs output.
+     List of jobs output, or a single NOTFOUND record.
```
ddimatos, line 127: Maybe something like this:
description:
List of the jobs output or a single NOTFOUND record when the job has not been run.

NOTFOUND could change depending on my earlier comments

plugins/module_utils/job.py
```diff
@@ -36,7 +36,7 @@ def job_output(job_id=None, owner=None, job_name=None, dd_name=None):
 
     Raises:
         RuntimeError: When job output cannot be retrieved successfully but job exists.
-        RuntimeError: When no job output is found
+        >>>>> NONONONO RuntimeError: When no job output is found
```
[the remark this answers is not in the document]
richp405: argh!  Yep...before I corrected the doc... the code was there.  I've committed the correction.

plugins/module_utils/job.py
richp405: This code is triggered when there is no return for the given JobID.

plugins/module_utils/job.py
```diff
@@ -76,9 +76,47 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    jobs = []
+    if out:
+        jobs = _parse_jobs(out)
+    if not jobs:
+        jobs = _job_not_found(job_id, owner, job_name, dd_name)
+
+    return jobs
+
+
+def _job_not_found(job_id, owner, job_name, dd_name):
```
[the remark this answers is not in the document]
richp405: I like the thought.  I've added it there, but tweaked the response to indicate 'no jobs found' (meaning ANY) as opposed to a specific job not being found.

plugins/modules/zos_job_output.py
richp405: I can see that

richp405: I changed the entries to:
    Returns:
        list[dict] -- The output information for a list of jobs matching specified criteria.
        If no job status is found, this will return an empty job code with msg=Job not found

Let me know if that needs more tweaking

plugins/module_utils/job.py
richp405: https://ibmmainframes.com/references/a21.html  steep name can be 8 characters, max, just like jobname.  So, it could be NOTFOUND.  I'll change to that, so we can see if that helps.

plugins/modules/zos_job_output.py
ddimatos: Looks good , thanks @richp405

plugins/modules/zos_job_query.py
ddimatos: @richp405 - seems that this one has been missed yet the other one is fixed
https://github.com/ansible-collections/ibm_zos_core/pull/260#discussion_r629592125

plugins/modules/zos_job_submit.py
ddimatos: @richp405 - seems that this one has been missed yet the other one is fixed or you are waiting for approval before updating others similar comments
https://github.com/ansible-collections/ibm_zos_core/pull/260#discussion_r629592125

plugins/module_utils/job.py
richp405: so, in job not found, should owner be "NONE" or None?

plugins/modules/zos_job_submit.py
richp405: I forgot about submit.

plugins/module_utils/job.py
ddimatos: This works for me, as we discussed this probably is better than 'none' to reduce some of the ambiguity of none.

richp405: system, subsystem, owner are now all none.  Step name within not found job is now blank.

plugins/module_utils/job.py
```diff
@@ -76,9 +73,52 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    job = {}
+
+    job["job_id"] = job_id
+    job["job_name"] = job_name
+    job["subsystem"] = None
+    job["system"] = None
+    job["owner"] = None
+
+    job["ret_code"] = {}
+    job["ret_code"]["msg"] = "Job Not Found"
```
ddimatos, line 97: Lets upper case this if you agree to match how we usually see an an error from z/OS for consistency.

plugins/module_utils/job.py
```diff
@@ -34,12 +34,9 @@ def job_output(job_id=None, owner=None, job_name=None, dd_name=None):
         job_name {str} -- The job name search for (default: {None})
         dd_name {str} -- The data definition to retrieve (default: {None})
 
-    Raises:
-        RuntimeError: When job output cannot be retrieved successfully but job exists.
-        RuntimeError: When no job output is found
-
     Returns:
         list[dict] -- The output information for a list of jobs matching specified criteria.
+        If no job status is found, this will return an empty job code with msg=Job not found
```
ddimatos, line 39: I know I am being overly critical here but "Job not found" is not equal to "Job Not Found", one has uppercase Not and Found.

plugins/module_utils/job.py
```diff
@@ -76,9 +73,52 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+
+    job["class"] = ""
+    job["content_type"] = ""
+
+    job["ddnames"] = []
+    dd = {}
+    dd["ddname"] = dd_name
+    dd["record_count"] = "0"
+    dd["id"] = ""
+    dd["stepname"] = "NOTFOUND"
```
ddimatos, line 110: If the Job is not found, would it be enough to just go with `None` for stepname? What is your thought @richp405

plugins/module_utils/job.py
```diff
@@ -90,6 +130,7 @@ def _parse_jobs(output_str):
 
     Returns:
         list[dict]: A list of jobs and their attributes.
+        If no job status is found, this will return an empty job code with msg=Job not found
```
ddimatos, line 133: See earlier comment about the msg "Job not found" not being equal to "Job Not Found"

plugins/module_utils/job.py
```diff
@@ -76,9 +73,52 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    dd["ddname"] = dd_name
+    dd["record_count"] = "0"
+    dd["id"] = ""
+    dd["stepname"] = "NOTFOUND"
+    dd["procstep"] = ""
+    dd["byte_count"] = "0"
+    job["ddnames"].append(dd)
+
+    if ovrr is not None:
+        job["ret_code"]["msg"] = "No jobs found"
```
ddimatos, line 116: The message case should be consistent with others, see prior message.  I am not referring to the fact that this is a plural use error, but the uppercase Not and Found in other messages. Error messages should all be standardized for users to parse on.

plugins/module_utils/job.py
```diff
@@ -76,9 +73,52 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    dd = {}
+    dd["ddname"] = dd_name
+    dd["record_count"] = "0"
+    dd["id"] = ""
+    dd["stepname"] = "NOTFOUND"
+    dd["procstep"] = ""
+    dd["byte_count"] = "0"
+    job["ddnames"].append(dd)
+
+    if ovrr is not None:
```
ddimatos, line 115: This error case (the plural one) seems to be missing the "job["ret_code"]["code"] = None" property, maybe I am misunderstanding why it was left off or maybe it is a coding error.

plugins/module_utils/job.py
```diff
@@ -311,12 +354,10 @@ def job_status(job_id=None, owner=None, job_name=None):
         owner {str} -- The owner of the job (default: {None})
         job_name {str} -- The job name search for (default: {None})
 
-    Raises:
-        RuntimeError: When job status cannot be retrieved successfully but job exists.
-        RuntimeError: When no job status is found.
-
     Returns:
         list[dict] -- The status information for a list of jobs matching search criteria.
+        If no job status is found, this will return an empty job code with msg=Job not found
```
ddimatos, line 359: See other messages about message case sensitivity above for `msg=Job not found`

plugins/modules/zos_job_output.py
ddimatos, line 83: I don't believe this description is accurately telling our users the outcome.

If a job is not found (assuming by name or id - I have to trace to code to see if its both) that the result would return a return code 00 and a msg "Job Not Found" or "No jobs found" , we don't return a NOTFOUND anymore for a job but it seems you still have it code for a job stepname which I commented below maybe we can just use "None" for a job step name when the job is not found. Either way I think we have some work to do here in the description/explanation. 

Couple of other things, maybe was you write the update you will see maybe why I was hesitant on supporting the plural an singular use case because now we need to instruct users with two different outcomes of a message depending on if its plural or not (depends which module is executing the code, i understand the plural use case) but lets talk about this. 

To be honest I am not even sure we should return a 00, but maybe because the module did not fail we should) for job["ret_code"]["msg_code"] = "00"

plugins/modules/zos_job_query.py
```diff
@@ -72,12 +72,13 @@
 RETURN = r"""
 changed:
   description:
-     True if the state was changed, otherwise False.
+    True if the state was changed, otherwise False.
```
ddimatos, line 75: I wonder if this doc is actually just copy/paste because in a query module (I need to check the code) do we ever change anything to cause the state to change?

plugins/modules/zos_job_query.py
```diff
@@ -72,12 +72,13 @@
...
   returned: always
   type: bool
 jobs:
   description:
-     The list of z/OS job(s) and status.
+    The output information for a list of jobs matching specified criteria.
+    If no job status is found, this will return an empty job code with msg=Job not found.
```
ddimatos, line 81: See other comments about case sensitivity for msg=Job not found.

plugins/modules/zos_job_submit.py
```diff
@@ -124,7 +124,8 @@
...
+     If no job status is found, this will return an empty job code with msg=Job not found.
```
ddimatos, line 128: Case sensitivity mentioned elsewhere for msg=Job not found.

plugins/modules/zos_job_submit.py
```diff
@@ -124,7 +124,8 @@
 RETURN = r"""
 jobs:
   description:
-     List of jobs output.
+     List of jobs output, or a single NOTFOUND record.
```
ddimatos, line 127: See my other comment , it appears this doc also needs to be updated, similar to my other comment.

plugins/modules/zos_job_submit.py
```diff
@@ -612,7 +613,7 @@ def query_jobs_status(module, jobId):
             )
     if not output and timeout == 0:
         raise SubmitJCLError(
-            "THE JOB CAN NOT BE QUERIED FROM JES (TIMEOUT=10s). PLEASE CHECK THE ZOS SYSTEM. IT IS SLOW TO RESPONSE."
+            "THE JOB CAN NOT BE QUERIED FROM JES (TIMEOUT=10s). PLEASE CHECK THE ZOS SYSTEM. IT IS SLOW TO RESPOND."
```
ddimatos, line 616: Can we change the case to more sentence structure (upper and lower case) since we have other edits now?

tests/functional/modules/test_zos_job_output_func.py
```diff
@@ -82,5 +82,19 @@ def test_zos_job_output_job_exists(ansible_zos_module):
...
+    hosts = ansible_zos_module
+    # This is for jira 5950: It is unlikely to find a job called "notajob"
+    # This is to see if the 'notfound' entry is returned
+    results = hosts.all.zos_job_output(job_name="NOTAJOB")
+    for result in results.contacted.values():
+        print(result)
         assert result.get("changed") is False
         assert result.get("jobs") is not None
+        assert result.get("jobs")[0].get("ret_code") is not None
+        assert result.get("jobs")[0].get("ret_code").get("msg") == "Job Not Found"
```
ddimatos, line 100: Case sensitivity here as well if we change the source.

tests/functional/modules/test_zos_job_output_func.py
```diff
@@ -82,5 +82,19 @@ def test_zos_job_output_job_exists(ansible_zos_module):
     hosts.all.file(path=TEMP_PATH, state="absent")
     results = hosts.all.zos_job_output(job_name="SAMPLE")
     for result in results.contacted.values():
+        print(result)
+        assert result.get("changed") is False
+        assert result.get("jobs") is not None
+
+
+def test_zos_job_output_nonexistant(ansible_zos_module):
```
ddimatos, line 90: How are you handling the plural version of your error message for query , "No jobs found"

plugins/module_utils/job.py
```diff
@@ -76,9 +73,52 @@ def _get_job_output(job_id="*", owner="*", job_name="*", dd_name=""):
...
+    job["job_id"] = job_id
+    job["job_name"] = job_name
+    job["subsystem"] = None
+    job["system"] = None
+    job["owner"] = None
+
+    job["ret_code"] = {}
+    job["ret_code"]["msg"] = "Job Not Found"
+    job["ret_code"]["code"] = None
+    job["ret_code"]["msg_code"] = "00"
```
ddimatos, line 99: Lets chat about this, because according to our contract/doc , we should return None for msg_code , see below "if and only if we can convert msg_code into an integer value we return it otherwise we return None" 

- msg_code - would the string value of the job error reduced down to something more consumable such as "00" ("00" | "04" | "08" | "S0C4"

- code - if and only if we can convert msg_code into an integer value we return it otherwise we return None
Something like msg_code :"00" can be converted to type int: code : 00
Something like msg_code :"ABEND", msg_code :"ABEND(S1234)", msg_code :"CANCELED", can not convert thus return None

plugins/module_utils/job.py
richp405: agreed

plugins/module_utils/job.py
richp405: I get it ... consistency.

plugins/module_utils/job.py
richp405: If we're sticking with pythonic solutions (NoneType), I agree.

plugins/module_utils/job.py
richp405: yep

plugins/module_utils/job.py
richp405: k

plugins/modules/zos_job_submit.py
richp405: I've tweaked it

plugins/modules/zos_job_submit.py
richp405: done

tests/functional/modules/test_zos_job_output_func.py
richp405: Thanks...that's the stuff that trips me up.

tests/functional/modules/test_zos_job_output_func.py
richp405: I'm not sure how to replicate that... it should only trigger if NO jobs are returned on the global query... while it's a possibility, I haven't seen it happen... any time I've run it there has been SOMETHING in the JES Queue.

plugins/module_utils/job.py
richp405: I think we agree code should be None, and the question is should msg_code be a simple integer?  Would changing the msg_code to "NOT FOUND" make sense?  I don't recall a length limit on that field, and that would become a string, not an integer, similar to 'SOC4".

plugins/module_utils/job.py
richp405: the plural case comes after the singular case, which already set the ret_code/code to None

ddimatos (approved):
Looks great @richp405 - took some work but we got it where we wanted, thanks for your patience.

PR #267, opened 2021-04-16
Updates to copy module to better explain behavior under particular co…

ddimatos:

##### SUMMARY

The zos_copy module:
-  had a typo corrected
- updated doc explaining the volume option, model_ds option
- updated doc explaining when to use volume and model_ds
- updated doc explaining what happens when no volume is specified
- updated doc to explain Co:Z is not supported and how to work around that

##### ISSUE TYPE

- Docs Pull Request


richp405 (approved):
Doc change.  Looks okay from here.

PR #275, opened 2021-04-24
Content/changes needed for zos ping

ddimatos:
##### SUMMARY
Changes change the zos_ssh plugin behavior to prioritize REXX based module execution over Python based modules such that the zos_ping module can correctly execute. Includes a dummy zos_ping.py module that contains the zos_ping module doc used for certified doc generation in automation hub. 

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -130,31 +130,31 @@ module-doc:
 	mkdir -p source/modules; \
 	fi
 
-	@if ! test -d ../plugins/modules/rexx_module_doc; then \
-	echo "Make ../plugins/modules/rexx_module_doc directory to extract REXX doc into temporary file."; \
-	mkdir -p ../plugins/modules/rexx_module_doc; \
-	else \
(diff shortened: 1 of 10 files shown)


ddimatos:
@richp405 - looks like it is all ready to go , got a connection error, looks like  ec33012a is down. Can wait till Monday so I can release v1.3.1

richp405 (approved):
commented out a big if, lots of doc tweaks, and a preference order change.  Looks fine.

ddimatos:
@richp405 - thanks and fyi the bit `if` i commented out was my original work around to my `make` file to generate doc for a non-python based module. Given I now have a work around where a non-python based module (REXX) can have a dummy python module with the same name contain only the python doc, i don't need that make file entry and this satisfies the Ansbile doc generation that happens on Automation Hub that was failing in 1.3.0.

PR #277, opened 2021-04-27
Changes made for module zos_ping to support doc generation

ddimatos:

##### SUMMARY
This change supports the doc generation that happens on Automation Hub and provides a fix such that the tooling can not have a python based module to extract doc from instead of a REXX based module (zos_ping). 
This also changes the execution order such that REXX based modules are prioritized over other such as python.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,54 @@
 Releases
 ========
 
+Version 1.3.1
+=============
+
+What's New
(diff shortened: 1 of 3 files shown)


richp405 (approved):
Standard doc changes... I'm surprised about removing automation hub from availability, but that's the only thing I wasn't expecting.

PR #283, opened 2021-06-28
Update common code used to obtain job content such it is not trucated by the unrelated record count

ddimatos:
##### SUMMARY
This commit fixes the bug such that when Job content is returned by the `zos_job_output` module (or any module using the jobs.py code to return job content) its not truncated by the incorrect use of of the record count variable. 

Record count does not equate to the lines in job content, the job content is 1 entry with regard to records from the record count, the job content just have its own iterator. No maximum has been set for this as I don't feel we don't have the justification to limit this and would rather the customer receive the full ouput. Should we ever need to limit you can set a REXX limiter like so `isflinelim = 500` , more on REXX and isfline [here](https://www-01.ibm.com/servers/resourcelink/svc00100.nsf/pages/zOSV2R3sc279028/$file/isfa600_v2r3.pdf) , [here](https://www.ibm.com/docs/en/zos/2.3.0?topic=isflog-special-variables-use-command) and [here](https://www.ibm.com/docs/en/zos/2.2.0?topic=panels-job-data-set-panel-jds)

No new tests cases need to be created, this was also a fixtest being evaluated by @th365thli

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_job_output module
jobs.py common utility

### DIFF
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -304,7 +304,6 @@ Say '-----NO JOBS FOUND-----'
 end
 else do
 do ix=1 to isfrows
-    linecount = 0
 
     Say '-----START OF JOB-----'
     Say 'job_id'||':'||value('JOBID'||"."||ix)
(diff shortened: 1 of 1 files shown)


richp405 (approved):
Change makes sense.... always starting at one.  One thing struck me though.... I don't see linecount used anywhere... not as output, nor as a rexx return... is it necessary at all?

PR #284, opened 2021-07-01
zos_ssh - rename retries setting

samdoran:
##### SUMMARY

There was a bug in Ansible that would result in task retries overriding the connection retries. Using a different name for the connection retries setting avoids this issue on versions of Ansible that are missing the bug fix.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`plugins/connection/zos_ssh.py`

##### ADDITIONAL INFORMATION
Related to https://github.com/ansible/ansible/pull/75155.

### DIFF
--- a/plugins/connection/zos_ssh.py
+++ b/plugins/connection/zos_ssh.py
@@ -141,7 +141,7 @@ DOCUMENTATION = """
             - key: ssh_extra_args
               section: ssh_connection
               version_added: '2.7'
-      retries:
+      reconnection_retries:
           # constant: ANSIBLE_SSH_RETRIES
           description: Number of attempts to connect.
           default: 3


ddimatos:
@samdoran - Thank you for the commit. I did look over [ansible/ansible#75155](https://github.com/ansible/ansible/pull/75155) and it looks like 75155 is going into Ansible 2.12 so this would be a breaking change for anyone using our collection with less than Ansible 2.12. 

We could branch it till then but wondering if this will be backported (I am a bit new to Ansible community and backports don't seem to a practice which I fully understand why).

That said, it just happens I was reworking this module and my next move was going to be to push a change to ansible-core to not incur the cost of having to maintain multiple versions for different levels of ansible. 

To buy us some time happen to be working on this [PR 285](https://github.com/ansible-collections/ibm_zos_core/pull/285) when yours came.

What I would expect in Ansible less than 2.12 with the `reconnection_retries` change:
```
KeyError: 'Requested entry (plugin_type: connection plugin: ansible_collections.ibm.ibm_zos_core.plugins.connection.zos_ssh setting: **retries** ) was not defined in configuration.'
EC33017A.vmec.svl.ibm.com | FAILED! => {
    "msg": "Unexpected failure during module execution.",
    "stdout": "
```

samdoran:
You could do a `try/except` in order to support the option name change.

```python
try:
    tries = self.get_option('retries')
except KeyError:
    tries = self.get_option('reconnection_retries')
```

> We could branch it till then but wondering if this will be backported

Since that change is a bugfix, it was [backported](https://github.com/ansible/ansible/pull/75191) to the current supported release (2.11) that is receiving general bug fixes. You can see more about the `ansible-core` release cycle [here](https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-release-cycle).

ddimatos:
@samdoran - thanks for the update, after this week I will have a good idea of my changes pass our regression testing and thus would no longer have a dependency on the fork (`zos_ssh.py`)  based on `ssh.py` thus eliminating this change.

PR #296, opened 2021-08-23
Enh5832

richp405:
##### SUMMARY
Changed the QuotedString routine to remove quotation marks instead of escaping them.
Added a line to the 'block' documentation to note that quotations would be removed.
Fixes 5832, where errors came from quotation marks in the block data.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_blockinfile.py

##### ADDITIONAL INFORMATION

- name: 1b Add to /test/profile.tmp quoted
      zos_blockinfile:
        src: "/test/profile.tmp"
        insertafter: "EOF"
        block: |
          "test text line 1b"
          "test text line 2b"
      register: result

The entry above, in a playbook, would crash out the run.  Doing it without the quotation marks, or with apostrophes works fine.  The issue is downstream, inside zoau or the command call chain.  It was easier to simply remove the quotation marks.

```
old routine, run against quotes would generate:
Module invocation had junk after the JSON data: ����@����@������KKK����@����@������KKK

and/or fail to run completely.  Anything returned would be invalid json, cause an error exit.

```

### DIFF
--- a/plugins/modules/zos_blockinfile.py
+++ b/plugins/modules/zos_blockinfile.py
@@ -61,6 +61,7 @@ options:
     description:
     - The text to insert inside the marker lines.
     - Multi-line can be separated by '\n'.
+    - Any double-quotation marks will be removed.
     required: false
     type: str
     default: ''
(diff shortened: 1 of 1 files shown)


richp405:
Pulled because I found a way to repair the return.

PR #303, opened 2021-10-11
Update zos_blockinfile to correctly manage double quotes in the block

richp405:
##### SUMMARY
When a user uses double quotes in the block section of the module, it fails with either of the below messages.
This pull request also includes an Ansible ingore.txt for version 2.13.

Case 1:
```
      zos_blockinfile:
        src: "/u/oeusr01/scripts/profile.tmp"
        insertafter: "EOF"
        block: |
          "test text line 1"
          "test text line 2"
      register: result
    - name: Response
      debug:
        msg: "{{ result }}"
```
Yeilds:
```
    "msg": "ZOAU dmod return content is NOT in json format",
    "rc": 1,
```

Case 2:
```
      zos_blockinfile:
        src: "/u/oeusr01/scripts/profile.tmp"
        insertafter: "export LIBPATH=${ZOAU_HOME}/lib:${LIBPATH}"
        block: |
          "test text line 1"
          "test text line 2"
      register: result
    - name: Response
      debug:
        msg: "{{ result }}"
```
Yields:
```
    "msg": "ZOAU dmod return content is NOT in json format",
    "rc": 16,
    "stderr": "BGYSC1302E Wrong number of parameters passed to dmod. 17 were given.\n",
    "stderr_lines": [
        "BGYSC1302E Wrong number of parameters passed to dmod. 17 were given."
    ],
```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_blockinfile

##### ADDITIONAL INFORMATION
The actual bug is coming from ZOAU dmod, where it returns a non-zero status.

While we can clean up the return data, it will still be non-zero because of the quotes.
I tried single, double, triple escapements, with generally the same results.
Shortest simple fix is to remove all double-quotes.  
Apostrophes are no issue.


ddimatos:
@richp405 - I know this is been completed for a while, given our circumstances it was understood we would have this delay. The changes are minimal which is good and thank you for the doc update; I do have some questions and comments.

1. Has an issue been raised to ZOAU for their awareness?
2. If you are removing the escaping to effectively remove double quotes for any text in a `block`, is there an alternative for users who do want to update text containing double quotes? 
3. Could we have possibly documented users to escape double quotes (my guess is no given this go back to an issue with ZOAU dmod). 
4. Do we need a test case for Double quotes, it would go a long ways to ensure we don't currently break but also for when zoau does support double quotes the test case would be in place. Maybe @fernandofloresg can help with this. 

Depending on your reply, I may need to update our changelog fragment from `bugfix` to `minor_changes` [here](https://github.com/ansible-collections/ibm_zos_core/blob/63326100189db7dd3e42573c0d4ca28e904a3a12/changelogs/fragments/303-correct-double-quotes.yml#L1) as I feel like this is more a change and not a bugfix. 

As a result  of this change we should open a new git issue to add support for double quotes in the block content for when ZOAU does incorporate that support. 

I am adding @fernandofloresg as a reviewer because he has made a number of changes to zos_blockinfile.

richp405:
@ddimatos : Answers:

1 - No... we had run out of people to notify when this came up.
2 - No, there is no alternative to include quotes... everything I tried, doubles-triples, escapes, double-escapes... each choked in a different way, and all caused various kinds of exceptions, because at that time, the command was going to zoau and then to a cmd/shell call.
3 - Originally, someone had added an escape in there, but that wasn't working, so I don't believe that was sent to the users.
4 - We don't have a test for quotes currently (I just checked)... This is a really convoluted test system (one of those with a mile of data at the top of the file).  It could be decent to add it, but it would still need changed when zoau can handle the quotation marks (we'd have to remove the remover)... python strings and quotes are tough to get right, because they're processed differently in functions versus data declarations.
Yes, this is more of a change than a bugfix.

ddimatos:
Thanks @richp405; 
- I will work with ZOAU to open an issue on `dmod`.
- I was afraid there would be no alternative to support quotes in the content, which means we must strip them out and let users know its not supported hence a `minor_change`. 
- This test suite has been on my radar for a rewrite, will factor that in with prioritization.

ddimatos:
Internal link to regression pipeline [successful log ](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/774/console)
image

ddimatos (approved):
Looks good @richp405 and thank you @fernandofloresg for the added review.

fernandofloresg:
LGTM, we discussed and don't have another option from the collection code. We opened a JIRA for ZOAU with regards to this issue NAZARE-9120.

PR #312, opened 2022-02-10
This function updates the existing code to leverage the latest API changes available in ZOAU 1.2.0

richp405:
##### SUMMARY
The overall pull request is in place to synchronize our code base with the latest ZOAU Python APIs. This in some places removes some interpreted REXX code while in other places removes shell calls into ZOAU. 

This does make use of ZOAU 1.2.0 but is not a hard dependency and remains backward compatible to other ZOAU versions. 

The interfaces and responses remain the same, no change to the spec; performance should have improved as well.  
This resolves internal Jira-6933.

In summary:

- module utility jobs - was updated to remove the useage of REXX and  replaced with ZOAU python APIs. This reduces code replication and it removes the need for REXX intepretretion which increases performance.
- ibm_zos_job_output - was updated to leverage the latest changes that removes the REXX code by calling the module utility jobs.
- ibm_zos_job_query - was updated to use the jobs module utility.
- ibm_zos_job_query - was updated to leverage the latest changes that removes the REXX code by calling the module utility jobs.
- zos_job_submit - was updated to include an additional error code condition JCLERR.
- zos_operater - was updated to remove the useage of REXX and replaced with ZOAU python APIs. This reduces code replication and it removes the need for REXX intepretretion which increases performance.
- zos_job_query - was updated to correct a boolean condition that always evaluated to "CANCELLED".

##### ISSUE TYPE

Technical Debt Pull Request

##### COMPONENT NAME

zos_job_output, zos_job_query, zos_job_operator


plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
         job_id {str} -- The job ID to search for (default: {None})
         owner {str} -- The owner of the job (default: {None})
         job_name {str} -- The job name search for (default: {None})
+        dd_name {str} -- If populated, return ONLY this DD in the job list (default: {None})
 
     Returns:
         list[dict] -- The status information for a list of jobs matching search criteria.
         If no job status is found, this will return an empty job code with msg=JOB NOT FOUND
-
+        new format: Job(owner=job[0], name=job[1], id=job[2], status=job[3], rc=job[4]))
```
ddimatos, line 129: Is this a left over comment, trying to understand how it relates to what is returned; sort of seems like you are documenting the dictionary returned but also looks like it was a development note.

plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
+        job_str (str): The content for a given dd.
 
     Returns:
-        Union[int, NoneType] -- Returns integer RC if possible, if not returns NoneType
+        list[dict]: A list of step names listed as "step executed" the related CC.
     """
+    stp = []
+    if "STEP WAS EXECUTED" in job_str:
+        pile = re.findall(r"(.*?)\s-\sSTEP\sWAS\sEXECUTED\s-\s(.*?)\n", job_str)
+        for match in pile:
```
ddimatos, line 170: Use of `pile` did make me smile, I have seen worse, not a deal breaker or anything that needs changing but for readability `for match in steps` or `step in steps` :)

plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
+            job["ret_code"]["msg"] = entry.status + " " + entry.rc
+            job["ret_code"]["msg_code"] = entry.rc
+
+            job["ret_code"]["code"] = ""
+            if len(entry.rc) > 0:
+                if entry.rc.isdigit():
+                    job["ret_code"]["code"] = int(entry.rc)
+
+            job["ret_code"]["msg_text"] = entry.status
+
```
ddimatos, line 217: This is merely a compliment in that appears you kept the code in sync with what the original interface defined the ret_code {} dictionary as (no action just a note for others and kuddos to you).

`
- msg: would would be the job error such as:
     eg; "JCL ERROR" | "ABEND(S|U)xxxx" | "CONV ABEND" | "SEC ERROR" | "SYS FAIL"| CANCELED"| "CC xxxx"
- msg_code: would the string value of the job error reduced down to something more consumable
     eg; "00" ("00" | "04" | "08" | "S0C4"
- msg_txt: is meant for supplying users with something helpful, human readable
	 eg; "The target is unreachable"
- code: if we can convert msg_code into an integer value we return it otherwise we return None
    eg; "00" can be converted to type int 00
`

plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
+                        continue
+
+                if "dataset" not in single_dd:
+                    continue
+
+                dd["ddname"] = single_dd["dataset"]
+                if "recnum" in single_dd:
+                    dd["record_count"] = single_dd["recnum"]
+                else:
+                    dd["record_count"] = "n/a"
```
ddimatos, line 238: If there is no record count (not sure what case this would be, maybe job has not executed) then is "n/a" a good response? I am thinking None for the pythonic way or "" for the non breaking way but I think our users understand the None paradigm.

plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
+
+                if "dsid" in single_dd:
+                    dd["id"] = single_dd["dsid"]
+                else:
+                    dd["id"] = "?"
+
+                if "stepname" in single_dd:
+                    dd["stepname"] = single_dd["stepname"]
+                else:
+                    dd["stepname"] = "UNAVAIL"
```
ddimatos, line 248: Same applies here, not sure what condition would allow us to experience no stepname but if there is one, don't you think the usage of None is a bit more programmatic friendly. think about it from an devops/automation perspective where there is no human on the other side and it sees a stepname "UNAVAIL" and acts on it, where if it got a None it would handle that case, but here it has no idea there is no stepname called UNAVAIL

plugins/module_utils/job.py
```diff
@@ -122,430 +114,190 @@ def _job_not_found(job_id, owner, job_name, dd_name, ovrr=None):
...
+
+                if "length" in single_dd:
+                    dd["byte_count"] = single_dd["length"]
+                else:
+                    dd["byte_count"] = 0
+
+                if ("stepname" in single_dd) and ("dataset" in single_dd):
+                    tmpcont = read_output(entry.id, single_dd["stepname"], single_dd["dataset"])
+                else:
+                    tmpcont = "no content available"
```
ddimatos, line 263: Wouldn't it be more efficient if you set tmpcont to None and then in some of the following logical checks eg; `if len(job["class"]) < 1` could be `if tmpcontent and  len(job["class"]) < 1` or wrap the entire block in another `if tmpcontent`, not a deal breaker just a suggestion.

plugins/modules/zos_job_submit.py
```diff
@@ -801,9 +803,7 @@ def run_module():
         except IndexError:
             pass
         except Exception as e:
-            result["err_detail"] = "{1} {2}.\n".format(
-                "Error during job submission.  The output is:", job_output_txt or " "
-            )
+            result["err_detail"] = "Error during job submission."
```
ddimatos, line 806: It looks like you corrected a blank message, given there is little we can provide the user they don't already know eg, job id, maybe we just be a bit more verbose such that it reads: "An error has occurred while submitting the requested job."

plugins/modules/zos_operator.py
```diff
@@ -256,123 +274,41 @@ def parse_params(params):
...
+        kwargs.update({"debug": "debug"})
 
     if params.get("wait"):
-        fulline += " WAIT"
-    else:
-        fulline += " NOWAIT"
+        wait = params.get("wait_time_s")
+        if wait:
+            kwargs.update({"timeout": "{0}".format(wait)})
+            kwargs.update({"parameters": "ISFDELAY={0}".format(wait)})
```
ddimatos, line 291: I glanced at ZOAU's [opercmd command api](https://www.ibm.com/docs/en/zoau/1.2.0?topic=modules-opercmd) and so no support the timeout and ISFDELAY which in this case both will be set to either True or False depending on the users entry. These are not returned parameters from what I gather from the doc so i wonder why we kept them.

ddimatos (commented):
@richp405 overall this looks great, not as much change as I though it would be. I did leave some comments for when you get a chance. Don't forget I have a PR full of changelog and other changes merging into this branch.
https://github.com/ansible-collections/ibm_zos_core/pull/338

plugins/module_utils/job.py
richp405: Yep... I'll re-push with that comment removed (there were too many things called job or jobs, so I left myself that note and forget to yank it.)

plugins/module_utils/job.py
richp405: bad habit of mine... I've changed pile to steps

plugins/module_utils/job.py
richp405: Thanks.

plugins/module_utils/job.py
richp405: We don't use the record count internally, so I've switched it to None.

plugins/module_utils/job.py
richp405: None may trip up a formatting string, but there should be no step header without a step name.  Switching to None.

plugins/module_utils/job.py
richp405: I switched to nested if with a Non default.  I'm just starting to get into the 'avoid else' thing.

plugins/modules/zos_job_submit.py
richp405: Yes.  If job_output_txt is None, this line caused exceptions.  If the error is there, stderr is already populated.  I'll change the text.

plugins/modules/zos_operator.py
ddimatos: I reach out to ZOAU and also went digging through the code and could not find the `timeout` support we requested, for now ill approve and merge and open an issue that would require us to come back and revisit this issue. @richp405

ddimatos:
The ci/cd and regression ran fine ,[internal link](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/781/)
image

ddimatos (approved):
@richp405 - looks good, thanks for the updates !!!

PR #313, opened 2022-03-04
Feature/2621/regex operator action query

rexemin:
##### SUMMARY

Enhancement 2621 (from Jira): add regex support for zos_operator_action_query.

Adds support for a new option in the module (message_filter) to filter raw outstanding messages using a valid Python regex.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_operator_action_query


ketankelkar:
just some thoughts:

plugins/modules/zos_operator_action_query.py:

1. in the examples section (line 109) maybe we can add a more descriptive 'name' about what the example does instead of what’s specified in the params, maybe something like: “Display all outstanding messages where the job name begins with ‘mq’, message id begins with ‘dsi’, on the system ‘mv29’ and which contain the pattern ‘*IMS.*’”.
2. is there any input sanitization done on the regex specified by user? i’m thinking that should go in the message_filter_type function (line 317). i noticed the other parameters go through the `validate_parameters_based_on_regex` function which perhaps won't work for a regex string but something similar might be a good idea to have. just something to think about.
3. can we simplify the logic in line 422? I couldn't think of anything without complicating the if code block.
`if message_filter is not None and not match_raw_message(match.string, message_filter):`. i think it's okay as is but it did take me a couple of takes to understand - just thinking in terms of future maintainability, perhaps adding a comment is the best we can do in this case.

tests/functional/modules/test_zos_operator_action_query.py
- add test case(s) for `"use_regex" : False`


Everything else looks good to me. Good job Ivan!

rexemin:
Thanks for the feedback! I pushed a [commit](https://github.com/ansible-collections/ibm_zos_core/pull/313/commits/64a1fbfc1d90b06f45b7b5bbe3528ac91d845e7d) with the changes you suggested, for point 2) there's no sanitization per se for the regex, but the `message_filter_type` function validates that it's a valid regex before using it to filter messages

If you have any other suggestions I'd like to know them, thanks again!

plugins/modules/zos_operator_action_query.py
```diff
@@ -55,6 +56,17 @@
...
     type: str
     required: false
+  msg_filter:
+    description:
+        - Return outstanding messages requiring operator action awaiting a
+          reply that match a regex filter.
+        - If the message filter is not specified, all outstanding messages
+          are returned regardless of their content.
+        - Valid Python regular expressions are supported. See L(the official 
+          documentation,https://docs.python.org/3/library/re.html) for more information.
```
ddimatos: Just a comment no need for change, I pointed my browser at "https://docs.python.org/library/re.html" and it resolved to "https://docs.python.org/3/library/re.html" somehow the latest which we might want to consider in the future if we are editing this module.

plugins/modules/zos_operator_action_query.py
```diff
@@ -72,11 +84,16 @@
   zos_operator_action_query:
       message_id: dsi*
 
-- name: Display all outstanding messages given job_name, message_id, system
+- name: Display all outstanding messages that have the text IMS READY in them
+  zos_operator_action_query:
+      msg_filter: ^.*IMS READY.*$
+
+- name: Display all outstanding messages given job_name, message_id, system, msg_filter
```
ddimatos: This is also minor:

"Display all outstanding messages given job_name, message_id, system, msg_filter" probably should be tweaked to "Display all outstanding IMS messages given a job_name, message_id and system using msg_filter."

plugins/modules/zos_operator_action_query.py
```diff
@@ -55,6 +56,17 @@
...
     required: false
+  msg_filter:
+    description:
+        - Return outstanding messages requiring operator action awaiting a
+          reply that match a regex filter.
+        - If the message filter is not specified, all outstanding messages
+          are returned regardless of their content.
+        - Valid Python regular expressions are supported. See L(the official 
+          documentation,https://docs.python.org/3/library/re.html) for more information.
+        - Regular expressions are compiled with the flag re.DOTALL.
```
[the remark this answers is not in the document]
ddimatos: I also updated this so marking it resolved.

ddimatos:
Internal pipeline r[egression link ran successfully ](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/800/console)
image

ddimatos (approved):
Great addition and happy to see the added test cases, thank you.

PR #321, opened 2022-04-27
Bugfix/8877/Fixes deletion of src on encoding error

rexemin:
The code failed while trying to copy files from a directory that needs to be converted, dirs_exist_ok=True should fix that. Additionally, I added two checks to always make sure that we don't accidentally delete source files when they're not temporary.

##### SUMMARY

Fixes NAZARE-8877: deletion of source files when encoding conversion fails.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_copy

##### ADDITIONAL INFORMATION

The bugfix will also change test cases a bit to make sure the module doesn't delete source files when encoding conversion fails.


tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1597,7 +1654,7 @@ def test_ensure_tmp_cleanup(ansible_zos_module):
             cmd="ls -l", executable=SHELL_EXECUTABLE, chdir="/tmp"
         )
         file_count_post = len(list(stat_dir.contacted.values())[0].get("stdout_lines"))
-        assert file_count_post <= file_count_pre
+        assert file_count_post == file_count_pre + 1
```
richp405, line 1657: hard-coding a number (like +1) is a little iffy here.  I'd suggest maybe file_count_post <= file_count_pre + 1, so that possible intermediate files don't trigger the assertion.  

This also doesn't address the possibility that a separate program (not us or ansible) has written or removed files from the temp folder during the run (/tmp is very common use).

richp405 (changes requested):
I'm concerned about an assertion based on file count.  Remember our test environments are sometimes shared, so this could fire up errors that are just bad timing.  Let's think about a different approach, like file enumeration instead of raw counts.

tests/functional/modules/test_zos_copy_func.py
ddimatos: It is a good point @richp405 makes, this unfortunately is a pattern happening in the test cases for zos_copy, i recently had to work on one , i modified it with a second assertion to not rely only on +1, maybe something like that can be helpful to review [here](https://github.com/ansible-collections/ibm_zos_core/blob/91309c783595c323fa4fc2d195d079bca48d84d4/tests/functional/modules/test_zos_copy_func.py#L1582)

rexemin: Thanks for the feedback, I agree with what @richp405 said. I'm thinking that file enumeration is probably the best way to make the test more robust. We could check the files with stat (like in the code you linked @ddimatos) but to make sure all the temp files (ansible payloads, encoding conversion dirs, etc.) that the module creates are not present on the system after a copy operation.

ddimatos: Regarding all the excessive files left behind, just an FYI @fernandofloresg has started that work item, think the first phase is identification of those files and what is responsible is currently in progress. It won't be a 1 sprint work item based on the CEE dumps and other things I see in there, so it will take time probably next quarter to track down where they are coming from.

ddimatos:
image
Internal regression[ link/log](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/1022/consoleText)

ddimatos (approved):
Looks good, thanks for this change !!!
@rexemin

PR #325, opened 2022-05-18
Nazare 8887 deletes mounted

### LINKED ISSUE #140: [Bug] [zos_data_set] should not ignore supplied volumes
##### SUMMARY
The `zos_data_set` module allows for the deletion of a dataset from a volume with `state: absent` and `volumes: "222222"`. However, when trying to delete an uncataloged dataset, the documentation says:
> If volumes is provided when state=absent and the data set is not found in the catalog, `zos_data_set` will check the volume table of contents to see if the data set exists. If the data set does exist, it will be cataloged and promptly removed from the system.

Therefore, if I have two datasets with the same name, but one is cataloged and one is not, even if I specify `volumes` to delete my uncataloged dataset, the cataloged dataset will instead be deleted. This can lead to the unintended deletion of datasets.

It seems that if `volumes` is provided, it shouldn't be ignored. It would make more sense to check the volume for the dataset regardless of whether it is cataloged or not, since providing a volume seems very intentional.

If this feature request were to not be implemented, could you please provide a recommendation on how to proceed?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`zos_data_set`

fernandofloresg:
##### SUMMARY

There is a bug when wanting to delete an uncatalogued dataset in certain volume and exists a catalogued dataset with the same name in a different volume. Even if we provide the volume serial for the uncatalogued one it will delete the catalogued one. 

Fixes #140 
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Modified module_utils/data_set.py (Changed behaviour of ensure_absent to take into account this scenario).
Modified module_utils/vtoc.py (Changed dd statement to VOLSER,vol ) in case there is not a VVDS in the volume.


plugins/module_utils/data_set.py
```diff
@@ -300,6 +306,85 @@ def data_set_cataloged(name):
...
+        Returns:
+            list{str} -- A list of volumes where the dataset is cataloged.
+        """
+        name = name.upper()
+        module = AnsibleModuleHelper(argument_spec={})
+        stdin = " LISTCAT ENTRIES('{0}') ALL".format(name)
+        rc, stdout, stderr = module.run_command(
+            "mvscmdauth --pgm=idcams --sysprint=* --sysin=stdin", data=stdin
+        )
+        delimeter = 'VOLSER------------'
```
richp405, line 324: The word should be spelled "delimiter", as in "it limits", not "delimeter", which looks like "it measures".

plugins/module_utils/data_set.py
```diff
@@ -300,6 +306,85 @@ def data_set_cataloged(name):
...
+            bool -- If the dataset is still present.
+            bool -- If given the volumes list and dataset name we need to continue with deleting the dataset as usual,
+            either by cataloging it and deleting or deleting a cataloged dataset.
+        """
+        changed = False
+        present = True
+        pending_to_delete_cataloged_dataset = False
+        # Get the list of volumes that the dataset is catalogued in.
+        cataloged_volume_list = DataSet.get_volume_list_for_cataloged_data_set(name)
+        if len(cataloged_volume_list) == 0:
```
richp405, line 371: I'd ask you to double-check if DataSet.get_volume_list_for_cataloged_data_set can not return "None".  If it can return none, you may need to check for that or len(None) would trigger an exception.  We've had so many odd issues from DataSet, I think it would be smart to double check.

richp405 (changes requested):
Double check the DataSet routines, fix the spelling error, and I'll pass it.

plugins/module_utils/data_set.py
fernandofloresg: Agree, will work on this.

plugins/module_utils/data_set.py
fernandofloresg: Will double check to make sure, do not remember testing that.

ddimatos: Looks ok, @fernandofloresg you wrote the [function](https://github.com/ansible-collections/ibm_zos_core/blob/1a025b441bdd632d052de963d4630fc0da8bf47f/plugins/module_utils/data_set.py#L310) seems to be the behavior you expect, nice addition.

ddimatos (approved):
Looks great, really nice additions to the utils and fixing a not so easy to find bug.

richp405 (approved):
Thank you for the fixes.  All good.

ddimatos:
Passed regression:
image

Internal [link/log](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/1066/consoleText) to regression

PR #338, opened 2022-06-18
Update the branch enh6933 used for pull request 312 by merging in the latest from branch dev.

ddimatos:
##### SUMMARY
This pull request resolves the divergent branches. It will bring branch enh6933 current with the content in the dev branch. 
This is a progressive step in the direction to update enh6933 so its pull requst 312 can easily merge with dev. 

##### ISSUE TYPE
- Docs Pull Request


richp405 (approved):
Looks okay.  I retract my previous question about ls versus ls -l, because I didn't realize the ansible shell pre-processed stdout that way.  Now I know.

ddimatos:
> Looks okay. I retract my previous question about ls versus ls -l, because I didn't realize the ansible shell pre-processed stdout that way. Now I know.

No worries @richp405 , it was a good point and question, i had to double back to make sure I did not error but then recalled why I could use ls.

PR #339, opened 2022-06-20
[zos_copy] Copy subdirectories when copying into USS

rexemin:
##### SUMMARY

Addresses a client's issue when trying to copy a local directory that has multiple subdirectories in its root into USS.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_copy

##### ADDITIONAL INFORMATION

The module already validates that the user is not trying to copy subdirectories into a PDSE, but this validation also affects trying to copy into USS. Since its possible to copy a complex tree structure into USS (and, in fact, the module already does this when using a remote source), the module now allows to copy subdirectories into USS from a local source.

When using `force=true` in the module parameters while copying a directory into USS, the module now returns `changed=true`.

### DIFF
--- a/changelogs/fragments/339-copy-sub-directories.yml
+++ b/changelogs/fragments/339-copy-sub-directories.yml
@@ -1,0 +1,6 @@
+bugfixes:
+  - >
+    zos_copy - fixes a bug that would not copy subdirectories. If the source
+    is a directory with sub directories, all sub directories will now be copied.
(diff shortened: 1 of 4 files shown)


ddimatos:
[Internal regression link](http://sprinkles1.fyre.ibm.com:8080/job/zosAnsible/812/console)
image

ddimatos (approved):
This looks great @rexemin , and thanks for all the added test cases.

PR #383, opened 2022-08-24
Update functional tests with added use cases for the connection api

aacheria:
##### SUMMARY

This pull request updates the functional test cases with some added negative tests that are testing various use cases such as exceptions.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/services/operations/connection.py
+++ b/services/operations/connection.py
@@ -80,15 +80,19 @@ class Connection:
         except BadHostKeyException as e:
             #if the server’s host key could not be verified
             print(e)
+            raise e
         except AuthenticationException as e:
             # authentication failed
             print(e)
+            raise e
         except SSHException as e:
             # if there was any other error connecting or establishing an SSH session
             print(e)
+            raise e
         except error as e:
             # if a socket error occurred while connecting
             print(e)
+            raise e
 
         return client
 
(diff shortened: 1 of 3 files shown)


ddimatos (changes requested):
@aacheria - hi, wondering if the exceptions are not flowing through after they were 'raise'd in the connection.py. If they are , I am not sure why you are not catching specific exceptions and inspecting them, for example and invalid key would probably raise a `BadHostKeyException` , even if we are going to update these later with our own, its still the correct approach unless the exceptions are not propagating

[Here is an example of one that I expect and then match to ensure its what I want to handle. ](https://github.com/ansible-collections/ibm_zos_core/blob/31b3fb1608c3b82a5995802d1abe1088595e8aeb/services/tests/test_cache.py#L165 )

Also be sure to pull , other changes have been made.

ddimatos:
Hi @aacheria - I did a quick pass, although technically not wrong to capture the state of a failed test with a boolean and use that to evaluate the test, its a bit ambiguous and not concise, it also bloats code a bit. I did link to an example above, for instance you have:
```
       passed = True
        try: 
        ......
        ..........
        except AuthenticationException as e:
            print(repr(e))
            passed = False
        assert not passed, "ASSERTION-FAILURE: Invalid username (string) does not raise exception"
```

Where you could have:
```
        try: 
        ......
        ..........
        except AuthenticationException as e:
            assert re.match(r'^AuthenticationException', repr(e))
````

The difference being its concise in what we are matching and returning the exception in the same string. In negative tests its not often we need an assertion message as we don't ever expect them to actually fail, if they do we will investigate.

ddimatos (changes requested):
See my thread comment, thank you

PR #394, opened 2022-08-31
command tests

aleonne:
##### SUMMARY
add unit tests for command services

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/services/tests/test_command.py
+++ b/services/tests/test_command.py
@@ -1,0 +1,98 @@
+from operations import command
+from operations.connection import Connection
+
+import os
+import re
+import unittest
+import sys
+import yaml
+sys.path.append('..')
+
+
+class TestCommandFunctionalTests(unittest.TestCase):
+    """
+    Test suite is used to drive command functional tests.
+
+    PyTest Usage:
+        cd {project}/ibm_zos_core/services/tests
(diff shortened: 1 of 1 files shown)


services/tests/test_command.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Run invalid command on target host
+        result = command._command(connection, "invalid")
+        stdout = result.to_dict()['stdout'][0]
+        assert "not found" in stdout,\
+            "ASSERTION-FAILURE: invalid command passed"
+
+    def test_command_run_command_on_host_set_enviroment(self):
+        'Test connecting to host with enviroment vars and run commands on host'
+        environment = {
```
ddimatos, line 65: This is not something I would hold a PR back on but a good update would be to do what you did with the connection properties and extract this environment configuration out of the code from being inline this way when the paths change as we update the mount/share, we don't need to update test cases.

ddimatos (approved):
Looks good, thank you, I did leave one comment but nothing that will prevent this from merging. 
```
This is not something I would hold a PR back on but a good update would be to do what you did with the connection properties and extract this environment configuration out of the code from being inline this way when the paths change as we update the mount/share, we don't need to update test cases.
```

PR #396, opened 2022-09-01
Update/9327/command service tests

aleonne:
##### SUMMARY
add environment vars to config file and update command_tests.py to use config file instead of environment dict
##### ISSUE TYPE
- Feature Pull Request


ddimatos (approved):
Looks good @aleonne - thank you.

PR #407, opened 2022-09-06
Deploy custom exceptions and update functional tests to use them.

aacheria:
##### SUMMARY
 Deploy custom exceptions and update functional tests to use them. #407 
##### ISSUE TYPE
- Feature Pull Request


ddimatos (approved):
Thank you Anushka, interesting the type errors are still propagating, ill have to follow the source to figure out where they are originating and see if we an wrap them, ill do that at a later time.

Issue #508, opened 2022-10-05
[Enhancement][pipeline] Update dependency finder to resolve a reference and object error


ddimatos:
### Enhancement or feature description

When the dependency finder runs in a new clone, I think the HEAD was what was cloned and not necessary the same commit hash/object in origin. 

It will yield to different warnings:
`see a error `refs/..... found at ./ but points to a different object' so its best to use the branch name`
`"warn: Are you sure you pushed 'HEAD' there?"`

and a non-zero return code of `1`

You can see the git request-pull source [here](https://github.com/git/git/blob/3dcec76d9df911ed8321007b1d197c1a206dc164/git-request-pull.sh#L119) and its code returning a non-zero error for a warning.

```
if test -z "$ref"
then
	echo "warn: No match for commit $headrev found at $url" >&2
	echo "warn: Are you sure you pushed '${remote:-HEAD}' there?" >&2
	status=1
elif test "$local_sha1" != "$remote_sha1"
then
	echo "warn: $head found at $url but points to a different object" >&2
	echo "warn: Are you sure you pushed '${remote:-HEAD}' there?" >&2
	status=1
fi
```

Issue #506, opened 2022-10-05
[Enabler][zos_mount] ccsid fix need to be forward ported to latest dev branch


ddimatos:
### Enabler description

zos_mount option `tag_ccsid` was corrected in a PR for 1.4.0-beta.2 , this needs to be forward ported to the dev branch to be included in future releases.

PR used for [1.4.0-beta.2](https://github.com/ansible-collections/ibm_zos_core/pull/502/)
Commit in question:
87990fe4bbb47464ea8e248378af1cf030acd08b
Branch left behind for cherry-picking or merging: `bugfix/352/zos-mount-parm-issue`

### Ansible module

zos_mount

ddimatos:
Branch `bugfix/352/zos-mount-parm-issue` deleted, closing

PR #511, opened 2022-10-05
Bugfix [zos_mount] ccsid type is string when it should be int, update dependency finder

ddimatos:
##### SUMMARY
This bugffix addresses zos_mount. It fixed option `tag_ccsid` to correctly allow for type int.
It addresses issue #506 

It also updates the dependency finder such that it can support pull request base branches that are not only `origin/dev`.
It addresses also issue #508 

##### ISSUE TYPE

- Bugfix Pull Request


ddimatos:
Pipeline log suffix `/job/zosAnsible/1253/console`
image

Issue #379, opened 2022-08-19
[Enhancement] Enable all tests once the patch for querying operator responses is ready


ddimatos:
ZOAU is providing a fix that will allow for all operator queries to show, till then this test suite has been disabled
Test suite [here](tests/functional/modules/test_zos_operator_action_query_func.py)

https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-8d0114f8cca6ca004cea300275c65156efd539c033b99ae5f622dc1dbd4a25f1R21

ddimatos:
My first attempt using my own backend configured to IBM Python 3.9 and ZOAU 1.2.1 had successful results. 

![Image](https://user-images.githubusercontent.com/25803172/197334897-54643dbd-96ac-4cf5-80c0-6347d96547f0.png)

I will make a PR and run it on the pipeline for comparison.

Issue #395, opened 2022-09-01
[Bug][Enabler] dependencyfinder.py can occasionally return the 3dot path prefix


ddimatos:
##### SUMMARY

NOTE: the rudimentary solution described is currently checked into the dev branch, this issue is to investigate and find another solution. 

As a result of some feature branches having `dev` merged into them as they had diverged over time the dependency analyzer would return a much larger change set that include many unrelated changes to a particular feature branch causing nearly a full test suite to run (hours). 

The solution I came up with was to use `git request-pull dev ./` instead of `git diff --name-status dev` and I enabled this with the new `-m` (minimum) option. 

For example if I ran git `request-pull dev ./` when this branch is checked out  `bugfix/6243/zos-copy-ds-member` I got back:
```
<some more unrelated text here that gets filter out> 
 changelogs/fragments/327-update-zos_copy-uss-to-pdse.yml |  8 +++++++
 plugins/modules/zos_copy.py                              | 87 ++++++++++++++++++++++++++++++++++++++++++---------------------------------
 tests/functional/modules/test_zos_copy_func.py           |  2 +-
 tests/helpers/zos_blockinfile_helper.py                  |  2 +-
 4 files changed, 59 insertions(+), 40 deletions(-)
```
where as if i ran this `git diff --name-status dev` i got back (which is later reduce based on the first bit M or D):
```
M       .gitignore
M       ansible.cfg
D       changelogs/fragments/318-sdsf-dependency-removed.yml
D       changelogs/fragments/321-fixes-deleting-src-on-encoding-error.yml
D       changelogs/fragments/322-new-zos-gather-facts.yml
A       changelogs/fragments/327-update-zos_copy-uss-to-pdse.yml
M       galaxy.yml
M       meta/runtime.yml
M       plugins/action/zos_job_submit.py
M       plugins/module_utils/job.py
D       plugins/module_utils/zoau_version_checker.py
M       plugins/modules/zos_copy.py
D       plugins/modules/zos_gather_facts.py
M       plugins/modules/zos_job_output.py
M       plugins/modules/zos_job_query.py
M       plugins/modules/zos_job_submit.py
M       plugins/modules/zos_operator.py
M       tests/dependencyfinder.py
M       tests/functional/modules/test_module_security.py
M       tests/functional/modules/test_zos_apf_func.py
M       tests/functional/modules/test_zos_copy_func.py
D       tests/functional/modules/test_zos_gather_facts.py
[...]

PR #546, opened 2022-10-23
Enable previously deprecated test cases for zos_operator and fix dependencyfinder

### LINKED ISSUE #547: [Enabler][dependencyfinder.py] fails to correctly provide a complete relative path
### Enabler description

Dependencyfinder.py is designed to run with `git request-pull` and perform analysis on which tests to run based on code changed. 

In a prior fix it was noticed that on occasion the code would return paths that were truncated with `...` instead of a relative path. Thus code like this was added to aid that:
```
            if "modules/" in line:
                if re.match('...', line):
                    line = line.replace("...", "tests")
```

Problem with this is that because of the truncation, it had to be matched only to `modules` but we have `modules` in a test path and plugin path, so how would the analyzer know which one it was looking out unless we walked a directory and kept an accounting of all the files under these and that iterator traversal would happen each time and be expensive.

For example when I run the command that is equivalent to 
```
    cmd = "git request-pull " + branch + " git@github.com:ansible-collections/ibm_zos_core.git " +  current_branch_name
```
with
```
 get_diff_pr = subprocess.Popen(
        ["git", "request-pull", branch, "git@github.com:ansible-collections/ibm_zos_core.git", current_branch_name],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        cwd=path,
```
I get 
```
ddimatos (2):
      Enable previously deprecated test cases
      Add trivial fragment

 .../546-enable-zos_operator-func-test.yml          |   3 +
 .../modules/test_zos_operator_action_query_func.py | 544 ++++++++++-----------
```
yet if i run the same command with `os.system` i get
```
ddimatos (2):
      Enable previously deprecated test cases
      Add trivial fragment

 changelogs/fragments/546-enable-zos_operator-func-test.yml      |   3 +++
 tests/functional/modules/test_zos_operator_action_query_func.py | 544 ++++++++++++++++++++++
```

running the command `git request-pull origin/dev git@github.com:ansible-collections/ibm_zos_core.git update/379/test-zos_operator_query-module` in shell:
```
      Enable previously deprecated test cases
      Add trivial fragment

 changelogs/fragments/546-enable-zos_operator-func-test.yml      |   3 +
 tests/functional/modules/test_zos_operator_action_query_func.py | 544 ++++++++++++++++++++++++++
```

So I realize it's not the git command failing us its the python subprocess and or more so, the shell being used by subprocess.

A shell command check of `echo $COLUMNS` shows me:
```
ddimatos:[ ~/git/github/ibm_zos_core/tests ]echo $COLUMNS
175
```
So it must be that this is what is needed by subprocess, to future proof, i set it to `500`
```
    get_diff_pr = subprocess.Popen(
        ["git", "request-pull", branch, "git@github.com:ansible-collections/ibm_zos_core.git", current_branch_name],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        cwd=path,
        env={'COLUMNS':'500'},
    )
```
and now that returns 
```
 changelogs/fragments/546-enable-zos_operator-func-test.yml      |   3 +++
 tests/functional/modules/test_zos_operator_action_query_func.py | 544 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### Ansible module

_No response_

ddimatos:

##### SUMMARY
This pull request will re-enable the test_zos_operator_func.py based on the fact they have been tested locally with zoau 1.2.1 such that now the issue #379  is resolved in ZOAU. 
image

This issues also resolves #547 which fixes a bug in the dependencyfinder where subprocess would return a truncated relative path because the columns value was not large enough. 

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/546-enable-zos_operator-func-test.yml
+++ b/changelogs/fragments/546-enable-zos_operator-func-test.yml
@@ -1,0 +1,3 @@
+trivial:
+- Enables zos_operator functional tests after testing ZOAU fix in 1.2.1.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/546)
--- a/tests/dependencyfinder.py
+++ b/tests/dependencyfinder.py
@@ -528,6 +528,7 @@ def get_changed_plugins(path, branch="origin/dev"):
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         cwd=path,
+        env={'COLUMNS':'500'},
     )
 
     stdout, stderr = get_diff_pr.communicate()
--- a/tests/functional/modules/test_zos_operator_action_query_func.py
+++ b/tests/functional/modules/test_zos_operator_action_query_func.py
@@ -18,296 +18,290 @@ __metaclass__ = type
 import pytest
 import unittest
 
+def test_zos_operator_action_query_no_options(ansible_zos_module):
+    hosts = ansible_zos_module
+    hosts.all.zos_operator(cmd="DUMP COMM=('test dump')")
+    results = hosts.all.zos_operator_action_query()
+    try:
+        for action in results.get("actions"):
+            if "SPECIFY OPERAND(S) FOR DUMP" in action.get("message_text", ""):
+                hosts.all.zos_operator(
+                    cmd="{0}cancel".format(action.get("number")))
+    except Exception:
+        pass
+    for result in results.contacted.values():
+        assert result.get("actions")
 
-class ModuleOperatorQueryTests(unittest.TestCase):
-    def setUp(self):
-        self.skipTest('Module tests disabled and waiting on ZOAU')
+def test_zos_operator_action_query_option_message_id(ansible_zos_module):
+    hosts = ansible_zos_module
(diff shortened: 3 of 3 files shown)


ddimatos:
image

Pipeline log suffix `/job/zos-multipath-4/481/console`

rexemin (commented):
So just to make sure, because the diff is a little janky, for the tests in `test_zos_operator_action_query_func.py` you only removed the class and left the test cases as they were before? @ddimatos

rexemin (approved):
Thanks @ddimatos, then the changes are pretty straightforward, good catch on the dependency finder bug.

ddimatos:
This PR also addresses issue https://github.com/ansible-collections/ibm_zos_core/issues/395

Issue #376, opened 2022-08-19
[Enhancement] Python submit failing to honor hfs=true


ddimatos:
Having to use work around _submit to avoid this issue. This will need to be reversed when zoau has a fix. 
See file module utils [job.py](plugins/module_utils/job.py)
See lines [242](https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-43f7641d31040832c1ee6d9297d962c8d025588af38b55987418737e8e4c42ddR242)
```
          # rstrip is a solution because we call zoau raw functions like _submit
            # which return values with a backslash
            job_id_stripped = entry.id.rstrip("\n")
```

See file [job_submit.py](plugins/modules/zos_job_submit.py)

Line [558](https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-84a6bdbc151b19c5b1b7fe67ee5e6e9f1c1e3c0832ca1ce85173af04c57f1fb1R558)

```
# Imports datasets, ZOAUException, JobSubmitException are only needed for the
# ZOAU submit(hfs=True) bug
```

Line [582](https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-84a6bdbc151b19c5b1b7fe67ee5e6e9f1c1e3c0832ca1ce85173af04c57f1fb1R582)
```
# Note that zoau submit will return None when wait to false, this is not desired
# so when wait = false, the raw version is called _submit.
```

Line [598](https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-84a6bdbc151b19c5b1b7fe67ee5e6e9f1c1e3c0832ca1ce85173af04c57f1fb1R598)
```
        else:
            # If wait is not set, fallback to our 10 second default using _submit()
            job_listing = jobs._submit(src, None, **kwargs)
            jobId = job_listing.stdout_response.rstrip("\n")
            print("PRINT {0}".format(jobId))
```

Line [613](https://github.com/ansible-collections/ibm_zos_core/pull/348/files#diff-84a6bdbc151b19c5b1b7fe67ee5e6e9f1c1e3c0832ca1ce85173af04c57f1fb1R613)

```
# Note: this method has a work around resulting from ZOAU bug when hfs = true.
# For now, the JCL is put into a temporary data set and run from ZOAU which has
# no issues doing. One limitation of this work around is that users can not
# set the HLQ used in the temporary data set. The second issue is that is that
[...]

Issue #418, opened 2022-09-15
[Enhancement] [zos_job_submit] No tests cases covering `wait_time_s`


ddimatos:
I noticed that the [zos_job_submit](https://github.com/ansible-collections/ibm_zos_core/blob/dev/tests/functional/modules/test_zos_job_submit_func.py) test suite has no test case that drives `wait_time_s` , happen to notice then when I was manually testing the change from the ZOAU dependency, we need some added test cases, probably at the time they were not sure how to accomplish this, this can be done with a JCL BPXBATCH job with a sleep. Ill share the JCL and later we can add it to the test cases and/or services framework.

```
//BPXSLEEP JOB MSGCLASS=A,MSGLEVEL=(1,1),NOTIFY=&SYSUID,REGION=0M
//USSCMD EXEC PGM=BPXBATCH
//STDERR  DD SYSOUT=*
//STDOUT  DD SYSOUT=*
//STDPARM DD *
SH ls -la /;
sleep 30;
/*
// 
```

ddimatos:
Addressed by PR #549

Issue #360, opened 2022-07-22
[Feature] [ibm_zos_job_submit] Does not support encoding options on remote USS or data sets


ddimatos:
The zos_job_submit module only converts encodings when the `location:LOCAL` , its not clearly documented is one problem.
This JIRA is requesting that we support minimally encoding for USS and if possible for DATA_SET

PR #549, opened 2022-10-24
Removes patch for when the src JCL is in USS, update zos_job_submit module and release 1.5 changes

ddimatos:
##### SUMMARY
This resolves issue #376 where JCL that is in USS would not submit and a temporary patch was put in place. This work undoes the patch and validates that ZOAU 1.2.1 `hfs=True` works correctly. 

This also addresses several other doc issues related to the default timeout of 10 seconds yet was documented at 60 seconds. 

Addresses issue #418 where there are no test cases to drive `wait_time_s`
Partially addresses issue #360 such that documentation has added to note that encoding is only supported for option `location` = `LOCAL`. 

Additionally:
- It cleans up code to remove unused dependencies to pass linting.
- It cleans up the unicode in our doc sections (hence the large binary looking diff)
- It cleans up some of the submit module wording used when DDs are not found, etc. 
- The 'wait' option has been deprecated because using option 'wait_time_s' implies the job is going to wait.
- The response 'message' property has been deprecated, all responses are now in response property 'msg'.
- The behavior changed when a volume is defined in the module options such that it will catalog the data set if it is not cataloged and submit the job. In the past, the function did not catalog the data set and instead performed I/O operations and then submitted the job. This behavior aligns to other module behaviors and reduces the possibility to encounter a permissions issue.
- The architecture changed such that the entire modules execution time now captured in the duration time which includes job submission and log collection. If a job does not return by the default 10 sec 'wait_time_s' value, it can be increased up to 86400 seconds.

##### ISSUE TYPE

- Bugfix Pull Request


ddimatos:
Local regression for job_submit
image

Last few lines of sanity test showing no warnings or errors:
```
Running sanity test 'use-compat-six' with Python 3.8
Running sanity test 'validate-modules' with Python 3.8
Running sanity test 'yamllint' with Python 3.8
ddimatos:[ ~/git/github/ibm_zos_core ]
```

ddimatos:
image
Url log suffix `/job/zos-multipath-4/513/`

richp405 (approved):
Looks okay... I will mention that some assertions & detail in assert messages in the test section were removed.  It won't hurt the process, but it may make it harder to trace some assertion failures (just on our side of the fence).

rexemin (approved):
Changes look good to me.

ddimatos:
> Looks okay... I will mention that some assertions & detail in assert messages in the test section were removed. It won't hurt the process, but it may make it harder to trace some assertion failures (just on our side of the fence).

Hi @richp405 , I went back and double checked, all assertions remained, might look like some were removed because of the assertion message that was line wrapping. I had originally added those assertion messages thinking they would help but after all the debugging i did for zos_submit, I found none of them helpful mainly because we get things like NoneType, etc. 

The most helpful thing was to print(result) for each test and just before that using `import inspect` i could print the test name using `print(inspect.stack()[0][3])` so that way I coordinate results and test cases easier. I think what we really need to do is deploy a few things to our growing test suite:
- Ability to run a subset (regression) across all test suites, ones we deem critical such that we can run sooner than 4 hrs and make that our nightly run , this can be done with `unitest`  , e.g.
```
nightly = TestSuite()
nightly.addTests(test_zos_copy_uss_to_pds)
nightly.addTests(test_zos_operator_cmd)
 runner.run(nightly)
```
- And then the second part of this would be how can we print the result on a failed test without (print stmts & pytest -s) because that level of verbosity is too much. My idea would be to expand our text figures or consider unitest [TextTestRunner](https://docs.python.org/3/library/unittest.html#unittest.TextTestRunner) as it has a stream option , guessing something like 
```if __name__ == '__main__':
    nightly = unittest.TestLoader().loadTestsFromTestCase(some_tests)
    runner = unittest.TextTestRunner(stream=open(os.devnull, 'w'))
    result = runner.run(nightly)
    if len(result.failures) or len(result.errors):
        print(inspect.stack()[0][3])
        print(result)
```
I figured i would open an issue for this enhancement because it really would increase productivity, our growing code base, demand, usage , etc , such tooling is going to be critical for us to be able to sustain the project.

Issue #408, opened 2022-09-08
[Feature][zos_copy] support reverting changes that only partially succeed during a failure


ddimatos:
##### FEATURE
The `zos_copy` module could benefit from some added state that tracks what changes are happening and on failure of a module ensure they are reverted. One such case is a directory of files being copied to PDS/E data set. Lets assume we have 10 files and part of the way through the copies start to fail. This could be because of space or some other reason but now we would exit and leave the PDS/E partially updated. I fee we should revert the changes to bring the state back to its original state. We may not be able to revert  back every dest state but can at least start with working the members. 

In `copy_to_member` this code could be updated to track state:
```
 for member in members:
                copy_src = "{0}({1})".format(src_data_set_name, member)
                if dest_member:
                    dest_copy_name = "{0}({1})".format(dest, dest_member)
                else:
                    dest_copy_name = "{0}({1})".format(dest, member)

                result = self.copy_to_member(copy_src, dest_copy_name)

                if result["rc"] != 0:
                    msg = "Unable to copy data set member {0} to data set member {1}".format(new_src, dest_copy_name)
                    raise CopyOperationError(
                        msg=msg,
                        rc=result["rc"],
                        stdout=result["out"],
                        stderr=result["err"]
                    )
```

PR #551, opened 2022-10-27
[Feature] [zos_copy] Restore members on module failure

rexemin:
##### SUMMARY

Implements #408.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_copy

##### ADDITIONAL INFORMATION

The module now keeps track of overwritten and new members when copying into a PDS/PDSE. When an error occurs during the copy operations, the `restore_backup` function restores overwritten members and erases new members, without reallocating the data set.

### DIFF
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -2251,7 +2251,6 @@ def run_module(module, arg_def):
                 dest_data_set=dest_data_set,
                 volume=volume
             )
-            raise Exception()
     except Exception as err:
         if dest_exists:
             restore_backup(dest_name, emergency_backup, dest_ds_type, use_backup)
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -1608,8 +1608,6 @@ def restore_backup(
                         response.stderr_response
                     )
             else:
-                # TODO: check that new dest also gets erased when it's newly allocated.
-                # TODO: check other restore_backup path (when allocation fails)
                 if not members_to_restore:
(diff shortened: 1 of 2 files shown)


richp405 (approved):
Looks okay to me.  I like the way you worked with lists there (hadn't seen that approach before).

plugins/modules/zos_copy.py
```diff
@@ -1581,15 +1592,55 @@ def restore_backup(dest, backup, dest_type, use_backup, volume=None):
...
+                    members_to_delete = []
+
+                for member in members_to_restore:
+                    response = datasets._copy(
+                        "{0}({1})".format(backup, member),
+                        "{0}({1})".format(dest, member)
+                    )
+
+                    if response.rc != 0:
+                        raise CopyOperationError(
```
fernandofloresg, line 1623: Maybe it would be a nice addition to, in case of error restoring members return a list of members restored and not restored. @rexemin Let me know your comments.

fernandofloresg (approved):
Overall looks good to me, I understand the use of list comprehension instead of simple for loops and think is justified, worst case we just go twice through the list of members.

plugins/modules/zos_copy.py
rexemin: I agree, that would help users understand better in what state their system is after the module fails. I'll add the lists to the error message.

fernandofloresg:
Screenshot 2022-11-15 at 13 02 20
Screenshot 2022-11-15 at 13 02 39

Issue #556, opened 2022-11-05
[Enabler][zos_job_submit] has not max_rc test cases 


ddimatos:
### Enabler description

The `job_submit` module option `max_rc` controls if a job should be considered a failure based on the jobs RC, there are no test cases driving this function and when I tried to add one, it did not perform as expected, this will address the test case and the update to support max_rc as it should.

### Ansible module

zos_job_submit

ddimatos:
Test cases added in PR #557

PR #557, opened 2022-11-05
Rework of module to support max_rc, timeout and jcl errors

### LINKED ISSUE #555: [Bug][zos_job_submit] When JCL ERROR occurrs the module returns back the error as a RC which is not correct
### Bug description

I purposely forced a syntax error in the JCL, the JCL had similar error when viewed from the job log:
````
13.46.51 JOB00406  IEFC452I HELLO - JOB NOT RUN - JCL ERROR  596  
```
The job never completed because it never started given the syntax error yet the module returned a RC of 596 which was suspicious to begin with, further investigation of the below snippet:
```
"ret_code":{
      "msg":"JCL ERROR  596",
      "msg_code":"596",
      "code":596,
      "msg_text":"JCLERR",
      "steps":[
         
      ]
```
I could see that the RC was coming from a ZOAU call:
```
entries = listing(job_id=job_id_temp)
.....
job["ret_code"]["msg"] = entry.status + " " + entry.rc
```

Actually the zoau version is 1.2.1, there is no option in the selection field yet for that, will update that as soon as the field is updated. 

### Playbook verbosity output

_No response_

### Contents of `ansible.cfg`

_No response_

### Contents of the inventory

_No response_

_No response_

### Ansible version

```shell
2.11.12
```


v1.4.0-beta.2

### IBM ZOAU version

v1.2.0

### z/OS version

2.4

### Ansible module

zos_job_submit

ddimatos:

##### SUMMARY
Issue resolves #555 and #556 which ended up being a small rewrite of the module which also includes missing test cases. 

This change ensures that if a max_rc is set then if a job fails with non-zero the max_rc will be the deciding factor if the module fails. max_rc allows for override of the jobs rc. 

This change also addresses the issue when a JCL Syntax error or JCL error in general occurs, the JCL is not interpreted as a return code, a patch was put in place to manage this until the ZOAU reported issue is resolved. 

##### ISSUE TYPE

- Bugfix Pull Request


ddimatos:
Draft for now because I will be pushing doc changes later to explain the modules behavior more clearly.

ddimatos:
Regression suffix `/job/zos-multipath-4/526/console`
image

richp405 (approved):
looks okay from here... I was surprised at changing the rc to integer type, but it looks complete.

ddimatos:
> looks okay from here... I was surprised at changing the rc to integer type, but it looks complete.
@richp405 any issue you forsee with doing so? I honestly don't know why it was a string to begin with given that ret_code.code is an integer as well.

richp405:
@ddimatos The integer makes more sense in my opinion.  I believe it was string because of some wierd output we got from the rexx code if bad jcl was submitted (invalid character).  With the new use of the zoau api, that shouldn't happen, so the integer definitely makes more sense.

ketankelkar (approved):
no issues jumped out, lgtm

Issue #554, opened 2022-11-04
[Enhancement] [zos_copy] Optimize walking of destination when mode is set


rexemin:
### Enhancement or featture description

Whenever `zos_copy` tries to copy files into a Unix System Services directory, it walks the entire file tree, starting at the root of the destination path, to get the permissions set so that it can restore them after the copy operation. If the destination has a really complex structure, this part of the module can take a long time and also be affected by symbolic links, so it would be better to limit this traversal to only the files that we know are going to be overwritten (files inside the source).

### Ansible module

zos_copy

PR #561, opened 2022-11-10
Enhancement/v1.4.0/zos copy dirs

### LINKED ISSUE #553: [Enhancement] [zos_copy] Create nested subdirectories when copying
### Enhancement or featture description

When copying files into Unix System Services, if the parent directories of a destination don't exist already on the system, the task fails. It is documented as the expected behavior in the module's documentation:

```
dest:
    description:
      ...
      - If C(src) and C(dest) are files and if the parent directory of C(dest)
        does not exist, then the task will fail
```

It would enhance the user experience if the module could also create missing directories in this case.

An example is trying to copy a text file into the path `/application/tempdir/workflows/`, where only `application/` but not `application/tempdir/` exist. The task would look like the following:

```
- name: Copy a text file.
      zos_copy:
        src: /tmp/a.txt
        dest: /application/tempdir/workflows/
```

And on ibm_zos_core 1.4.0-beta.2, the result is this: `fatal: [zvm]: FAILED! => {"changed": false, "msg": "Destination /application/tempdir/workflows/ is not writable", "rc": null, "stderr": "[Errno 129] EDC5129I No such file or directory.: '/application/tempdir/workflows/'", "stderr_lines": null, "stdout": null, "stdout_lines": null}`

### Ansible module

zos_copy

rexemin:
##### SUMMARY

Fixes #553 and #554.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_copy

##### ADDITIONAL INFORMATION

For #553: added a test case for the creation of missing parents.
For #554: the module now only gets the original permissions for files and subdirectories that will be overwritten, instead of walking the whole directory tree for the destination.

Additionally, when the source is a file and the destination is a directory, the module computes the final destination path before trying to backup. Before this change, it tried to backup the whole destination directory, when it only needed to backup one file.

### DIFF
--- a/docs/source/modules/zos_copy.rst
+++ b/docs/source/modules/zos_copy.rst
@@ -68,7 +68,7 @@ dest
 
   ``dest`` can be a USS file, directory or MVS data set name.
 
-  If ``src`` and ``dest`` are files and if the parent directory of ``dest`` does not exist, then the task will fail
+  If ``dest`` has missing parent directories, they will be created.
 
   If ``dest`` is a nonexistent USS file, it will be created.
 
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -1094,7 +1094,7 @@ class USSCopyHandler(CopyHandler):
             # Restoring permissions for preexisting files and subdirectories.
             for filepath, permissions in original_permissions:
                 mode = "0{0:o}".format(stat.S_IMODE(permissions))
-                self.module.set_mode_if_different(os.path.join(dest_dir, filepath), mode, False)
+                self.module.set_mode_if_different(os.path.join(dest, filepath), mode, False)
         except Exception as err:
             raise CopyOperationError(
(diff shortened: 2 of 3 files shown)


richp405 (commented):
Looks okay so far... just wanted you to know I looked through it.  Let me know when you're ready for formal review - richp

rexemin:
Thanks for the preliminary review @richp405. I think this PR is ready for a formal review, especially around the test cases. I didn't add new ones for #554, as that is somewhat tested already by `test_copy_local_dir_and_change_mode`, but if you have any suggestions for adding more test coverage please let me know.

richp405 (approved):
looks fine

rexemin:
Ran the Jenkins pipeline on the current state of this branch.

Screen Shot 2022-11-14 at 11 58 14
Screen Shot 2022-11-14 at 11 58 37

fernandofloresg (approved):
LGTM.

ketankelkar (approved):
lgtm

Issue #587, opened 2022-12-12
[Bug] [zos_copy] Issue with module zos_copy z/OS in z/OS core 1.4.0 with attributes force=True and backup=False


jbyibm:
### Bug description

We just give a try with IBM Ansible Core 1.4.0, and we have an issue we don't understand with zos_copy. (A blocker for our customer)

We have set force=True and backup=False. This should force the copy even if the destination member is already present in the target  PDS and do not create a backup.  In our use case the member is already present in the target PDS.

But we have this message. With all results values sets to null.
We see in the message that zos_copy is trying to create a backup. Why backup=False?

How to invest ? We have nothing  in the response (null value).

```
fatal: [zos_host_01]: FAILED! => {
    "changed": false,
    "failed_when_result": true,
    "invocation": {
        "module_args": {
            "backup": false,
            "dest": "NAZARE.GENAPP.BUG140.LOADLIB(LGACDB02)",
            "force": true,
            "is_binary": true,
            "remote_src": true,
            "src": "/tmp/wd_deploy/jenkins/zos_host_01/ansible.izgdvvv2/CICSLOAD/LGACDB02.CICSLOAD"
        }
    },
    "msg": "BackupError('An error occurred during backup: \"Unable to backup NAZARE.GENAPP.BUG140.LOADLIB(LGACDB02) to NAZARE.GENAPP.BUG140.LOADLIB(T14IN0H9)\"')",
    "rc": null,
    "stderr": null,
    "stderr_lines": null,
    "stdout": null,
    "stdout_lines": null
}
```

### Playbook verbosity output

```
fatal: [zos_host_01]: FAILED! => {
    "changed": false,
    "failed_when_result": true,
    "invocation": {
        "module_args": {
            "backup": false,
            "dest": "NAZARE.GENAPP.BUG140.LOADLIB(LGACDB02)",
            "force": true,
            "is_binary": true,
            "remote_src": true,
            "src": "/tmp/wd_deploy/jenkins/zos_host_01/ansible.izgdvvv2/CICSLOAD/LGACDB02.CICSLOAD"
        }
    },
    "msg": "BackupError('An error occurred during backup: \"Unable to backup NAZARE.GENAPP.BUG140.LOADLIB(LGACDB02) to NAZARE.GENAPP.BUG140.LOADLIB(T14IN0H9)\"')",
    "rc": null,
    "stderr": null,
    "stderr_lines": null,
    "stdout": null,
    "stdout_lines": null
}
[...]

PR #590, opened 2022-12-14
[zos_copy] Bugfix/588/zos copy emergency backup

### LINKED ISSUE #588: [Bug] [zos_copy] Exclude emergency backup creation when Force = True.

As happened in #587, there is an emergency backup being created while `backup` is False, in that case we should not try to create a backup and restore existing dataset.

Since the doc mentions: 

```yaml
backup:
    description:
      - Specifies whether a backup of the destination should be created before
        copying data.
      - When set to C(true), the module creates a backup file or data set.
      - The backup file name will be returned on either success or failure of
        module execution such that data can be retrieved.
    type: bool
    default: false
    required: false
```

### Ansible module

zos_copy

rexemin:
##### SUMMARY
Fixes #588.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_copy

##### ADDITIONAL INFORMATION

Removed the use of an emergency backup. If a user wants a backup of the destination, the original backup code is still in the module and it's run when `backup=True`.

### DIFF
--- a/changelogs/fragments/588-update-emergency-backup.yml
+++ b/changelogs/fragments/588-update-emergency-backup.yml
@@ -1,0 +1,5 @@
+minor_changes:
+- zos_copy - enhanced to only create destination backups when module option
+  `backup` is true, removing emergency backups meant to restore the system
+  to its initial state in case of a module failure.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/589)
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -6,7 +6,7 @@ namespace: ibm
 name: ibm_zos_core
 
 # The collection version
-version: 1.4.0
+version: 1.4.1
 
 # Collection README file
 readme: README.md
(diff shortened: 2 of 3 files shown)


fernandofloresg:
Changed the approach, instead of removing the code we better keep it, it is working and only had 1 issue reported. Rather, a new approach is taken, whenever we have the Force flag set as true we do not care about data integrity, so we don't create the backup.

fernandofloresg:
Screenshot 2023-01-02 at 16 02 42
Screenshot 2023-01-02 at 16 03 05

ketankelkar (approved):
LGTM, nice catch around the intention of the 'force' option

fernandofloresg:
> Just reviewed the changes made by @fernandofloresg, they look good. Definitely feels like the behavior is now more in line with the intention of `force`. Two questions:
> 
> 1. Did you have to run the Jenkins pipeline more than once due to errors during the recovery backup?
> 2. I remember that we discussed the possibility of not stopping module execution if the recovery backup fails, what was the conclusion to that?

Hey @rexemin,
1. Not for this version.
2. I don't think is needed anymore, at least for 1.4.0, if backup fails during normal copy user has the ability to force the copy, essentially removing that limitation on backup failure.

ddimatos:
Also addresses issue #587

PR #614, opened 2023-01-19
Update make file to read exported var VENV else default to 'venv'

ddimatos:

Overrides the original behavior of the Makefile that always expected one venv called 'venv'. By reading an exported var VENV, the make file and be used to interact with any number of virtual environments expediting localized testing. 

For example I will pass in my own requirements.txt then instruct the Makefile which make file and directory to build it for.
```
export VENV=venv-2.11
make vsetup req=requirements-ac-2.11.12.txt
```
```
ls -la 
drwxr-xr-x  10 ddimatos  staff     320 Jan 18 20:56 venv-2.11
```

Now I can do this:
```
ddimatos:[ ~/git/github/ibm_zos_core ]export VENV=venv-2.11
ddimatos:[ ~/git/github/ibm_zos_core ]make sanity version=3.8
Running sanity test "action-plugin-docs"
Running sanity test "ansible-doc"
```
or switch between other VENVs buy exporting the env var VENV , eg `export VENV=venv-2.14`

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,11 @@
 CURR_DIR := $(shell pwd)
 WHO := $(shell whoami)
 HOST_PYTHON = python3
-VENV = venv
+# VENV = venv
+# VENV := $(shell echo $$VENV)
+VENV := $(shell echo "$${VENV:-defaultValue}")
 VENV_BIN=$(VENV)/bin
+
 ZOS_PYTHON_DEFAULT=3.8
 ZOAU_DEFAULT=1.1.1
 # Test if docker is running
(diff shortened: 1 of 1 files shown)


ddimatos:
> Nice straightforward solution. I'm guessing -defaultValue is defined elsewhere, as it wasn't in the changes.

Ouch, good catch Rich, i was testing with that value "defaultValue" to see if it would actually create a venv called defaultValue if the ENV var VENV was not exported and forgot to change it back to what the real default is `venv`. Thanks for that catch.

Issue #595, opened 2023-01-09
[Enabler] CI/CD update to support and allow Ansible 2.13 selection 


ddimatos:
### Enabler description

Update our pipeline to allow for 2.13 ansible-core to be selected and run with regression. 

Includes any ignore files.

ddimatos:
Signed-off-by: ddimatos <dimatos@gmail.com>

This addresses community work items:
- [ansible 2.12](https://github.com/ansible-collections/ibm_zos_core/issues/617)
- [ansible 2.13](https://github.com/ansible-collections/ibm_zos_core/issues/595)
- [ansible 2.14](https://github.com/ansible-collections/ibm_zos_core/issues/596)
- [ansible devel (latest)](https://github.com/ansible-collections/ibm_zos_core/issues/618)

This completed work item adds support:
- to select any one of these ansible/ansible-core versions ['2.9.27', '2.11.12', '2.12.10', '2.13.7', '2.14.1', 'LATEST']
- that aligns our frozen versions to what the community source and requirements are for the selectable versions
- that streamlines the jenkins file with reusable vars to build a dockerfile
- adds much more verbosity such one can validate what was built and what the packages are including a dependency tree in case something gets auto updated
- with more comments with explanations removing one from guessing why something was done
- for ability to select if verbosity should be included in the log
- for ability to run ONLY sanity tests which will skip our function and unit tests stage (this is helpful when you want to ansible-test only code)
- Jenkins build log update to show the target, pyz version, ansible/ansible-core version and branch for easier browsing. 
- update to ansible.cfg to handle both connection headers for pipelining given 2.11 changes from `ssh_connection` to `connection` , a header is ignored by ansible if it does not know it, its fine to have both here.

Complete screenshot of the above, the only reason 305 failed was related to actual ansible-test issues that are not meant to be addressed in this work item. That is release work but the pipeline did work as designed. All pipelines were inspected to ensure pip packages remained. 
<img width="1315" alt="image" src="https://user-images.githubusercontent.com/25803172/214653382-243b3fa5-60fc-4ac8-8a84-3c033bbbfda8.png">
<img width="1254" alt="image" src="https://user-images.githubusercontent.com/25803172/214654214-8d7c3243-2530-41cc-9b3e-aa611437c56b.png">

ddimatos:
I also added the logic to correctly build a galaxy package since build_ignore is not technically supported in ansible 2.9 and we had to have ignore entries for docs and test directories before.

[...]

PR #627, opened 2023-02-03
Update/595 596 617 618/ansible core 2 12 latest support

ddimatos:
##### SUMMARY
More details to come but this pull request (draft) is in place to update our collections certification for all versions of ansible 2.9, 2.11 and now the added 2.12 - 2.15.
This pull request addresses issues #595 , #596, #617, #618 which are all related to Ansible Certifications. The changes needed for certification minimally impacted modules such that:
- All the prior doc meta that used M(zos_module_name) are not using link style L(zos_module_name,./zos_module_name.html) in an effort to correct the incorrect linking one uses the fully qualified collection name (FQCN). Using the FQCN does not generate a link in our documentation, it just makes it bold, this should be investigated in another git issue. 
- Updates some of the source defaults to match documentation defaults to avoid using ignore<version>.txt entries. 
- Also initializes a few variables that were not in the code
- zos_mount had a default "" (empty string) , could not find a reason for this so I updated the code 
- zos_mvs_raw had a global variable that was only used locally, it was updated to meet certification 
- One code change driven by the defaults and certification to note is the zos_data_set a mutually exclusive check was disabled but according the the prior commented sections should not pose any issues (space_type and name - were  prior disabled. I think the mutual exclusion logic is good to have enabled once again, it poses no issue for module execution but does allow defaults to appear in the non batch block of the response when a user is using zos_data_set batch mode. We would need to create an issue and look at our updating our better arg parse or updating the module code such that if it detects bach mode to null out the non-batch defaults. 
```
        mutually_exclusive=[
            ["batch", "name"],
            # ["batch", "state"],
            # ["batch", "space_type"],
            # ["batch", "space_primary"],
            # ["batch", "space_secondary"],
```

Non certification changes included:
- updates the Make file tooling used by development and introduces some new scripts in a `scripts/` directory.
- disabling our test clean up which means a manual clean up on the test system, i noticed in some cases the jobs were cleaned up a bit to fast and I could not track down the issues, utimately this is good to have so I will add it back in another pull request but a better solution needs to be in place that `adhoc.all.command(cmd="opercmd '$PJ(*)'")` not occur always and when a failure happens that it not be done, possibly this logic be added to all tear down tests, ill create another issue for this. 
- Minor test updates made such as adding wait times to jobs, changing the evaluation of how time is calculated or timed jobs using wait_time_s, ensuring data sets don't exist by adding zos_data_set absent and replace and some pprints that are a bit verbose now but needed to diagnose random issues that later can be removed that are likely a result of the target system response time. 
- removal of requirements.txt, this logic is now in our pipeline
- updated and new ignore files for certification, trying to keep them as small as possible as they are included with collections and reviewed by the team during an upload. Would rather manage issues in code over an exclusion in the ignore file.

##### ISSUE TYPE
- Feature Pull Request


ddimatos:
image

ddimatos:
I understand a large PR like this is not ideal, it tends to happen when getting into release tasks as many externals impact a change such as certification.

richp405 (approved):
Looks good to me... 2 yaml and 1 ini file need endings at some future point, but that won't stop a build or run.

PR #628, opened 2023-02-06
Updates the makefile and related scripts for the latest mounts

ddimatos:
##### SUMMARY
This update supports the latest mounts we have for use with our zoau and python installations. I appear to be the only one maintaining this and using this and there is no impact to collection content thus this pull request can be merged without necessarily regression. 

Changes include lexicographical order, more comments, new mount scripts.


ddimatos:
Bypassing and auto merging so that this can be cherry picked into 1.5 to help with release tasks

Issue #390, opened 2022-08-30
[Bug] [zos_operator] Module documentaion examples are incorrect


adn1107:
- name: Execute an operator command to show active jobs
  zos_operator:
    cmd: 'd u,all'

- name: Execute an operator command to show active jobs with verbose information
  zos_operator:
    cmd: 'd u,all'
    verbose: true

- name: Execute operator command to show jobs, waiting up to 5 seconds for response
  zos_operator:
    cmd: 'd u,all'
    wait_time_s: 5
    wait: false

- name: Execute operator command to show jobs, always waiting 7 seconds for response
  zos_operator:
    cmd: 'd u,all'
    wait_time_s: 7
    wait: true

the command should be 'd a,all' not 'd u,all'

ddimatos:
Moving to Q4 - iter3 as a result of #376 taking longer than expected

ddimatos:
Thank you @adn1107

PR #648, opened 2023-02-22
Update the module with the correct examples and descriptions

ddimatos:
##### SUMMARY
Fixes zos_operator module examples, fixes issue #390 

##### ISSUE TYPE

- Docs Pull Request


ddimatos:
No code change, no regression being run.

ketankelkar (approved):
lgtm -- good to see example docs that match up with module changes, also good to see different types of examples (display unit and display jobs rather than "d u,all" for all examples

Issue #428, opened 2022-09-21
[Enhancement][1] Support Jinja templates in modules


ddimatos:
### Enhancement description

Jinja2 templates can provide a powerful way to reuse `src` content, imagine you have some JCL or other `src` that reusable if  there was a way to replace the variables yet the src remains consistent. 

Today our modules do not accept Jinja as part of the `src` e.g. `src: my.jcl.j2` where JCL might look like:

my.jcl.j2:
```
//HELLO    JOB (T043JM,JM00,1,0,0,0),'HELLO WORLD - JRM',CLASS=R,
//             MSGCLASS=X,MSGLEVEL=1,NOTIFY=S0JM
//STEP0001 EXEC PGM=IEBGENER
//SYSIN    DD DUMMY
//SYSPRINT DD SYSOUT=*
//SYSUT1   DD *
{{ message }}
/*
//SYSUT2   DD SYSOUT=*
//
```

and your playbook you have:
```
---
- hosts: zvm
  collections:
    - ibm.ibm_zos_core
  gather_facts: no
  environment: "{{ environment_vars }}"
  vars:
      message: "Hello World"

  tasks:
     - name: submit a job
        zos_job_submit:
        src: my.jcl.j2
        location: LOCAL
       register: result
```

Now the JCL would have the vars substituted.

The community has a `template` module dedicated to taking *.j2 and `copy` for files, I don't know if it makes sense for us model after that given several of our modules could benefit from templating, those being `zos_copy`, `zos_job_submit` and  maybe `zos_fetch` to name a few. I think this issue needs to also perform a bit of research before developing. 

There is a decent introductory blog on this topic [here](https://www.linuxtechi.com/configure-use-ansible-jinja2-templates/)
and maybe also review to better understand the [difference between the community copy and template modules](https://dev.to/koh_sh/difference-between-ansible-s-copy-and-template-module-41e6).

### Ansible module

zos_copy, zos_fetch, zos_job_submit

PR #667, opened 2023-03-13
Enhancement/428/jinja2 template support

rexemin:
##### SUMMARY
Enhances the collection to add support for Jinja2 templates in `zos_copy` and `zos_job_submit`.

Addresses issues:
#428 
#487 
#488 
#698 
#699 
#710

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_copy, zos_job_submit, and a new module util named template.


plugins/modules/zos_copy.py
```diff
@@ -2673,6 +2676,30 @@ def main():
                     sms_management_class=dict(type="str", required=False),
                 )
             ),
+            use_template=dict(type='bool', default=False),
```
richp405, line 2679: Shouldn't this have a 'required' entry in it?

rexemin: Are you asking because of the possibility that a user may set options in `template_parameters` while ignoring `use_template`? It may need a `required` entry, I need to verify that

plugins/action/zos_copy.py
```diff
@@ -150,14 +160,59 @@ def run(self, tmp=None, task_vars=None):
                             dict(src=src, dest=dest, changed=False, failed=True)
                         )
                         return result
+
+                    if use_template:
+                        try:
```
fernandofloresg, line 165: I think try block should only be needed when using template not in variable assignment. I suggest to use the block from line 172 to line 179.

plugins/action/zos_copy.py
```diff
@@ -150,14 +160,59 @@ def run(self, tmp=None, task_vars=None):
...
+                        for f in files
                     )
                 else:
                     if mode == "preserve":
                         task_args["mode"] = "0{0:o}".format(
                             stat.S_IMODE(os.stat(src).st_mode)
                         )
+
+                    if use_template:
+                        try:
```
fernandofloresg, line 196: same comment as above.

plugins/action/zos_job_submit.py
```diff
@@ -32,6 +35,14 @@ def run(self, tmp=None, task_vars=None):
             return result
 
         module_args = self._task.args.copy()
+
+        use_template = _process_boolean(module_args.get("use_template"))
+        if use_template and module_args.get("location") != "LOCAL":
+            result["failed"] = True
```
fernandofloresg, line 41: If result is a dict consider using .update() function when updating multiple lines. Debate me on this, is just a comment about visuals. You can ignore this

plugins/action/zos_job_submit.py
```diff
@@ -94,6 +105,28 @@ def run(self, tmp=None, task_vars=None):
 
             tmp_src = self._connection._shell.join_path(tmp, "source")
 
+            if use_template:
+                try:
+                    template_parameters = module_args.get("template_parameters", dict())
+                    encoding = module_args.get("encoding", dict())
+
+                    renderer = template.create_template_environment(
```
fernandofloresg, line 113: Same comment as in zos_copy action plugin.

plugins/doc_fragments/template.py
```diff
@@ -0,0 +1,116 @@
...
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
+options:
+  use_template:
+    description:
+      - Whether the module should treat C(src) as a Jinja2 template and
+        render it before continuing with the rest of the module.
+      - Only valid when C(src) is a local file or directory.
+      - All variables defined in inventory files, vars files and the playbook
```
fernandofloresg, line 27: Do you know what is the order of precedence when getting the variables from different source? Might be worth it adding into the docs.

plugins/doc_fragments/template.py
fernandofloresg, line 1: Great that you found out how to import docs into other modules. Super useful!

plugins/module_utils/encode.py
fernandofloresg, line 1: Beware of this addition, I've seen this function being detected as missing of new addition in git prs. Just check this was not duplicated after merge.

plugins/doc_fragments/template.py
```diff
@@ -0,0 +1,116 @@
...
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
+options:
+  use_template:
+    description:
+      - Whether the module should treat C(src) as a Jinja2 template and
```
fernandofloresg, line 24: Do you know which version of Jinja2 is supported? do we select it or are dependent on python version?

tests/functional/modules/test_zos_copy_func.py
fernandofloresg, line 1: Do we need a test for copying with different encodings? I mean, does it even make sense, the controller most probably will be a unix/linux system with utf-8 or similar, and will update the template visually. But maybe if the file came from a different z/OS system with some strange codepage could make sense.

fernandofloresg (changes requested):

plugins/doc_fragments/template.py
rexemin: The[ Ansible documentation has the complete hierarchy](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) the engine uses. I'll add a note about it and link it in the doc fragment.

plugins/doc_fragments/template.py
rexemin: I don't know for sure which version of Jinja is installed by each version of Ansible, but since it's a dependency for the engine, we don't really control it.

plugins/modules/zos_copy.py
rexemin: I think I misunderstood your comment yesterday. If I remember correctly, arguments need to have either a `required` entry or a `default` one. By using `default` the engine infers `required=False`.

richp405: That's what I was asking about!  Okay, this looks okay to me.

richp405 (approved):

plugins/action/zos_job_submit.py
rexemin: I don't think I need to debate you, I browsed other modules and all of them use `update` when updating multiple items in a dictionary. Thanks for bringing it up!

tests/functional/modules/test_zos_copy_func.py
rexemin: I could add a test for templates encoded in ISO8859-1, which is the other most common encoding.

fernandofloresg: What do you think about this? @ketankelkar @ddimatos

AndreMarcel99 (commented):
I really like the code is too clean and also I understand well how it works.

plugins/action/zos_copy.py
rexemin: I moved down the try block. It still includes the creation and assigment of the `TemplateRenderer` object because its `__init__` method can throw two exceptions.

fernandofloresg: Got it!

plugins/action/zos_copy.py
fernandofloresg: Not needed as "I moved down the try block. It still includes the creation and assigment of the TemplateRenderer object because its __init__ method can throw two exceptions."

fernandofloresg (approved):
Looks ok to me. Would have preferred to split it into multiple PRs, each for a module for easier review.

rexemin:
Screenshot 2023-06-26 at 15 42 38
Screenshot 2023-06-26 at 15 42 56

plugins/module_utils/template.py
```diff
@@ -0,0 +1,336 @@
...
+            ))
+
+        try:
+            temp_template_dir = tempfile.mkdtemp()
+        except FileExistsError as err:
+            raise FileExistsError("Unable to create directory for rendered templates: {0}".format(
+                to_native(err)
+            ))
+        except PermissionError as err:
+            raise PermissionError("Error while trying to access temp directory for templates: {0}".format(
```
ddimatos, line 236: Very small, is it `temp` , `tmp` just play it safe and say temporary. Python does have precedence rules for how it finds a temporary directory to write into so its not always /tmp because i have had to instruct users on how to override this so that our modules an write into a bigger directory, basically saying maybe play it safe and if it makes sense, just say 'temporary'

plugins/module_utils/template.py
```diff
@@ -0,0 +1,336 @@
...
+            temp_parent_dir = tempfile.mkdtemp()
+            last_dir = os.path.basename(self.template_dir)
+            temp_template_dir = os.path.join(temp_parent_dir, last_dir)
+            os.makedirs(temp_template_dir, exist_ok=True)
+        except FileExistsError as err:
+            raise FileExistsError("Unable to create directory for rendered templates: {0}".format(
+                to_native(err)
+            ))
+        except PermissionError as err:
+            raise PermissionError("Error while trying to access temp directory: {0}".format(
```
ddimatos, line 293: Same comment here regarding `temp`

ddimatos: And just below this.

ddimatos (changes requested):
@rexemin overall this really looks great, its quite clean the test cases were creative, this as new territory for us and you did a nice job. I did leave a few comments, most of them are simply doc and ok if want to pass on them but did have a couple of points /questions in the code you should have a look at and let me know if its something worth changing.

ddimatos:
One additional comment, I don't think i saw any new return values coming back to the module, is there a need for any, such as `use_template` that might help us debug issues in the future? Maybe we don't need it.

plugins/action/zos_copy.py
```diff
@@ -150,14 +160,61 @@ def run(self, tmp=None, task_vars=None):
...
+                            renderer = template.create_template_environment(
+                                template_parameters,
+                                src,
+                                template_encoding
+                            )
+                            template_dir, rendered_file = renderer.render_file_template(
+                                os.path.basename(src),
+                                task_vars.get("vars", dict())
+                            )
+                        except Exception as err:
```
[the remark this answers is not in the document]
rexemin: I'll add a cleanup just in case. The method can throw a variety of exceptions, but there are two that could appear while writing a new file and leave incomplete files.

plugins/action/zos_job_submit.py
```diff
@@ -94,6 +107,29 @@ def run(self, tmp=None, task_vars=None):
...
+                    )
+                except Exception as err:
+                    result["msg"] = to_text(err)
+                    result["failed"] = True
+                    result["changed"] = False
+                    result["invocation"] = dict(module_args=module_args)
+                    return result
+
+                source_full = rendered_file
+
```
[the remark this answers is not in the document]
rexemin: Yes, thanks for catching this.

rexemin:
I don't think we need a new return value, the task will have the value for 'use_template' whenever the user adds `-vvv` or `-vvvv`

plugins/action/zos_copy.py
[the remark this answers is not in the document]
ddimatos: Thank you, always great to see your willingness to work with me.

plugins/action/zos_copy.py
```diff
@@ -187,6 +244,10 @@ def run(self, tmp=None, task_vars=None):
             task_vars=task_vars,
         )
 
+        # Erasing all rendered Jinja2 templates from the controller.
+        if template_dir:
+            shutil.rmtree(template_dir)
```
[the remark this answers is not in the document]
ddimatos: Thank you

ddimatos:
This is pretty much ready for merge (I resolved the ones I saw updated) but wondering on one unresolved comment from @fernandofloresg and then the other ones from myself if they are going to be passed on or corrected. @rexemin

ddimatos (approved):
Looks great, so excited to see this and its well done, thank you so much @rexemin

Issue #574, opened 2022-11-30
[Bug] [zos_find] zos_find return the filtered_data_sets list once BGYSC1103E (dataset not found) returned by dls.


snowind:
### Bug description

zos_find: vall versions

Reproduce the problems with playbooks:

On z/OS, PDS "IMSTEST.USER.A" contains 2 members, "IMSTEST.USER.B" also cantains 3 members.  "IMSTEST.DS.NOTEXIST " is not existed. 

Test case #1: 
- name: Find all members in a given list of PDS patterns
  zos_find:
    patterns: '.*'
    pds_patterns:
      - IMSTEST.DS.NOTEXIST 
      - IMSTEST.USER.A
      - IMSTEST.USER.B
   register: pds_members

Test case #2: 
- name: Find all members in a given list of PDS patterns
  zos_find:
    patterns: '.*'
    pds_patterns:
      - IMSTEST.USER.A
      - IMSTEST.DS.NOTEXIST 
      - IMSTEST.USER.B
   register: pds_members

Tese case #1 failed and matched 0 returned. 
Test case #2 return the members in IMSTEST.USER.A only, no members in the "IMSTEST.USER.B" returns. 

In the zos_find logic, once the "BGYSC1103E" message is returned from dls command, the loop is ended and  the filtered_data_sets is returned, that explains why no remaining PDS in the pds_patterns list  is checked, once there is one PDS is not exist.

filtered_data_sets = dict(ps=set(), pds=dict(), searched=0)
    patterns = pds_paths or patterns
    for pattern in patterns:
        rc, out, err = _dls_wrapper(pattern, list_details=True)
        if rc != 0:
            if "BGYSC1103E" in err:
                return filtered_data_sets

### Ansible version

```shell
ansible 2.10.17
  config file = None
  configured module search path = ['/Users/ZhouHui/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/ZhouHui/opt/anaconda3/lib/python3.8/site-packages/ansible
  executable location = /Users/ZhouHui/opt/anaconda3/bin/ansible
  python version = 3.8.3 (default, Jul  2 2020, 11:26:31) [Clang 10.0.0 ]
```

### IBM z/OS Ansible core Version

v1.3.1

### IBM ZOAU version

v1.1.1

### z/OS version

z/OS 2.4 

### Ansible module

zos_find

richp405:
[...]

PR #668, opened 2023-03-14
574find

richp405:
##### SUMMARY
zos_find stops searching when given a non-existant pds_data set

Addresses #574 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_find: given a pds_patterns list value, the search stops after finding the first data set that does. NOT exist.

##### ADDITIONAL INFORMATION
I've created a playbook to manually test this... content me if you need it.

```
zos_find:
        patterns: '.*'
        pds_patterns:
          - IMSTESTU.DS.NOTEXIST
          - IMSTESTU.DDL.PSB.SOURCE
          - IMSTESTU.DDL.DBD.SOURCE
      register: pds_memberspaste below

```

image

### DIFF
--- a/plugins/modules/zos_find.py
+++ b/plugins/modules/zos_find.py
@@ -336,7 +336,8 @@ def data_set_filter(module, pds_paths, patterns):
         rc, out, err = _dls_wrapper(pattern, list_details=True)
         if rc != 0:
             if "BGYSC1103E" in err:
-                return filtered_data_sets
+                # return filtered_data_sets
+                continue
 
             module.fail_json(
                 msg="Non-zero return code received while executing ZOAU shell command 'dls'",
--- a/.github/ISSUE_TEMPLATE/collaboration_issue.yml
+++ b/.github/ISSUE_TEMPLATE/collaboration_issue.yml
@@ -1,7 +1,7 @@
-name: Report a bug
-description: Request that a bug be reviewed. Complete all required fields.
-title: "[Bug] <title> "
-labels: [Bug]
+name: Request a Collaboration
+description: Request collaboration with a member of this team. Complete all required fields.
+title: "[Collaboration] <title> "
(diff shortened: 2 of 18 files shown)


rexemin (approved):
Straightforward fix, thanks for adding the functional test for it too.

plugins/modules/zos_find.py
```diff
@@ -336,7 +336,8 @@ def data_set_filter(module, pds_paths, patterns):
         rc, out, err = _dls_wrapper(pattern, list_details=True)
         if rc != 0:
             if "BGYSC1103E" in err:
-                return filtered_data_sets
+                # return filtered_data_sets
+                continue
```
ddimatos, line 340: At first I agreed with this, but then it appears if when we do encounter a BGYSC1103E we will continue and go  onto hit the fail_json code.It sort of looks like we don't want to fail at this point and instead review the remaining results to see if we can find something.

Your test case seems to cover the scenario below but still feel there is something off.

In the users example here, returned 0 because we failed at the first one, when we could have returned 2 found.
```
pds_patterns:
- IMSTEST.DS.NOTEXIST
- IMSTEST.USER.A
- IMSTEST.USER.B
```
and here 1 was returned when it could have been 2 as well.
```
- IMSTEST.USER.A
- IMSTEST.DS.NOTEXIST
- IMSTEST.USER.B
```

ddimatos (changes requested):

plugins/modules/zos_find.py
ddimatos: Ah, you are right, sorry about that @richp405, the loop will continue, I think i had done 1 too many code reviews that day.

ddimatos (approved):
Looks good @richp405 thanks for the time to get these bugs crushed :)

Issue #358, opened 2022-07-22
[Feature][ibm_zos_data_set][L2] Support disposition share (DISP=SHR)


ddimatos:
##### FEATURE

Module ibm_zos_data_set could benefit from supporting the disposition share (DISP=SHR) so that a member can be updated when a PDS/E has been open but a started task and will not be released. This would allow the module to force a change. 

This will require that [ZOAU support DISP=SHR in ZOAU utility ](https://www.ibm.com/docs/en/zoau/1.2.0?topic=reference-mrm)`mrm`.

An example of this is a member being deleted using zos_data_set module:
```
    - name: "Delete the member(s) from PDS(s)"
      zos_data_set:
        name: "HLQ.MLQ.LLQ(APP1)"
        state: absent
        type: MEMBER
```

Results in:
```
"DatasetMemberDeleteError('An error occurred during deletion of data set member\"HLQ.MLQ.LLQ(APP1)\". RC=1')", "names": ["HLQ.MLQ.LLQ(APP1)"]}
````

Source of interest in the module:
```
        Raises:
            DatasetMemberDeleteError: When data set member deletion fails.
        """
        rc = datasets.delete_members(name)
        if rc > 0:
            raise DatasetMemberDeleteError(name, rc)
```

Relates to #357 
IBM Internal JIRA issue ZOAU-9264
IBM internal discussion reference info `CSLEJ8VGV/p1658245204937069`

ketankelkar:
Finally got a test case set up to simulate the data-set-in-use condition to ensure we maintain as complete of a functional test suite as possible. I expect this test case set up will be beneficial to a number of other [enhancements](https://github.com/ansible-collections/ibm_zos_core/issues/357) for zos_blockinfile, [zos_lineinfile](https://github.com/ansible-collections/ibm_zos_core/issues/356) and potentially additional zos_data_set use cases.

Finally ready to begin feature development

ketankelkar:
Development is complete. Local test cases seem to be running just fine, PR is open, pending a successful pipeline run (ran into pipeline performance issues last night)

PR #718, opened 2023-03-31
zos_data_set - add force parameter to enable member delete via disp shr

ketankelkar:
##### SUMMARY

leverages new ZOAU capability `mrm -f`to open a pdse in disp=shr mode and facilitate member delete while pdse is in use by another process.

Addresses issue #358 

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/plugins/modules/zos_data_set.py
+++ b/plugins/modules/zos_data_set.py
@@ -44,6 +44,10 @@ options:
         If I(state=absent) and the data set does exist on the managed node,
         remove the data set, module completes successfully with I(changed=True).
       - >
+        If I(state=absent) and I(type=MEMBER) and I(force=True), the data set
+        will be opened with I(DISP=SHR) such that the entire data set can be
+        accessed by other processes while the specified member is deleted.
+      - >
         If I(state=absent) and I(volumes) is provided, and the data set is not
         found in the catalog, the module attempts to perform catalog using supplied
         I(name) and I(volumes). If the attempt to catalog the data set catalog is successful,
@@ -253,9 +257,11 @@ options:
         delete operation which results in the data set you are updating to be
         simultaneously updated by others.
       - This is helpful when a data set is being used in a long running process
-        such as a started task and you are wanting to update or read.
-      - The C(-f) option enables sharing of data sets through the disposition
-        I(DISP=SHR).
+        such as a started task and you are wanting to delete a member.
+      - The I(force=True) option enables sharing of data sets through the
+        disposition I(DISP=SHR).
+      - The I(force=True) only applies to data set members when I(state=absent)
+        and I(type=MEMBER).
     type: bool
     required: false
     default: false
@@ -283,6 +289,11 @@ options:
           - >
             If I(state=absent) and the data set does exist on the managed node,
             remove the data set, module completes successfully with I(changed=True).
+          - >
+            If I(state=absent) and I(type=MEMBER) and I(force=True), the data
+            set will be opened with I(DISP=SHR) such that the entire data set
+            can be accessed by other processes while the specified member is
+            deleted.
           - >
             If I(state=absent) and I(volumes) is provided, and the data set is not
             found in the catalog, the module attempts to perform catalog using supplied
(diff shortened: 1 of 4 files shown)


ketankelkar:
image

plugins/modules/zos_data_set.py
```diff
@@ -247,6 +247,18 @@
         that is not available, then the value C(TMPHLQ) is used.
     required: false
     type: str
+  force:
+    description:
+      - Specifies that the data set can be shared with others during a member
+        delete operation which results in the data set you are updating to be
+        simultaneously updated by others.
+      - This is helpful when a data set is being used in a long running process
```
ddimatos, line 255: I think this doc came from zos_blockinfile which is slightly different in its use with `force` where that force is applied to the data set in its entirety for read and write. Where here ware only supporting member deletes as DISP=SHR.

So maybe alter this:
```
      - This is helpful when a data set is being used in a long running process
        such as a started task and you are wanting to update or read.
```
to 
```
      - This is helpful when a data set is being used in a long running process
        such as a started task and you delete a member.

File: plugins/modules/zos_data_set.py [hunk 2]
```diff
@@ -247,6 +247,18 @@
...
+        such as a started task and you are wanting to update or read.
+      - The C(-f) option enables sharing of data sets through the disposition
```
ddimatos, line 257: I did fix the zos_blockfile doc a day ago, this should read:
```
      - The I(force=true) option enables sharing of data sets through the disposition
         I(DISP=SHR).
      - The I(force=true) only applies to data set members when I(state=absent) and
         I(type=MEMBER).
```

I added an extra line to highlight even more this support is just for data set members

plugins/modules/zos_data_set.py
```diff
@@ -467,6 +479,18 @@
         type: bool
         required: false
         default: false
+      force:
+        description:
+          - Specifies that the data set can be shared with others during a member
```
ddimatos, line 484: I think this doc came from zos_blockinfile which is slightly different in its use with `force` where that force is applied to the data set in its entirety for read and write. Where here ware only supporting member deletes as DISP=SHR.

So maybe alter this:
```
      - This is helpful when a data set is being used in a long running process
        such as a started task and you are wanting to update or read.
```
to 
```
      - This is helpful when a data set is being used in a long running process
        such as a started task and you delete a member. 
```


I did fix the zos_blockfile doc a day ago, this should read:

      - The I(force=true) option enables sharing of data sets through the disposition
         I(DISP=SHR).
      - The I(force=true) only applies to data set members when I(state=absent) and
         I(type=MEMBER).

I added an extra line to highlight even more this support is just for data set members

plugins/modules/zos_data_set.py
```diff
@@ -247,6 +247,18 @@
         that is not available, then the value C(TMPHLQ) is used.
     required: false
     type: str
+  force:
```
ddimatos, line 250: I think we might want to mention the usage of state and force in the state doc, eg there in the doc they describe how the two together behave 

```If state=present and replace=True .....```

Where we might want:
```
If state=absent and force=True, force will only apply to data set members and the data set will be opened with I(DISP=SHR) such that the entire data set can be accessed by other processes while the member is deleted. 
```

ddimatos (changes requested):
Nice work @ketankelkar , much larger work item than I had thought, especially the test case.Some minor change requests.

ddimatos:
I had no place to tag this ,but also module_utils/data_sets.py copyright year should be updated. Its minor but easy enough to do.

plugins/modules/zos_data_set.py
ketankelkar: how embarrassing of me, you're right, i did lift it from that doc but i thought i'd modified it

plugins/modules/zos_data_set.py
ketankelkar: good point, i'm modifying your suggestion to "if `type =member` ..." rather than having "only apply to data set members" -- i think that's more explicit and quicker to read/understand

plugins/modules/zos_data_set.py
ketankelkar: i kept the "are wanting to..." portion from the copied section unless you specifically omitted it

plugins/modules/zos_data_set.py
ketankelkar: i wonder if there's a path to using document fragments here since we have 2 of the same docs except one is nested under "batch"

ketankelkar:
func tests still passing
image

plugins/modules/zos_data_set.py
ddimatos: There is a path to use doc fragements, we can do this , and have been wanting to for idk, couple of yrs now :) , I share more on this in one of our meetings.

ddimatos (approved):
Looks good now, thanks so much for the changes @ketankelkar

PR #721, opened 2023-04-03
Add positional wildcard support to module zos_job_query

### LINKED ISSUE #323: [Feature] [zos_job_query]More asterisk support for z/OS ansible modules
I have a playbook using zos_job_query to query jobname "CPY*RS*"

      zos_job_query:
        job_name: "CPY*RS*"

but it failed and report the following error:
Value of unknown type: <class 'RuntimeError'>, Failed to validate the job name: CPY*RS*\n

      zos_job_query:
        job_name: "CPY*"

Describe the solution you'd like:
It looks like only support asterisk * as the suffix of the jobname now.
It would be better to have more than 1 asterisk support and even other wildcard support.

richp405:
##### SUMMARY
Adds embedded and multiple asterisk support to zos_job_query
Fixes #323 
##### ISSUE TYPE
Feature pull request AND bugfix pull request

##### COMPONENT NAME
zos_job_query showed the problem.  Actual issue was is search condition of job/_get_job_status

##### ADDITIONAL INFORMATION
If the job id or name is more than an asterisk (example look for "JOB003*), or multiple asterisk (J*003*)
This caused limited results and potential failures, because the asterisks didn't work correctly
This has been changed to use fnmatch (note: this IS case insensitive)

examples:
 - name: Check for I*I* embedded star
      zos_job_query:
        job_name: I*I*
      register: result

 - name: Check for STC*3* job embedded star
      zos_job_query:
        job_id: STC*3*
      register: result

```paste below

```


plugins/modules/zos_job_query.py
```diff
@@ -208,18 +208,26 @@ def run_module():
...
     except Exception as e:
         module.fail_json(msg=e, **result)
     result["jobs"] = jobs
     module.exit_json(**result)
 
 
+# validate_arguments rturns a tuple, so we don't have to rebuild the job_name string
 def validate_arguments(params):
     job_name_in = params.get("job_name")
+    job_name_final = job_name_in
```
fernandofloresg, line 227: job_name_final appears to not be used after this assignment

plugins/modules/zos_job_query.py
fernandofloresg, line 1: variable names o, m, n make difficult for maintainers to read over the code, I had to read over it multiple times to get the idea. Could you give them more descriptive names?

tests/functional/modules/test_zos_job_query_func.py
fernandofloresg, line 1: This test file seems missing cases, is there a way to add a positive test case? Running a couple of jobs and looking for them?

fernandofloresg (changes requested):
Overall change looks good to me, comments were added in the review

changelogs/fragments/323-zos-job-query-handle-multiple-wildcards.yml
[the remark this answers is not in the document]
richp405: Added.

plugins/modules/zos_job_query.py
richp405: forgot to remove it... was part of a test set.

plugins/modules/zos_job_query.py
```diff
@@ -229,32 +237,58 @@ def validate_arguments(params):
...
+                        o = job_id_short
+
+            if not m and not o:
                 raise RuntimeError("Failed to validate the job id: " + job_id)
     else:
         raise RuntimeError("Argument Error:Either job name(s) or job id is required")
     if job_id and owner:
         raise RuntimeError("Argument Error:job id can not be co-exist with owner")
 
+    # return job_name_final, id, owner;
```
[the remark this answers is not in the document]
richp405: yep

plugins/modules/zos_job_query.py
richp405: Changed to test_basic, test_star and test_result.  Hope that clears it up... Sorry for the 'letters'... it's a bad habit from my ancient c days.  Thanks for pointing it out.

tests/functional/modules/test_zos_job_query_func.py
richp405: The original case in there, which was 'list all jobs' was the positive case... I added an addition test of wildcards agains job_id in addition to job_name.  Would that work, or do you think I should run and test a specific job (since that's done in other tests)?

plugins/modules/zos_job_query.py
fernandofloresg: I'd imagine. Thanks for changing it but comments still have the letters

tests/functional/modules/test_zos_job_query_func.py
fernandofloresg: I would say yes to make sure we find them. Btw ran the Jenkins pipeline and test "test_zos_job_query_multi_wildcards_func" failed :

Screenshot 2023-04-03 at 3 24 23 PM

plugins/modules/zos_job_query.py
richp405: ack!!! I fixed the code and skipped the comments.

tests/functional/modules/test_zos_job_query_func.py
richp405: When copying the test, I didn't rename it.  I'll fix and re-run the afternoon.

plugins/modules/zos_job_query.py
```diff
@@ -208,53 +208,86 @@ def run_module():
...
+                ix = job_name_in.find("*")
+                if ix >= 0:
+                    job_name_short = job_name_in[0:ix + 1]
+                    test_result = job_name_pattern.search(job_name_short)
+                    if not test_result:
+                        test_result = job_name_pattern_with_star.search(job_name_short)
+
+            # so now, fail if neither m, n, or o=m/n(short) found a match
+            if not test_result:
+                raise RuntimeError("Failed to validate the job name: " + job_name_in + " ix was " + ix + " short was " + job_name_short)
```
ddimatos, line 256: @richp405 - is error meant for users , seems very specific to the code, eg `ix was` , users will not know what `ix` is.  I think just returning the value they gave us is enough, no need to share with them the logic of short and ix. 

Wondering if it should be "Unable to find the job name: {0}".format(str(job_name_in))

plugins/modules/zos_job_query.py
```diff
@@ -208,53 +208,86 @@ def run_module():
...
-                raise RuntimeError("Failed to validate the job name: " + job_name_in)
+            test_basic = job_name_pattern.search(job_name_in)
+            test_star = job_name_pattern_with_star.search(job_name_in)
+            # logic twist: o must be non-null value from m or n
+            test_result = test_basic
+            if test_star:
+                test_result = test_star
+
+            job_name_short = "unused"
+            # if neither m nor n were non-null, check if the string needed to be truncated to the first *
```
ddimatos, line 245: m/n comments don't map to the updated code , its minor just might be helpful for someone in the future.

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -33,3 +33,22 @@ def test_zos_job_query_func(ansible_zos_module):
...
+def test_zos_job_query_multi_wildcards_func(ansible_zos_module):
+    hosts = ansible_zos_module
+    results = hosts.all.zos_job_query(job_id="STC*3*")
+    pprint(vars(results))
+    for result in results.contacted.values():
+        assert result.get("changed") is False
+        assert result.get("jobs") is not None
+
+# test to show multi wildcard in Job_name query won't crash the search
+def test_zos_job_query_multi_wildcards_func(ansible_zos_module):
```
ddimatos, line 48: I don't see  a job submitted thus we are relying on the ones on the system which works for now but not for portability in the future, but this is easy enough to fix. What caught my eye was the  `job_name="JOB*1*"` seems to be a JOB ID, i logged on to my system to find a job name starting with JOB but all I found was Job IDs. Could this be an error?

ddimatos:
We should update our doc further highlighting the now supported wildcard placement.

ddimatos (commented):
We should update our doc further highlighting the now supported wildcard placement.

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -33,3 +33,22 @@ def test_zos_job_query_func(ansible_zos_module):
     for result in results.contacted.values():
         assert result.get("changed") is False
         assert result.get("jobs") is not None
+
+
+# test to show multi wildcard in Job_id query won't crash the search
```
ddimatos, line 38: Are we able to support job names that start with * , what would that look like if we do for very busy system?

ddimatos (changes requested):
I don't know where to put these next two comments as there is no change for me to comment where I want so I will add them here.

1) We should update our doc further highlighting the now supported wildcard placement.
2) We should add some examples to the doc to highlight the capabilities now, some examples with Job Names, Job Ids wild card placement.

tests/functional/modules/test_zos_job_query_func.py
richp405: Yes... not as different as you might think: regardless of the mask, we ask for something like jls *, then filter the results in the python.  If it's job_id, we get a smaller list, up to the first asterisk; it it's job_name, we get them all.

tests/functional/modules/test_zos_job_query_func.py
richp405: I fixed, and expanded both cases to run a 'hello' jcl, then check for the results.  Haven't gotten through pipeline, as someone else is running it, but will try again at end of day

plugins/modules/zos_job_query.py
richp405: That's simpler and better.  Making the change.

tests/functional/modules/test_zos_job_query_func.py
richp405: Yes..Fernando mentioned this as well... the new tests run a copy of the hello jcl.  Job ID takes out 2 characters and uses asterisks for search, while job name search looks for "HE*L*".... That should go through the pipeline this evening.

richp405:
@ddimatos I didn't see a widget to reply to your comment about documentation, so I'm adding this at the end... I tweaked (minor changes) to the documentation at the top, with examples and the data type descriptions, to expand on the asterisk support.  I'll ping you once this passes the pipeline.

tests/functional/modules/test_zos_job_query_func.py
richp405: I called a function with un-named parameters.  Fixed:

image

ddimatos (approved):
Looks good to me @richp405 , we need one more reviewer and we are good to merge.

ddimatos:
The last regression for this was successful 
image

fernandofloresg (approved):
Looks good to me. Thanks for doing all the requested changes.

richp405:
One more edit, and through the pipeline

image

fernandofloresg:
Thanks @richp405

Issue #619, opened 2023-01-21
[Bug] [zos_copy] Mode set for files is applied to destination directory


rexemin:
### Bug description

When setting `mode` and copying the contents of a directory, the new permissions are applied to the destination directory too. This is a deviation from the behavior of the community module, where the permissions are only applied to the files that are getting copied.

Example task:
```yaml
tasks:
    - name: Copy remote text files.
      zos_copy:
        src: ./tests/
        dest: /test/
        mode: '0644'
        force: true
        remote_src: false
```

And where `/test/` has permissions `755` initially. Output from `ls -la` after using zos_copy:
```
drw-r--r--   2 BPXROOT  SYS1        8192 Jan 20 16:08 test
```

Output from `ls -la` after using the community copy module:
```
drwxr-xr-x   2 BPXROOT  SYS1        8192 Jan 20 16:06 test
```

### Contents of `ansible.cfg`

```YAML
[defaults]
forks = 25
# remote_tmp = /u/ansible/tmp
# remote_port = 22
# debug = True
# verbosity = 1

[ssh_connection]
pipelining = True

[colors]
verbose = green
```

### Contents of the inventory

```YAML
source_system:
  hosts:
    zvm:
      ansible_host: ***
      ansible_user: "omvsadm"
      ansible_python_interpreter: "/python3/usr/lpp/IBM/cyp/v3r8/pyz/bin/python3.8"
```

### Ansible version

```shell
ansible [core 2.11.12] 
  config file = /Users/iamorenosoto/Documents/IBM/Development/GitHub/553_554/playbooks/ansible.cfg
  configured module search path = ['/Users/iamorenosoto/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/iamorenosoto/.pyenv/versions/3.8.12/envs/gh-553-554/lib/python3.8/site-packages/ansible
  ansible collection location = /Users/iamorenosoto/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/iamorenosoto/.pyenv/versions/gh-553-554/bin/ansible
  python version = 3.8.12 (default, Jan 17 2022, 16:02:56) [Clang 13.0.0 (clang-1300.0.27.3)]
  jinja version = 3.1.2
  libyaml = True
```

### IBM z/OS Ansible core Version

v1.4.0

### IBM ZOAU version

v1.1.1

### z/OS version

2.3

### Ansible module

zos_copy

Issue #417, opened 2022-09-15
[Bug] [zos_blockinfile] Can quotes in content can be supported


ddimatos:
##### SUMMARY
This is a research item to identify the cases we might be missing.
Today in zos_blockinfle we support no quotes in the content block, in fact if we see them we strip them out (issue #303). 

In discussion it came up that possibly this might need to be supported:
```
- name: Update USS script
      zos_blockinfile:
        src: "/u/scripts/script.sh"
        insertafter: "EOF"
        block: |
             cat "//'{{ DS_NAME }}'"
             cat "//'{{ DS_NAME_2 }}'"
      register: result
    - name: Response
      debug:
        msg: "{{ result }}"
```

This could also easily apply to a data set of type PDS/E where a member might have content like wanting to append some `cat` statements. 
```
//BPXSLEEP JOB MSGCLASS=A,MSGLEVEL=(1,1),NOTIFY=&SYSUID,REGION=0M
//USSCMD EXEC PGM=BPXBATCH
//STDERR  DD SYSOUT=*
//STDOUT  DD SYSOUT=*
//STDPARM DD *
SH ls -la /;
sleep 30;
/*
// 
```

This could possibly impact ZOAU `dmod` dependency depending on the research. 
Review applicable ZOAU JIRA 9120

ddimatos:
It appears ZOAU 1.2.2 resolved the issue in their python api's for issue 9120, this work item now should evaluate that fix from ZOAU an then remove the logic in the zos_blockinfile module that strips quotes.

PR #723, opened 2023-04-04
zos_copy mode is applied to the destination directory, a deviation from the communtiy module behavior.

AndreMarcel99:
##### SUMMARY
Fixed bug #619  Mode set for files is applied to destination directory by change the logic of changing files

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
The new logic is one is not a folder so only a file the mode will change not everything to be copy by the module

Captura de pantalla 2023-04-03 a la(s) 19 34 02

### DIFF
--- a/changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
+++ b/changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
@@ -1,0 +1,3 @@
+minor_changes:
+- zos_copy - wasn enhanced to work as copy comunitty module for the destitny folder do not change mode 
+  
--- a/changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
+++ b/changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
@@ -1,3 +1,3 @@
 minor_changes:
-- zos_copy - wasn enhanced to work as copy comunitty module for the destitny folder do not change mode 
-  
+- zos_copy - Fixed a bug where the module would change the mode for a directory when copying into it the contents of another.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/723)
--- a/changelogs/fragments/417-can-quotes-in-content-can-be-supported.yml
+++ b/changelogs/fragments/417-can-quotes-in-content-can-be-supported.yml
@@ -1,3 +1,3 @@
 bugfixes:
-- The use of quotes for any sentence is now in use for the users in blockinfile module
-- The message for the user about a correct insertion includes the quotes
+- The use of quotes for any sentence it wasn't support in use for the users in blockinfile module
+- The message for the user about a correct insertion did not include the quotes
--- a/changelogs/fragments/417-can-quotes-in-content-can-be-supported.yml
+++ b/changelogs/fragments/417-can-quotes-in-content-can-be-supported.yml
@@ -1,0 +1,3 @@
+bugfixes:
+- The use of quotes for any sentence is now in use for the users in blockinfile module
+- The message for the user about a correct insertion includes the quotes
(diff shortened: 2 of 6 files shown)


changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+- zos_copy - wasn enhanced to work as copy comunitty module for the destitny folder do not change mode
```
rexemin, line 2: Couple of typos in this fragment. I would also describe the work done on zos_copy to be a bugfix, rather than an enhancement. Something like this:
Fixed a bug where the module would change the mode for a directory when copying into it the contents of another.

rexemin (commented):
Thank you so much for working on this issue. The change to zos_copy is pretty straightforward, just a couple of questions from me:
- Does the community module not change any permissions on directories? My guess is that it doesn't, so good catch on that
- You're mixing two bugfixes in this PR, did you open another PR for #417?

changelogs/fragments/619-Mode-set-for-files-is-applied-to-destination-directory.yml
AndreMarcel99: - Ok I will make the changes in the changelogs.
- https://github.com/ansible-collections/ibm_zos_core/issues/417 is already waiting for approval I do not know why I had here the solution too, another change
- And lastly I will verify about copy folder to folder and if it's need just add another verification

ddimatos (changes requested):
@AndreMarcel99 this looks good, i'll request changes since the changelog fragment has not been updated yet. 

My only other comment is if we needed a test case in place to prevent regression on this especially since this tied back to a support case, we would not want to relapse in function, although its not likely given the code changes is so small. 

At minimum at least show the output of the playbook in the issue/pr for us to see it works and some `ls` commands.

AndreMarcel99:
The playbook as:
- hosts : zvm
  collections : 
    - ibm.ibm_zos_core
  gather_facts: no
  vars:
    ZOAU: "/zoau/v1.2.2"
    PYZ: "/python2/usr/lpp/IBM/cyp/v3r2/pyz"
    source: ./tests/
    destt: /test/
  
  tasks:
    - name: Delete al files inside folder dest
      shell:  
        cmd: rm -r {{ destt }}
        
    - name: Create dest folder
      shell:
        cmd: mkdir {{ destt }}
      register: result
    
    - name: Adjust mode of dest folder
      shell:
        cmd: chmod 755 {{ destt }}
      register: result
    
    - name: Verify mode of destiny 
      shell:
        cmd: ls -la {{ destt }}
      register: result
    - name: Response
      debug:
        msg: "{{ result }}"
- name: Copy remote text files zos_copy.
      zos_copy:
        src: "{{ source }}"
        dest: "{{ destt }}"
        mode: 644
        force: true
        remote_src: false
    - name: Response zos_copy
      debug:
        msg: "{{ result }}"

    - name: See permisions on dest 
      shell:
        cmd: ls -la {{ destt }}
      register: result
    - name: Response permissions after
      debug:
        msg: "{{ result }}"
Captura de pantalla 2023-04-06 a la(s) 9 06 03
This is the results before the change.
After:
Captura de pantalla 2023-04-06 a la(s) 9 14 42

rexemin (approved):
Thanks for making the changes to the code, looks good. I reviewed the playbook you used and seems to work as intended now. IIRC there's already a testcase for permissions for files and directories.

One last thing: you could also use the [builtin `file` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html) from Ansible to create (by using `state: directory`) and delete directories (`state: absent`) on a managed node.

ddimatos:
@AndreMarcel99 thanks for sharing the playbook, @rexemin thanks for the code review as well, all looks good to me as well.

ddimatos (approved):
Thanks for the changes and good work @AndreMarcel99

Issue #678, opened 2023-03-20
[Bug] [zos_copy] File normalizing is making assumptions that could pose encoding mixup


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Are the dependencies a supported version?

- [X] The dependencies are supported.

### IBM Z Open Automation Utilities

v1.2.2

### IBM Enterprise Python

v3.9.x

### IBM z/OS Ansible core Version

v1.5.0

### ansible-version

v2.14.x

### z/OS version

v2.4

### Ansible module

zos_copy

### Bug description

This code [here](https://github.com/ansible-collections/ibm_zos_core/blob/8f9faec3dd41d0486b5d6c4d87dc3e2c8077ed3a/plugins/modules/zos_copy.py#L2434) which is in 1.4.1, 1.5.x and dev (1.6.x) is making some suspicious decisions. 

1) If the file say is in 1037 and it happens to be untagged it will be [set to 1047](https://github.com/ansible-collections/ibm_zos_core/blob/8f9faec3dd41d0486b5d6c4d87dc3e2c8077ed3a/plugins/modules/zos_copy.py#L2445) , this would be break any encoding conversions that might occur later on.
```
                if src_tag == "untagged":
                    src_tag = encode.Defaults.DEFAULT_EBCDIC_USS_CHARSET
```

2) If bullet (1) did occur then this case would be executed because 1037 is not in 1047 and while a new temp file is created in the system locale , we would be `iconv`'ing from: `1037` to:`1047` yet:
- the original source might have been 1047 (see bullet 1) 
- this is forced because 1037 is not in 1047 when it might have already been 1047, unnecessary conversion possibly causing issues.
```
                if src_tag not in encode.Defaults.DEFAULT_EBCDIC_MVS_CHARSET:
                    fd, converted_src = tempfile.mkstemp()
                    os.close(fd)

                    enc_utils.uss_convert_encoding(
                        new_src,
                        converted_src,
                        src_tag,
                        encode.Defaults.DEFAULT_EBCDIC_MVS_CHARSET
                    )
                    copy_handler._tag_file_encoding(converted_src, encode.Defaults.DEFAULT_EBCDIC_MVS_CHARSET)
                    new_src = converted_src
```

I propose to eliminate and rewrite this section:
```
                if src_tag not in encode.Defaults.DEFAULT_EBCDIC_MVS_CHARSET:
                    fd, converted_src = tempfile.mkstemp()
[...]

Issue #664, opened 2023-03-03
[Bug] [zos_copy] Sometimes the record length computation throws an encoding error


rexemin:
### Bug description

If a file has a different encoding other than IBM-1047, the function `get_record_length` can throw an encoding error when the file is opened and an invalid byte is read. This invalid byte corresponds to another encoding being used.

This bug was first encountered in an L2 support case, and as such, it will need to be addressed in following releases of the collection.

### Ansible version

```shell
ansible [core 2.11.12] 
  config file = /Users/iamorenosoto/Documents/IBM/Development/GitHub/553_554/playbooks/ansible.cfg
  configured module search path = ['/Users/iamorenosoto/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/iamorenosoto/.pyenv/versions/3.8.12/envs/gh-588/lib/python3.8/site-packages/ansible
  ansible collection location = /Users/iamorenosoto/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/iamorenosoto/.pyenv/versions/gh-588/bin/ansible
  python version = 3.8.12 (default, Jan 17 2022, 16:02:56) [Clang 13.0.0 (clang-1300.0.27.3)]
  jinja version = 3.1.2
  libyaml = True
```

### IBM z/OS Ansible core Version

v1.4.0, v1.5.0-beta.1

### IBM ZOAU version

v1.1.1

### z/OS version

2.3

### Ansible module

zos_copy

ddimatos:
This is  just adding to my cp37 comment, the link i provided I did not highlight the section i as eluding to:
```
I did notice that the DB2 V8 precompiler, with C and CPP, uses 37 by
default, not 1047. (But I'm used to ??( and ??), anyway. Mental note to
self - override DB2 precompiler step with CP 1140.)
```
We don't have to do anything but just thought it was an interesting data point.

PR #725, opened 2023-04-04
Bugfix/678/1.5.0 encoding normalization zos copy

rexemin:
##### SUMMARY
Fixes both #678 and #664 for 1.5.0, since both have to do with encoding issues in zos_copy.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_copy

##### ADDITIONAL INFORMATION
These issues deal with problems we've seen when users are trying to copy files that are not encoded in UTF-8 or IBM-1047, and even worse, when those files may have CRLF endings.

I added tests to verify that encoding normalization works for the following two cases:
1. When trying to compute the record length for a dataset.
2. When trying to get rid of newlines from files with CRLF endings.

### DIFF
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -926,15 +926,20 @@ class CopyHandler(object):
             {bool} -- True if the file uses CRLF endings, False if it uses LF
                       ones.
         """
+        # Python has to read the file in binary mode to not mask CRLF
+        # endings or enable universal newlines. If we used encoding="cp037",
+        # we would get '\n' as the line ending even when the file uses '\r\n'.
         with open(src, "rb") as src_file:
-            # readline() will read until it finds a \n.
-            content = src_file.readline()
+            # Reading the file in 1024-byte chunks.
+            content = src_file.read(1024)
 
-        # In EBCDIC, \r\n are bytes 0d and 15, respectively.
-        if content.endswith(b'\x0d\x15'):
-            return True
-        else:
-            return False
+            while content:
+                # In EBCDIC, \r\n are bytes 0d and 15, respectively.
+                if b'\x0d\x15' in content:
+                    return True
+                content = src_file.read(1024)
+
+        return False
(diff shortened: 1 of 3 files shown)


rexemin:
Jenkins pipeline
Screenshot 2023-04-04 at 16 16 02
Screenshot 2023-04-04 at 16 16 19

richp405 (approved):
looks okay to me.  I was surprised about IBM037 (cp 37)... used to hearing about 1047 instead, but I looked it up, and you are correct.

plugins/modules/zos_copy.py
```diff
@@ -1454,7 +1465,7 @@ def get_file_record_length(file):
     """
     max_line_length = 0
 
-    with open(file, "r") as src_file:
+    with open(file, "r", encoding="utf-8") as src_file:
```
ddimatos, line 1468: I have a question, so `get_file_record_length` is used to determine record length of a file should we enter into that use case. My question is you open in utf-8 always but if the source is 1047, python would try to read its utf-8. I thought opening a file with encoding is just overriding the default behavior which detects the systems locale but the file's actual encoding has to match what you pass as encoding.

In other words, can you always use utf-8 when the source could be another encoding or am I missing something in the code.

ddimatos: I might have figured it out, looks like you normalize on this block of code:
```
                    os.close(fd)
                    encode_utils.uss_convert_encoding(
                        new_src,
                        converted_src,
                        src_tag,
                        "UTF-8"
                    )
```

ddimatos (approved):
I am approving this @rexemin, nice work. I do have one question that I may have figured out but could use some clarity on. Thank you .

plugins/modules/zos_copy.py
rexemin: Yes, before getting to `get_file_record_length` the file gets normalized.

Issue #356, opened 2022-07-22
[Enhancement][ibm_zos_lineinfile]  Support disposition share (DISP=SHR)


ddimatos:
##### FEATURE

Module ibm_zos_lineinfile could benefit from supporting the disposition share (DISP=SHR) so that a member can be updated when a PDS/E has been opened buy a started task and will not be released. This would allow the module to force a change. 

[See ZOAU dmod -f](https://www.ibm.com/docs/en/zoau/1.2.0?topic=reference-dmod)

PR #731, opened 2023-04-11
Add force (disp=shr) option to zos_lineinfile and update zos_blockinfile with the same locking test case.

AndreMarcel99:
##### SUMMARY
Add force to options in zos_lineinfile module to enable DISP=SHR.
Addresses issue #356 
##### ISSUE TYPE
- Enhancement for functionality

##### COMPONENT NAME
Add the force option for the user and with that could modify a member while a task is held by another member.

##### ADDITIONAL INFORMATION
Add force option, set it to false as default, part of the options of ZOAU module use it.
Add all test cases for force with or without regex and the beginning or at the end even fail test case if force is false and a member is held.
Finally add a pair of test cases of force option in blockinfile with force option of succeed or fail.

Captura de pantalla 2023-04-11 a la(s) 10 44 18

### DIFF
--- a/plugins/modules/zos_lineinfile.py
+++ b/plugins/modules/zos_lineinfile.py
@@ -175,12 +175,16 @@ options:
     default: IBM-1047
   force:
     description:
-      - The case of a member of a PDSE held by a started task
-        other members cannot be acces/modified but is requiere for another
-        member from the same PDSE allow to work.
+      - Specifies that the data set can be shared with others during an update
+        which results in the data set you are updating to be simultaneously
+        updated by others.
+      - This is helpful when a data set is being used in a long running process
+        such as a started task and you are wanting to update or read.
+      - The C(force) option enables sharing of data sets through the disposition
+        I(DISP=SHR).
     required: false
     type: bool
-    default: False
+    default: false
 notes:
   - It is the playbook author or user's responsibility to avoid files
     that should not be encoded, such as binary files. A user is described
--- a/plugins/modules/zos_lineinfile.py
+++ b/plugins/modules/zos_lineinfile.py
@@ -476,7 +476,7 @@ def main():
     # state=absent, delete lines with matching regex pattern
     if parsed_args.get('state') == 'present':
         return_content = present(src, quotedString(line), quotedString(regexp), quotedString(ins_aft), quotedString(ins_bef), encoding, firstmatch,
-                                    backrefs, force)
+                                 backrefs, force)
     else:
         return_content = absent(src, quotedString(line), quotedString(regexp), encoding, force)
     stdout = return_content.stdout_response
--- a/plugins/modules/zos_lineinfile.py
+++ b/plugins/modules/zos_lineinfile.py
@@ -175,9 +175,9 @@ options:
     default: IBM-1047
   force:
     description:
-      - The case of a member of a PDSE held by a started task 
-      other members cannot be acces/modified but is requiere for another 
-      member from the same PDSE allow to work 
+      - The case of a member of a PDSE held by a started task
+      other members cannot be acces/modified but is requiere for another
+      member from the same PDSE allow to work.
     required: false
     type: bool
     default: False
@@ -331,7 +331,7 @@ def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs
     )
 
 
-def absent(src, line, regexp, encoding,force):
+def absent(src, line, regexp, encoding, force):
     """Delete lines with matching regex pattern
 
     Arguments:
@@ -405,7 +405,7 @@ def main():
         firstmatch=dict(arg_type="bool", required=False, default=False),
         backrefs=dict(arg_type="bool", dependencies=['regexp'], required=False, default=False),
         tmp_hlq=dict(type='qualifier_or_empty', required=False, default=None),
-        force=dict(arg_type='bool',required=False, default=False),
+        force=dict(arg_type='bool', required=False, default=False),
         mutually_exclusive=[["insertbefore", "insertafter"]],)
 
     try:
@@ -475,9 +475,10 @@ def main():
     # state=present, insert/replace a line with matching regex pattern
     # state=absent, delete lines with matching regex pattern
     if parsed_args.get('state') == 'present':
-        return_content = present(src, quotedString(line), quotedString(regexp), quotedString(ins_aft), quotedString(ins_bef), encoding, firstmatch, backrefs,force)
+        return_content = present(src, quotedString(line), quotedString(regexp), quotedString(ins_aft), quotedString(ins_bef), encoding, firstmatch,
+        backrefs, force)
     else:
-        return_content = absent(src, quotedString(line), quotedString(regexp), encoding,force)
+        return_content = absent(src, quotedString(line), quotedString(regexp), encoding, force)
     stdout = return_content.stdout_response
     stderr = return_content.stderr_response
     rc = return_content.rc
--- a/plugins/modules/zos_lineinfile.py
+++ b/plugins/modules/zos_lineinfile.py
@@ -173,6 +173,14 @@ options:
     required: false
     type: str
     default: IBM-1047
+  force:
+    description:
+      - The case of a member of a PDSE held by a started task 
+      other members cannot be acces/modified but is requiere for another 
+      member from the same PDSE allow to work 
+    required: false
+    type: bool
+    default: False
 notes:
   - It is the playbook author or user's responsibility to avoid files
     that should not be encoded, such as binary files. A user is described
@@ -218,6 +226,14 @@ EXAMPLES = r"""
     regexp: '^(.*)User(\d+)m(.*)$'
     line: '\1APPUser\3'
     backrefs: yes
+
+- name: Add a line to a member while a task is in execution
+  zos_lineinfile:
+    src: "{{ DATASET_NAME }}(MEM3)"
+    insertafter: EOF
+    line: 'Should be working know Mothefucker'
+    force: True
+
 """
 
 RETURN = r"""
@@ -271,7 +287,7 @@ except Exception:
 DS_TYPE = ['PS', 'PO']
 
 
-def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs):
+def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs, force):
     """Replace a line with the matching regex pattern
     Insert a line before/after the matching pattern
     Insert a line at BOF/EOF
@@ -292,6 +308,7 @@ def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs
         encoding: {str} -- Encoding of the src.
         first_match: {bool} -- Take the first matching regex pattern.
         backrefs: {bool} -- Back reference
+        force: {bool} -- force for modify a member part of a task in execution
 
     Returns:
         str -- Information in JSON format. keys:
@@ -310,10 +327,11 @@ def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs
         backref=backrefs,
         state=True,
         debug=True,
+        force=force,
     )
 
 
-def absent(src, line, regexp, encoding):
+def absent(src, line, regexp, encoding,force):
     """Delete lines with matching regex pattern
 
     Arguments:
@@ -322,6 +340,7 @@ def absent(src, line, regexp, encoding):
             regexp will be ignored.
         regexp: {str} -- The regular expression to look for in every line of the src.
         encoding: {str} -- Encoding of the src.
+        force: {bool} -- force for modify a member part of a task in execution
 
     Returns:
         str -- Information in JSON format. keys:
@@ -329,7 +348,7 @@ def absent(src, line, regexp, encoding):
             found: {int} -- Number of matching regex pattern
             changed: {bool} -- Indicates if the source was modified.
     """
-    return datasets.lineinfile(src, line, regex=regexp, encoding=encoding, state=False, debug=True)
+    return datasets.lineinfile(src, line, regex=regexp, encoding=encoding, state=False, debug=True, force=force)
 
 
 def quotedString(string):
@@ -364,7 +383,8 @@ def main():
         backup_name=dict(type='str', required=False, default=None),
         firstmatch=dict(type='bool', default=False),
         encoding=dict(type='str', default="IBM-1047"),
-        tmp_hlq=dict(type='str', required=False, default=None)
+        tmp_hlq=dict(type='str', required=False, default=None),
+        force=dict(type='bool', default=False)
     )
     module = AnsibleModule(
         argument_spec=module_args,
@@ -385,6 +405,7 @@ def main():
         firstmatch=dict(arg_type="bool", required=False, default=False),
         backrefs=dict(arg_type="bool", dependencies=['regexp'], required=False, default=False),
         tmp_hlq=dict(type='qualifier_or_empty', required=False, default=None),
+        force=dict(arg_type='bool',required=False, default=False),
         mutually_exclusive=[["insertbefore", "insertafter"]],)
 
     try:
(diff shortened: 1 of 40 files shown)


plugins/modules/zos_lineinfile.py
```diff
@@ -364,7 +387,8 @@ def main():
         backup_name=dict(type='str', required=False, default=None),
         firstmatch=dict(type='bool', default=False),
         encoding=dict(type='str', default="IBM-1047"),
-        tmp_hlq=dict(type='str', required=False, default=None)
+        tmp_hlq=dict(type='str', required=False, default=None),
+        force=dict(type='bool', default=False)
```
ketankelkar, line 391: looks like there's a missing `required=False`
```
force=dict(type='bool', required=False, default=False)
```

tests/functional/modules/test_zos_blockinfile_func.py
ketankelkar, line 1: add a `2023` to the copyright line
```
# Copyright (c) IBM Corporation 2020, 2022, 2023
```

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -1238,7 +1248,7 @@ def test_uss_block_insertafter_eof_with_backup_name(ansible_zos_module):
             for result in results.contacted.values():
                 assert result.get("stdout") == TEST_ENV["TEST_CONT"]
     finally:
-        ansible_zos_module.all.file(path=USS_BACKUP_FILE, state="absent")
+        ansible_zos_module.all.file(path=USS_BACKUP_FILE, state="absent")
```
ketankelkar, line 1251: looks like there's a trailing space character here, PEP8 might ding you for that

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -1498,6 +1508,17 @@ def test_ds_block_insertafter_eof_with_backup(ansible_zos_module, dstype, encodi
...
+#@pytest.mark.ds
+#@pytest.mark.parametrize("dstype", DS_TYPE)
+#def test_ds_block_insertafter_regex_force(ansible_zos_module, dstype):
+#    TEST_ENV["DS_TYPE"] = dstype
+#    DsGeneralForce(
+#        ansible_zos_module, TEST_ENV,
+#        TEST_INFO["test_ds_block_insertafter_regex_force"],
+#        TEST_INFO["expected"]["test_uss_block_insertafter_regex_defaultmarker"]
+#    )
+
```
ketankelkar, line 1520: was this test case supposed to remain commented out?

tests/functional/modules/test_zos_lineinfile_func.py
ketankelkar, line 1: update copyright years here too
```
# Copyright (c) IBM Corporation 2020, 2022, 2023
```

tests/helpers/zos_blockinfile_helper.py
```diff
@@ -155,3 +183,132 @@ def DsGeneralResultKeyMatchesRegex(test_name, ansible_zos_module, test_env, test
...
+        # submit jcl
+        hosts.all.shell(cmd="submit call_c_pgm.jcl", chdir="/tmp/disp_shr/")
+
+        # pause to ensure c code acquires lock
+        time.sleep(5)
+    
+        blockinfile_results = hosts.all.zos_blockinfile(**test_info)
+        for result in blockinfile_results.contacted.values():
+            pprint(result)
+            assert result.get("changed") == 1
```
ketankelkar, line 236: more semantics/style/good practice -- `result.changed` is a bool, so it should get compared to a bool. I know `1` resolves to `true` in python so this technically works, but I'd say this line is better off as:
```
assert result.get("changed") == True
```

tests/helpers/zos_blockinfile_helper.py
```diff
@@ -155,3 +183,132 @@ def DsGeneralResultKeyMatchesRegex(test_name, ansible_zos_module, test_env, test
...
+        # submit jcl
+        hosts.all.shell(cmd="submit call_c_pgm.jcl", chdir="/tmp/disp_shr/")
+
+        # pause to ensure c code acquires lock
+        time.sleep(5)
+    
+        blockinfile_results = hosts.all.zos_blockinfile(**test_info)
+        for result in blockinfile_results.contacted.values():
+            pprint(result)
+            assert result.get("changed") == 0
```
ketankelkar, line 303: see comment above about comparing to `bool`

tests/helpers/zos_lineinfile_helper.py
```diff
@@ -159,3 +186,126 @@ def DsGeneralResultKeyMatchesRegex(test_name, ansible_zos_module, test_env, test
...
+        # call line infile to see results      
+        results = hosts.all.zos_lineinfile(**test_info)
+        pprint(vars(results))
+        if test_env["ENCODING"] == 'IBM-1047':
+            cmdStr =r"""cat "//'{0}'" """.format(test_info["path"])
+            results = hosts.all.shell(cmd=cmdStr)
+            pprint(vars(results))
+            for result in results.contacted.values():
+                print()
+                assert result.get("stdout").replace('\n', '').replace(' ', '') == expected.replace('\n', '').replace(' ', '')
```
ketankelkar, line 243: what happens if `test_env["ENCODING"] != 'IBM-1047'`? i don't see an else statement to catch that -- is it supposed to be an automatic success? Just want to be cautious of false "passing" test scenarios

ketankelkar, line 243: also noticed no check of `results['changed']` unlike the other ones -- was this an intentional test design choice?

tests/helpers/zos_lineinfile_helper.py
```diff
@@ -159,3 +186,126 @@ def DsGeneralResultKeyMatchesRegex(test_name, ansible_zos_module, test_env, test
...
+        hosts.all.shell(cmd="xlc -o pdse-lock pdse-lock.c", chdir="/tmp/disp_shr/")
+        # submit jcl
+        hosts.all.shell(cmd="submit call_c_pgm.jcl", chdir="/tmp/disp_shr/")
+        # pause to ensure c code acquires lock
+        time.sleep(5)  
+        # call line infile to see results      
+        results = hosts.all.zos_lineinfile(**test_info)
+        pprint(vars(results))
+        for result in results.contacted.values():
+            assert result.get("changed") == 0
```
ketankelkar, line 300: see previous comment about comparing `bool` to `bool` instead of `bool` to `int`

ketankelkar (changes requested):
Really good work here! Great attention to detail and thoroughness in the test cases. Good job! Just a few minor comments, many stylistic

ketankelkar (approved):

tests/helpers/zos_blockinfile_helper.py
```diff
@@ -117,8 +145,8 @@ def DsGeneral(test_name, ansible_zos_module, test_env, test_info, expected):
         results = hosts.all.shell(cmd=cmdStr)
         for result in results.contacted.values():
             pprint(result)
-            assert result.get("stdout") == expected
-            # assert result.get("stdout").replace('\n', '').replace(' ', '') == expected.replace('\n', '').replace(' ', '')
+            #assert result.get("stdout") == expected
+            assert result.get("stdout").replace('\n', '').replace(' ', '') == expected.replace('\n', '').replace(' ', '')
```
ddimatos, line 149: If you replace `\n`'s for stdout and for expected and then replace ' ''s with '' for expected and stdout you have effectively balanced both sides , I have to wonder why this was needed over just using the original `result.get("stdout") == expected` ? Is there something with the `\n` and ' '  breaking our pytest cases they need to be stripped out?

tests/helpers/zos_lineinfile_helper.py
ddimatos: This is a good catch @ketankelkar - I am sorry if you mentioned a few weeks ago, not sure I was part of that discussion but if you run zos_lineinfile multiple times and it comes back with `results['changed'] == true and yet its a failure, either we have a bug or there is on in dmod.
We can investigate that in a separate git issue but that needs to exist, did you create an issue for it, put details and recreate steps into it? 

There is only one place changed is set to true and that is [here](https://github.com/ansible-collections/ibm_zos_core/blob/ad0506cd66563c36c3c4193f373842f7b16eb474/plugins/modules/zos_lineinfile.py#L507), so that means `dmod` failed and either we did not inspect correctly or they returned changed = true. 

This needs to be followed through.

ddimatos (changes requested):
I have only 2 comments, that are more questions than anything, once we discuss this I am happy to approve. This looks good @AndreMarcel99 thanks for adding the additional locking test case to lineinfile , glad @ketankelkar was able to help.

tests/helpers/zos_lineinfile_helper.py
ddimatos: This is a good point, its marked resolved but see no code that addresses this. Although not that familiar with the test cases for this module it looks like there is a check for encodings that are not equal to 1047, that is done in [set_ds_test_env](https://github.com/ansible-collections/ibm_zos_core/blob/enhancement/356/Support_disposition_share_DISP%3DSHR/tests/helpers/zos_lineinfile_helper.py#L140), it appears this function is used to normalize the target you are about to test to get it to an particular encoding. There is a list of 3 encodings ([ENCODING = ['IBM-1047', 'ISO8859-1', 'UTF-8']](https://github.com/ansible-collections/ibm_zos_core/blob/ad0506cd66563c36c3c4193f373842f7b16eb474/tests/functional/modules/test_zos_lineinfile_func.py#L68)) that are often used but seem to have been skipped in the more recent tests cases added. 

I would say all these test cases need to be rewritten as they are difficult to follow but that is not in plan, but for now this is a valid question that was resolved and I don't see any response or code change to support the resolved comment. 

I can understand encodings being new to folks, but we need to ensure we are doing our testing adequately

tests/helpers/zos_blockinfile_helper.py
AndreMarcel99: I'm not too sure, but for the datasets of type 'PDSE' sometimes mark that are not equal after a few attempts of tests

tests/helpers/zos_lineinfile_helper.py
AndreMarcel99: I guess it could go along with another little task for nexts iterations

AndreMarcel99: I didn't understand well the problem sorry just avoid the change of encoding but I work on it and I realize somethings:

Well I use the functions from other helpers but like here it doesn't check output to expected if it's another encoding too and is the function use for DS types [here](https://github.com/ansible-collections/ibm_zos_core/blob/ad0506cd66563c36c3c4193f373842f7b16eb474/tests/helpers/zos_lineinfile_helper.py#L138 ) also the other function call to make the assertion in case for another encoding just base with these `assert int(result.get("stdout")) != 0`. Could you explain me about the [assertion](https://github.com/ansible-collections/ibm_zos_core/blob/ad0506cd66563c36c3c4193f373842f7b16eb474/tests/helpers/zos_lineinfile_helper.py#L124)?

I'm working on the false passing cases for all other encodings.

ddimatos: Unfortunately this test code is written in such a way its quite difficult to follow, from what i see it looks like ` set_ds_test_env` will check if the from encoding is ibm-1047, when it's not it ,it uses the encoding module to change the local data to match the too encoding. 

Eg, if the list of encodings being traversed lands on ISO-88591, it changes the content meaning it writes it out as ISO-88591 so that way it when it copies it to the target data set, the blockinfile/linefile tests can operate on the data in the encoding that matches. 

A data set can not contain ISO-88591 and then the blockinfile module think its IBM-1047 else it will break, no regex will work. 

So i think this function sets up the data on the target for the ansible module so that it does not break. To answer your questions now , `assert int(result.get("stdout")) != 0` , after all this copying and encoding conversions are done, the shell module is going to cat the data set to make sure there is content in the data set, and thus stdout should not be be false, (using 0 is poor idea here). 

This `cat` that checks for data really is not enough , it just tells you a copy was successful, it does not reverse the encoding to validate the enconding was successful but that is ok for now, as I said these tests cases should all be rewritten one at a time an managed within the test function not spread out over a large dictionary and a bunch of helpers. 

I think we could live with the way its written now, but we are limiting our encoding testing, whats funny is the modules manage all the encoding yet we have all this added complexity , there are easier ways to do this. Spend a bit of time on it, if its starts to make sense lets work on it, if we need to talk that is fine, also Fernando knows this code fairly well.

AndreMarcel99: Is kind of solve with the last push

AndreMarcel99: I got a problem with the design because with zos_encode could work but time not to well, try with 
`iconv -f {0} -t {1} "//'{2}'" > "//'{2}'" """.format(test_info["encoding"], 'IBM-1047', test_info["path"])` to past all text with IBM-1047 except for the line added to return to IBM-1047 but had a problem because it change the message and throw garbage 
Captura de pantalla 2023-04-19 a la(s) 15 43 11

AndreMarcel99: I already add to cover all cases, but not verify as I would like the encodings

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -568,7 +587,42 @@
 export PYTHONPATH
 export PKG_CONFIG_PATH
 export PYTHON_HOME
-export _BPXK_AUTOCVT"""),
+export _BPXK_AUTOCVT""",
```
ddimatos, line 590: This is only a comment, no action needed. We really did not need to extend this sample data to test force but it made it easier for the developer @AndreMarcel99 to see his changes working, in case you are reviewing this I wanted you to understand why the data was extended, although we would ask to remove it, i see no point, its just adding some extra reading time. :)

tests/helpers/zos_blockinfile_helper.py
AndreMarcel99: It was a problem about the file stay there the solution it was with the deletion of file for every [test](https://github.com/ansible-collections/ibm_zos_core/blob/edd2a44eb3c80ccb22316c6c81a3b8cc983574e7/tests/helpers/zos_lineinfile_helper.py#L276)

tests/helpers/zos_lineinfile_helper.py
ddimatos: Ok, after speaking with you , you were able to resolve this in the test code, so it was not no an issue in the module ore dmod code is how I understood it. I will resolve this.

ddimatos: Understood, I am not sure without some hands on investigation but I think the issue is the source file encoding is not what is iconv is being told is correct. 

Eg, if for a file on your controller is encoding in UTF8 (for example) and you tell iconv -f ISO-8859-1 -t IBM-1047, its going to end up like garbage because it mapped a UTF8 file as ISO-8859-1 to IBM-1047, this is a guess but i wonder if what you had seen was just a source issue i describe.

tests/helpers/zos_blockinfile_helper.py
ddimatos: Thanks, so the issue was resolved when you cleaned up the test case, likely what i commented [here](https://github.com/ansible-collections/ibm_zos_core/pull/731#discussion_r1175753191)

Ill paste my comment:
```


Understood, I am not sure without some hands on investigation but I think the issue is the source file encoding is not what is iconv is being told is correct.

Eg, if for a file on your controller is encoding in UTF8 (for example) and you tell iconv -f ISO-8859-1 -t IBM-1047, its going to end up like garbage because it mapped a UTF8 file as ISO-8859-1 to IBM-1047, this is a guess but i wonder if what you had seen was just a source issue i describe.

```

ddimatos (approved):
Looks good, thanks for following up with the updates.

AndreMarcel99:
Captura de pantalla 2023-04-24 a la(s) 17 20 53

PR #641, opened 2023-02-14
Bugfix/601 copy lib member 1.5.0

fernandofloresg:
##### SUMMARY

Fixed error when a load lib member was copied to another load lib. Apparently error is found in stdout instead of err.

Fixes #601 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_copy

##### ADDITIONAL INFORMATION

```paste below

```

PR #736, opened 2023-04-14
Add copy lib member test case for 1.5.0

fernandofloresg:
##### SUMMARY

The fix for this issue was merged in #641, but the test case took more time to finalize.

##### ISSUE TYPE

- Enhancement Pull Request

##### COMPONENT NAME

module_utils/zostest_zos_copy_func.py 

##### ADDITIONAL INFORMATION

```paste below

```


fernandofloresg:
Also, based on comments by Demetri changed the changelog for the fix to the correct PR number in fragment.

fernandofloresg:
Screenshot 2023-04-13 at 8 36 48 PM

Screenshot 2023-04-13 at 8 37 46 PM

changelogs/fragments/641-copy-loadlib-member.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
```
ddimatos, line 1: My only comment is if this should be a trivial fragment since we have fragment 601 for the code change. Trivial fragments don't get included in the change-log, they are meant to help us, given this is a test cas only, seems like a good idea. 

Here is the one I did for your other PR if you agree to change:
```
trivial:
- zos_copy - Adds a test cases to ensure copying from a PDS/E member containing
  a loadlib to another PDS/E member loadlib member for issue 601.
  (https://github.com/ansible-collections/ibm_zos_core/pull/734)
```

ddimatos (changes requested):

changelogs/fragments/641-copy-loadlib-member.yml
fernandofloresg: I get your point and didn't intend to include a changelog fragment for a new test case, but rather, this is correcting the fact that 601 is the issue number, and by your comments "Format of the fragment name (try to keep it somehow short yet helpful), don’t use the issue number use the PR number thus you will need to have a PR draft before you can get a PR number. [...]" in this PR I removed 601 fragment and added the correct PR # 641 for the fix in the 1.5.0 branch.

changelogs/fragments/641-copy-loadlib-member.yml
```diff
@@ -0,0 +1,3 @@
...
+- zos_copy - Copy failed from a loadlib member to another loadlib member. Fix now looks for error in stdout in the if statement to use -X option.
+(https://github.com/ansible-collections/ibm_zos_core/pull/641)
```
richp405, line 3: Shouldn't this be pull/736?
Also, github is flagging the fact that there is no newline at the end of the file.

richp405 (changes requested):
code looks good to me.  I think the fragment needs 2 changes: use this PR number (736) and add newline at end.

changelogs/fragments/641-copy-loadlib-member.yml
fernandofloresg: No, as I mentioned to Demetri "in this PR I removed 601 fragment and added the correct PR # 641 for the fix in the 1.5.0 branch.", there is no change log fragment for this PR (736) bc of the triviality of the test case.

About the no newline, this is the first time I get feedback about it. I don't know if is an issue when building the change log but I'll change it.

changelogs/fragments/641-copy-loadlib-member.yml
ddimatos: I am ok with this, i see why we would not want a changelog fragment but I was thinking a bit further down the line, we had discussed a pipeline check to look for changelog fragments.

If we add the check in our pipepline, we won't have a way to easily determine if the PR should have a fragment or not, hence I was just encouraging trivial usage as it won't be part of our changelog but would pass our pipeline in the future.

ddimatos: Since there is a fragement now correcting the old one , that is  good, thus I am ok with this now.

ddimatos (approved):
Thank you, I am good with this, thanks for the forward port.

richp405 (approved):

changelogs/fragments/641-copy-loadlib-member.yml
ddimatos: Marking this as resolved as it appears to be given the reviewer has passed the PR

Issue #660, opened 2023-02-28
[Bug] zos_operator reported failure caused by unrelated error response messages


cdlwhui:
### Bug description

We have the following task in a playbook. The last time we ran this task it reported failure even though  the command was issued successfully.  

```
    - name: STRACE
      zos_operator:
          cmd: 'RO *ALL,TRACE ST,BUFSIZ=20M'
      register: straceout
```

From the output below. it looked like recognizing the message “ TP3       2023058  05:26:31.86  STC07450   +ITP208I OPEN ACB FAILED FOR DISPLAY MONITOR FACILITY - RETURN CODE=08 ERROR FIELD=5A - APPLID NOT FOUND” as the error response message  of command “RO *ALL,TRACE ST,BUFSIZ=20M”. But actually the error message was totally unrelated with the command “RO *ALL,TRACE ST,BUFSIZ=20M”.  The 3 messages from system TP3 are other messages unrelated with the command. Sometime there are messages occurred on the user console together with the response of user’s command which should because they are generated at the same time. It looked like zos_operator will catch all the messages as the response of command  and report error as long as there are error messages.

This issue does not happen every time. Just when there are other messages occurred on user console at the same time and there are error messages.  And I think it maybe difficult to recreate it.  But if it is designed that zos_operator will catch all the messages occurred  as the responces of command and report error according to these messages, this issue will be always there. 

```
{
    "msg": "Error detected: TP3       2023058  05:26:31.86  STC07450   +ITP208I OPEN ACB FAILED FOR DISPLAY MONITOR FACILITY - RETURN CODE=08 ERROR FIELD=5A - APPLID NOT FOUND",
    "changed": false,
    "rc": 0,
    "content": [
        "ST1       2023058  05:26:31.54             ISF031I CONSOLE ANSIPL ACTIVATED",
        "ST1       2023058  05:26:31.54            -RO *ALL,TRACE ST,BUFSIZ=20M",
        "TP3       2023058  05:26:31.83  STC07450   +ITP016I TPNS VERSION 3, RELEASE 5.0, SERVICE LEVEL 0601, MARCH 3, 2009, 14:32",
        "TP3       2023058  05:26:31.86  STC07450   +ITP003I TPNS INITIALIZATION COMPLETE",
[...]

PR #762, opened 2023-04-27
Bugfix/660/zos operator reported failure caused by unrelated error response messages

AndreMarcel99:
##### SUMMARY

Change zos_operator module by delete the search of keywords as triggers for failures in a a send operator.

Fixed bugfix with option one #660 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Separate the response and delete search of keywords.

##### ADDITIONAL INFORMATION
Separate the stdout and stderr response to send in a fail only the stderr. Delete the search of keywords as error in the stdout.

```
ok: [zvm] => {
    "msg": {
        "changed": true,
        "cmd": "DUMP COMM=('ERROR DUMP')",
        "content": [
            "EC01147A   2023118  15:05:26.00             ISF031I CONSOLE OMVS0000 ACTIVATED",
            "EC01147A   2023118  15:05:26.00            -DUMP COMM=('ERROR DUMP') ",
            "EC01147A   2023118  15:05:26.00             09 IEE094D SPECIFY OPERAND(S) FOR DUMP COMMAND"
        ],
        "elapsed": 5.81,
        "failed": false,
        "rc": 0,
        "wait_time_s": 1
    }
}
```
Captura de pantalla 2023-05-01 a la(s) 20 44 54

### DIFF
--- a/plugins/modules/zos_operator.py
+++ b/plugins/modules/zos_operator.py
@@ -244,14 +244,13 @@ def run_module():
             else:
                 module.fail_json(msg="Expected response to be more than 2 lines.", **result)
         else:
-            module.fail_json(
-                              msg=("A non-zero return code was received : {0}. Review the response for more details.").format(result["rc"]),
-                              cmd=result["cmd"],
-                              elapsed_time=result["elapsed"],
-                              wait_time_s=result["wait_time_s"],
-                              stderr=str(error) if error != None else result["content"],
-                              stderr_lines=str(error).splitlines() if error != None else result["content"],
-                              changed=result["changed"],)
+            module.fail_json(msg=("A non-zero return code was received : {0}. Review the response for more details.").format(result["rc"]),
+                            cmd=result["cmd"],
+                            elapsed_time=result["elapsed"],
+                            wait_time_s=result["wait_time_s"],
+                            stderr=str(error) if error is not None else result["content"],
+                            stderr_lines=str(error).splitlines() if error is not None else result["content"],
+                            changed=result["changed"],)
     except Error as e:
         module.fail_json(msg=repr(e), **result)
     except Exception as e:
--- a/plugins/modules/zos_operator.py
+++ b/plugins/modules/zos_operator.py
@@ -219,14 +219,14 @@ def run_module():
         result["content"] = []
         stdout = rc_message.get("stdout")
         if stdout is not None:
-          for out in stdout.split("\n"):
-            result["content"].append(out)
+            for out in stdout.split("\n"):
+                result["content"].append(out)
         stderr = rc_message.get("stderr")
         error = []
         if stderr is not None:
-          for err in stderr.split("\n"):
-            error.append(err)
-            result["content"].append(err)
+            for err in stderr.split("\n"):
+                error.append(err)
+                result["content"].append(err)
         # call is returned from run_operator_command, specifying what was run.
         # result["cmd"] = new_params.get("cmd")
         result["cmd"] = rc_message.get("call")
@@ -243,7 +243,7 @@ def run_module():
                 module.fail_json(msg="Expected response to be more than 2 lines.", **result)
         else:
             module.fail_json(
-                msg="Non-zero response received: " + str(result["rc"] + "\n" +". Stderr: " + str(error)), **result
+                msg="Non-zero response received: " + str(result["rc"] + "\n" + ". Stderr: " + str(error)), **result
             )
     except Error as e:
         module.fail_json(msg=repr(e), **result)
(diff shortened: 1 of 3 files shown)


tests/functional/modules/test_zos_operator_func.py
```diff
@@ -49,8 +49,18 @@ def test_zos_operator_invalid_command(ansible_zos_module):
     hosts = ansible_zos_module
     results = hosts.all.zos_operator(cmd="invalid,command", verbose=False)
     for result in results.contacted.values():
-        assert result.get("changed") is False
-        assert result.get("exception") is not None
+        assert result.get("changed") is True
```
richp405, line 52: Just need to double-check: you want this test to show changed is True when processing an invalid command?

AndreMarcel99: Yes because in the rc is 0 even with that command, even if bring in the message error words. Because that's the solution suggested in the conversation to be transparent with the result. Maybe I could change part of the code if bring that error keywords to keep changed as False, but be transparent with the response.

richp405: okay, that makes sense.

plugins/modules/zos_operator.py
```diff
@@ -247,26 +239,13 @@ def run_module():
...
-                    elif "unidentifiable" in linetocheck.lower():
-                        result["exception"] = "Unidentifiable detected: " + linetocheck
-                        result["changed"] = False
-                        module.fail_json(msg=result["exception"], **result)
             else:
                 module.fail_json(msg="Expected response to be more than 2 lines.", **result)
         else:
             module.fail_json(
-                msg="Non-zero response received: " + str(result["rc"]), **result
+                msg="Non-zero response received: " + str(result["rc"] + "\n" + ". Stderr: " + str(error)), **result
```
ddimatos, line 248: My second comment is why are you adding "Stderr" like so, as if its a key:value pair we want to pass back. You have the dictionary **result from what I can see has ALMOST everything you need. 

What I would do is either populate the `**result` further such that

Or create a dynamic one with what the docs state will returned. For example:
```
module.fail_json(msg=(
                        "A non-zero return code was received : {0}. Review the response for more details.").format(result["rc"]),
                    stderr=str(error),
                    stderr_lines=str(error).splitlines(),
                   cmd=result["cmd"],
                   stdout=str(stdout),
                   stdout_lines=str(stdout).splitlines(),
                   wait_time_s...
                   elapsed_time....
                  and so on....)
```
My personal preference is you start with result dictionary and keep appending as you go along, start by putting all the information you know is needed per our [response contract](https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/modules/zos_operator.html#return-values). 

And the nice thing is if you build it as you move along in the code you can be append to the dictionary when values are not None. For example, in your `module.fail_json` you did not check if this `Stderr: " + str(error))` was not none. 

If you build the dictionary as you go along you can use the checks you have that are checking if stdout or stderr are not none and then at the very end in the exception you can just return `module.fail_json(msg="Your mess here", **result)`

tests/functional/modules/test_zos_operator_func.py
ddimatos: Makes sense, this is because we are transparent now and simply return back and no longer scan for keywords.

ddimatos (changes requested):
Some minor comments, I think we still have a bit more to do here, apologize for the churn @AndreMarcel99 , overall the work is good, if you have questions or my comments are not clear ,please reach to me.

plugins/modules/zos_operator.py
AndreMarcel99: Here is my response two both comments. I change to the result dictionary to the [response contract](https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/modules/zos_operator.html#return-values). Just change the content for stderr and stderr lines, also verify if are None just put all the content.

AndreMarcel99: I run the pipeline for [Ansible Certification] (Captura de pantalla 2023-05-05 a la(s) 10 33 08) and is ready.

richp405 (approved):
That looks much cleaner now.

plugins/modules/zos_operator.py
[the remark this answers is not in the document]
ddimatos: Looks good thanks so much @AndreMarcel99

ddimatos: @AndreMarcel99 looks great thanks so much for the added effort !!!

ddimatos (approved):
Thanks so much for addressing the comments, looks good @AndreMarcel99

Issue #770, opened 2023-05-17
[Documentation] Versions of Ansible/Ansible Core and z/OS are slightly outdated


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Documentation issue description

Since ZOAU 1.2.2 requires z/OS 2.4 , the release notes must be updated.
In the readme it speaks to Ansible versions supported 2.9 <=2.15, this is a bit misleading because it should speak to Ansible (2.9), Ansible Base (2.10) , Ansible Core 2.10 or later. 

This will also address github issue template updates. 

### IBM z/OS Ansible core Version

v1.4.0-beta.2

ddimatos:
<img width="662" alt="image" src="https://github.com/ansible-collections/ibm_zos_core/assets/25803172/941730cf-0bee-4805-92ff-c558745f19c2">

<img width="997" alt="image" src="https://github.com/ansible-collections/ibm_zos_core/assets/25803172/c2ef71cb-76d3-4aa7-bb21-784831440303">

PR #771, opened 2023-05-17
Update docs with ansible/ansible-core version, AAP and fix the dated git issue templates

ddimatos:
##### SUMMARY
Addresses issue #770 

##### ISSUE TYPE
- Docs Pull Request


ddimatos:
image

ddimatos:
image

ddimatos:
I see no reason to run regression, no source was changed.
On another note, this does speak to ibm_zos_core version 1.5.0 but we have no maintenance branch started and don't know if its needed to start one based on a doc change. 

We can't update the README in Galaxy without a 1.5.1 release thus see no point to start a maintenance branch until we have module bug. The general docs https://ibm.github.io/z_ansible_collections_doc/index.html will be updated with the release notes one this merges.

.github/ISSUE_TEMPLATE/doc_issue.yml
```diff
@@ -40,20 +40,23 @@ body:
...
-        - v1.1.0
-        - v1.2.1
-        - v1.3.0
-        - v1.3.1
-        - v1.3.3
-        - v1.3.5
+        - v1.9.0
+        - v1.8.0
+        - v1.7.0
+        - v1.6.0
```
AndreMarcel99, line 49: I prefer the new order from newest to oldest by descending. Just two personal doubts
Before were include the betas version, wouldn't be nice to keep with that including the 1.6.0-beta at the moment?
Also Is ok to add all that next versions?
The second doubt apply to the the 3 files changes in ISSUE_TEMPLATE.

richp405 (approved):

AndreMarcel99 (approved):
Just two personal doubts
Before were include the betas version, wouldn't be nice to keep with that including the 1.6.0-beta at the moment?
Also Is ok to add all that next versions?
The second doubt apply to the the 3 files changes in ISSUE_TEMPLATE.

.github/ISSUE_TEMPLATE/doc_issue.yml
ddimatos: @AndreMarcel99 - I think what you prefer the descending (newest to oldest) is what how it will show in the git issue, this PR is actually correcting the behavior that you are commenting on. GH issues show the selection box as it is ordered, have a look at [here](https://raw.githubusercontent.com/ansible-collections/ibm_zos_core/dev/.github/ISSUE_TEMPLATE/doc_issue.yml) and you will see the order starts with 1.0 to later versions, so we are both in agreement, and this is aligning to your comment.

ddimatos: This template is for documentation, our docs really don't align to a version (we have work items to have version support on our docs in the backlog in JIRA (not yet ported to GH), I figured it might be good to have a version to associate to a doc , maybe its better we do use a `x` for the last bit , eg 1.6.x. Our docs are slightly different where with issues related to modules and source we actually do need to include beta's. I could push this out with `x`'s instead.

ddimatos: I think its ok to add all the next versions, its more or less future proofing, if we find users are selecting them when there is no version we can remove them, what I find worse that happens is we are not keeping up with updates to the templates hence I added a few more future release versions. Not sure what is worse, not having the templates updated with the most current release or having future releases, I tend to think its worse to not have the templates updated so when someone tries to open an issue, the version is not there.

ddimatos: I suppose another option is as part of our release strategy we update the templates, that could solve this problem.

ddimatos:
> Just two personal doubts Before were include the betas version, wouldn't be nice to keep with that including the 1.6.0-beta at the moment? Also Is ok to add all that next versions? The second doubt apply to the the 3 files changes in ISSUE_TEMPLATE.

Good points @AndreMarcel99 - i will add the beta's back to the bug templates, i think copy and paste was to easy :) and overlooked that. I meant to keep the 1.6.0-beta.1 and 1.7.0-beta.1 and upwards , but older we don't need beta's after they GA. I'll push some changes.

ddimatos:
@AndreMarcel99 - I updated the approach and added the beta's to all, the order should be descending even thought it appears not, I also only included the latest of each release, this is based on the idea if a bug is opened say on 1.4.0 and we have 1.4.1, we should ask the user to be on the latest of that version should we need to perform a fix.
eg, 
```
        - v1.9.0
        - v1.9.0-beta.1
        - v1.8.0
        - v1.8.0-beta.1
        - v1.7.0
        - v1.7.0-beta.1
        - v1.6.0
        - v1.6.0-beta.1
        - v1.5.0
        - v1.4.1
        - v1.3.6
        - v1.2.1
        - v1.1.0
        - v1.0.0
```

PR #772, opened 2023-05-17
Encode files recursively and test case for keep behavior.

### LINKED ISSUE #419: [Enhancement] [zos_copy] Return the dynamically created destintation attributres
##### FEATURE
The zos_copy module covers many edge cases for users who set a destination that does not exists. For example if the src is USS and only a dest data set name is provided but does not exist, zos_copy goes and figures out the record length , block size, primary and secondary space, block format etc, as you can see many attributes are decided in these edge cases. It would be nice if the response included the attributes (or at least worth discussing) in the event that what the zos_copy module decides are reasonable attributes and they are not or worse it fails and support teams are then challenged to recreate and guess what attributes were decided when it failed for the user.

### LINKED ISSUE #616: [Bug] zos_copy does not encode files recursively
### Bug description

Steps to reproduce are simple, just try running `zos_copy` with a folder that contains files on multiple levels, e.g.
- test1
    - file1.sh
    - test-inner
        -  file2.sh
Then run the following task:
```
- name: Copy a folder
  zos_copy:
    src: "{{src_path}}"
    dest: "{{dest_path}}"
    force: yes
    is_binary: no
    validate: yes
    encoding:
      from: ISO8859-1
      to: IBM-1047
```

All files will be copied correctly, and all will be tagged, however only `file1.sh` will actually be converted to IBM-1047.
It's easy to see why if you look at the code:
```python
path, dirs, files = next(os.walk(dir_path))
        enc_utils = encode.EncodeUtils()
        for file in files:
            full_file_path = path + "/" + file
```
You only call `next` once, should be:
```python
enc_utils = encode.EncodeUtils()   # make it outside the loop
for path, dirs, files in os.walk(dir_path):
    for file in files:
        # do your thing
```

### Playbook verbosity output

_No response_

### Contents of `ansible.cfg`

_No response_

### Contents of the inventory

_No response_

_No response_

### Ansible version

```shell
core 2.11.10
  python version = 3.7.9 (default, Jan 19 2021, 19:42:53) [Clang 12.0.0 (clang-1200.0.32.28)]
  jinja version = 3.0.3
  libyaml = True
```

v1.4.0

### IBM ZOAU version

v1.2.0

### z/OS version

_No response_

### Ansible module

_No response_

AndreMarcel99:
##### SUMMARY
Add a cycle to look inside the whole folder of source and iterate to content if there is any other folder.
Fixes #616  

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Encode files recursively and test case for the behavior.

##### ADDITIONAL INFORMATION
All the test that generate dependencies run well locally just need to verify the full.

Captura de pantalla 2023-05-22 a la(s) 21 47 44


tests/functional/modules/test_zos_data_set_func.py
```diff
@@ -809,42 +809,54 @@ def test_data_set_temp_data_set_name_batch(ansible_zos_module):
     ["HFS", "ZFS"],
 )
 def test_filesystem_create_and_mount(ansible_zos_module, filesystem):
+    hosts = ansible_zos_module
```
rexemin, line 812: Is this change in `test_zos_data_set_func.py` related to the issues the pipeline was having last week?

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -468,11 +468,57 @@ def test_copy_dir_to_existing_uss_dir_not_forced(ansible_zos_module):
         hosts.all.file(path=dest_dir, state="absent")
 
 
+@pytest.mark.uss
+def test_copy_inner_folders_and_validate_recursive_encoding(ansible_zos_module):
```
rexemin, line 472: Good test, I think it would also be really helpful if you added something similar but copying local files and checking that their contents are correct.

I'd also say that `subdirs` instead of `inner_folders` may be a more standard way of calling what you're testing.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -252,26 +252,26 @@ def link_loadlib_from_cobol(hosts, ds_name, cobol_pds):
 ])
 def test_copy_file_to_non_existing_uss_file(ansible_zos_module, src):
     hosts = ansible_zos_module
-    dest_path = "/tmp/zos_copy_test_profile"
+    src_path = "/tmp/zos_copy_test_profile"
```
rexemin, line 255: I'm curious why you changed `dest_path` to `src_path` across many tests, did you change that on all of them?

changelogs/fragments/772-Encode-files-recursively-and-test-case-for-keep-behavior.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - zos_copy - Zos_copy did not encode inner content of a inner folder on
```
rexemin, line 2: Probably nitpicking here, but the fragment should start with `zos_copy` instead of `Zos_copy`.

I'd also change `of a inner folder on the source is copy to the destiny` to `inside subdirectories once the source was copied to the destination`

changelogs/fragments/772-Encode-files-recursively-and-test-case-for-keep-behavior.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - zos_copy - Zos_copy did not encode inner content of a inner folder on 
+    the source once is copy to the destiny.
+    Fix now encodes all content in a source folder include inner files of
```
rexemin, line 4: Here I'd say that you could change it to `Fix now encodes all content in a source directory, including subdirectories.`

rexemin (changes requested):

tests/functional/modules/test_zos_data_set_func.py
AndreMarcel99: Yes

tests/functional/modules/test_zos_copy_func.py
AndreMarcel99: I'll change it and add the test case

tests/functional/modules/test_zos_copy_func.py
AndreMarcel99: An unnecessary change, I already change it to normal again

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -468,6 +468,56 @@ def test_copy_dir_to_existing_uss_dir_not_forced(ansible_zos_module):
...
+
+    try:
+        hosts.all.file(path=inner_src_path, state="directory")
+        hosts.all.file(path=inner_file, state = "touch")
+        hosts.all.file(path=outer_file, state = "touch")
+        hosts.all.shell(cmd="echo '{0}' > '{1}'".format(text_outer_file, outer_file))
+        hosts.all.file(path=dest_path, state="directory")
+        hosts.all.shell(cmd="echo '{0}' > '{1}'".format(text_inner_file, inner_file))
+
+        copy_res = hosts.all.zos_copy(content=src_path, dest=dest_path, encoding={"from": "ISO8859-1", "to": "IBM-1047"}, remote_src=src["is_remote"])
```
fernandofloresg, line 494: How is this working when remote_src is False? Here is the deal, when src_path is created using `hosts.all.file(path=inner_src_path, state="directory")` it is acting upon the host, so local shouldn't have a folder named "/tmp/level_1/". 

Also, I don't think `src` needs to be a dict at this step

```
@pytest.mark.parametrize("src", [
    dict(is_remote=True),
    dict(is_remote=False),
])
```

Why not use? 

```
@pytest.mark.parametrize("src", [ True, False])
```

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -468,6 +468,56 @@ def test_copy_dir_to_existing_uss_dir_not_forced(ansible_zos_module):
...
+        outer_file_text_aft_encoding = hosts.all.shell(cmd="cat {0}".format(outer_file))
+        for text in outer_file_text_aft_encoding.contacted.values():
+            text_outer = text.get("stdout")
+        for text in inner_file_text_aft_encoding.contacted.values():
+            text_inner = text.get("stdout")
+
+        assert text_inner == text_inner_file
+        assert text_outer == text_outer_file
+    finally:
+        hosts.all.file(path=inner_file, state="absent")
```
fernandofloresg, line 514: Should be enough to remove the hierarchy only with these two right? 

hosts.all.file(path=src_path, state="absent") // "/tmp/level_1/"
hosts.all.file(path=dest_path, state="absent") // "/tmp/test/"

tests/functional/modules/test_zos_copy_func.py
fernandofloresg: Please review the test case when remote_src is False, I don't get it.

fernandofloresg (changes requested):

tests/functional/modules/test_zos_copy_func.py
fernandofloresg: What was the issue with the original test? Why was it throwing a false positive? Is that something we need to look ? Did you recreate the false positive ?

AndreMarcel99: The issue was that I put in the cat statement the path of src not of dest and then with the change I realized some more errors of paths

fernandofloresg: Then that is that we need to look into right? If we are throwing a false positive and you've detected the issue we should have a GH issue for it at least. Could you provide more details on that?

AndreMarcel99: There is no problem in other test cases. Is this [instruction](https://github.com/ansible-collections/ibm_zos_core/blob/8e598dfe042f42c4ec3a0099d1c977fd7d85b863/tests/functional/modules/test_zos_copy_func.py#L264) where usually just copy a file there is no problem, but in my case for folders need to change a little bit, I'm don't interested in the dest_path. I need the dest_path plus the subidr.

AndreMarcel99:
Captura de pantalla 2023-05-30 a la(s) 17 20 23

fernandofloresg (approved):
LGTM! Thanks for addressing our comments

rexemin (approved):
Thanks for working on this, looks good!

PR #773, opened 2023-05-17
Return the dynamically created destination attributes

### LINKED ISSUE #419: [Enhancement] [zos_copy] Return the dynamically created destintation attributres

AndreMarcel99:
##### SUMMARY
Return the values of block_size, record_format, record_length, space_primary, space_secondary, space_type and type information for the user.
Fixed #419 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Return a message in the response of dest created and the values created.

##### ADDITIONAL INFORMATION
The function to create and assign the values of size, and record_format, record_length did not return the values, now return all to give the information to the user, and also with the message of dest created as True.

```
ok: [zvm] => {
    "msg": {
        "changed": true,
        "dest": "TEST.DATA(MEM)",
        "dest_created": true,
        "destination_attributes": {
            "block_size": 32760,
            "record_format": "FB",
            "record_length": 45,
            "space_primary": 2,
            "space_secondary": 1,
            "space_type": "K",
            "type": "PDSE"
        },
        "failed": false,
        "is_binary": false,
        "src": "/SYSTEM/tmp/file"
    }
}
```
Captura de pantalla 2023-05-23 a la(s) 10 09 40


changelogs/fragments/773-Return-the-dynamically-created-destintation-attributres.yaml
fernandofloresg, line 1: This description is confusing, I would keep the last part. I liked better how you stated the problem in the PR description :

"Return the values of block_size, record_format, record_length, space_primary, space_secondary, space_type and type information for the user."


This could be better with something like "Adds the MVS data set attributes of a dynamically created destination in the return output"

changelogs/fragments/773-Return-the-dynamically-created-destintation-attributres.yaml
```diff
@@ -0,0 +1,6 @@
+minor_changes:
+- zos_copy - zos_copy covers many edge cases for users who set a destination 
+  that does not exists but the information about creation was not included
+  in the output. This enhancement adds the to the output information to about
+  the destination created.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/721)
```
fernandofloresg, line 6: The pull request number is wrong

plugins/modules/zos_copy.py
```diff
@@ -2449,7 +2452,7 @@ def run_module(module, arg_def):
 
     try:
         if not is_uss:
-            res_args["changed"] = allocate_destination_data_set(
+            res_args["changed"], res_args["dynamic_values_dest"] = allocate_destination_data_set(
```
fernandofloresg, line 2455: I think we could use "dest_data_set_attrs" or similar instead of "dynamic_values_dest". I get that those are the values created dynamically but at the end of the day they're just the attributes of the data set. Same for the action plugin variable name.

plugins/modules/zos_copy.py
```diff
@@ -2006,6 +2006,7 @@ def allocate_destination_data_set(
 
     Returns:
         bool -- True if the data set was created, False otherwise.
+        dest_params -- values of the dataset allocated if exist is and empty dictionary
```
fernandofloresg, line 2009: This return statement is confusing. I would prefer something like a dict with the attributes of the allocated destination data set if it was created by the module

fernandofloresg (changes requested):
Why is there no zos_copy test case update? I don't think a new one is created but at least add some assertions in tests like :
test_copy_file_to_non_existing_pdse
test_copy_ps_to_non_existing_ps
test_copy_dir_crlf_endings_to_non_existing_pdse

to name a few.

changelogs/fragments/773-Return-the-dynamically-created-destintation-attributres.yaml
rexemin: I agree with Fernando's comments about how the fragment is currently written.

plugins/modules/zos_copy.py
```diff
@@ -2014,8 +2015,10 @@ def allocate_destination_data_set(
     # empty dataset was created for the user by an admin/operator, and they don't have permissions
     # to create new datasets.
     # These rules assume that source and destination types are compatible.
+    # Ensure if the dest_exist pass and empty dict
```
rexemin, line 2018: Could you explain this comment a bit more? It's a little confusing to me at the moment.

rexemin (changes requested):
This change should improve the user experience quite a bit. I'd like to echo Fernando's comments about adding assertions to existing tests, it's important to make sure the attributes returned by the module are correct.

fernandofloresg (changes requested):
Hi @AndreMarcel99 I updated the change log description and also renamed it to shorten the name.

Thanks for updating the code.

When I asked to adding tests to our existing ones that create a dest data set, I referred to actually checking that the attribute values returned were same as expected, but adding a flag saying that dests was created is nice too. To assert that values are same as expected we need to know the rules, I don't know them and was surprised when just doing a copy with content: "Example inline content" created a dest data set with record length 1024... I guess that will have its explanation in performance reasons or is just a default.

We will need to check with @rexemin about that, but I think that is over this task because could lead to some debate. Please create a new GH issue with medium priority or similar to add those assertions in the tests.

One last change before approving, please add the new return values into zos_copy module documentation it is under a variable name RETURN at line 555. Sorry didn't spotted that before.

fernandofloresg:
Screenshot 2023-05-29 at 4 25 09 PM

Executed the pipeline and looks like your latest changes are breaking something. error is when running Ansible-test sanity

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
     returned: success
     type: str
     sample: SAMPLE.SEQ.DATA.SET
+dest_created:
+    description: Boolean value that indicate the creation of dest.
```
rexemin, line 567: I think you could rework this to "Indicates whether the module created the destination".

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+    returned: success and if dest is create by the module.
```
rexemin, line 568: Typo here: "is create" instead of "was created".

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+    type: bool
+    sample: true
+destination_attributes:
+    description: Values of a dest create by the module.
```
rexemin, line 572: I think instead of saying "values" you could just say "Attributes", and you could rework it to "Attributes of the destination when created by the module"

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+    returned: success and values of the dest created by module.
```
rexemin, line 573: Wouldn't this information also be returned when "success and destination was created by the module"? I may be wrong on this one.

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+    type: dict
+    contains:
+      block_size:
+        description:
+          Block size of the dataset.
+        type: int
+        sample: 32760
+      record_format:
+        description:
+          Type of record neither binary or not.
```
rexemin, line 583: What do you mean here by "neither binary or not"?

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+        type: str
+        sample: FB
+      record_length:
+        description:
+          Length of the dataset.
```
rexemin, line 588: Here it would be the record length of the dataset, not just the length.

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+        type: int
+        sample: 45
+      space_primary:
```
rexemin, line 591: For `space_primary` and `space_secondary`, I think it would be better to say "Allocated space for the dataset" instead of "Needed size..."

I don't understand well what you mean in `space_type`, but I think you meant to say something like "Unit of measurement for space".

plugins/modules/zos_copy.py
```diff
@@ -1977,6 +2032,45 @@ def is_member_wildcard(src):
...
+def get_attributes_of_any_dataset_created(
+    dest,
+    src_ds_type,
+    src,
+    src_name,
+    is_binary,
+    volume=None
+):
+    """
+    Get the attributes of dataset created by any case to ensure give the information
```
rexemin, line 2044: I'm not sure what you mean in this description.

plugins/modules/zos_copy.py
```diff
@@ -1977,6 +2032,45 @@ def is_member_wildcard(src):
...
+    Arguments:
+        dest (str) -- Name of the destination data set.
+        src_ds_type (str) -- Source of the destination data set.
+        src (str) -- Name of the source data set, used as a model when appropiate.
+        src_name (str) -- Extraction of the source name without the member pattern.
+        is_binary (bool) -- Whether the data set will contain binary data.
+        volume (str, optional) -- Volume where the data set should be allocated into.
+
+    Returns:
+        params (dict) -- Parameters created for the dataset created as name, type,
```
rexemin, line 2056: You're repeating "created" twice, maybe the first time you meant to write "used"?

plugins/modules/zos_copy.py
```diff
@@ -2006,6 +2100,9 @@ def allocate_destination_data_set(
 
     Returns:
         bool -- True if the data set was created, False otherwise.
+        dest_params (dict) -- Parameters created for the dataset created as name,
```
rexemin, line 2103: Same comment as above in `get_attributes_of_any_dataset_created`.

plugins/modules/zos_copy.py
```diff
@@ -2014,8 +2111,10 @@ def allocate_destination_data_set(
     # empty dataset was created for the user by an admin/operator, and they don't have permissions
     # to create new datasets.
     # These rules assume that source and destination types are compatible.
+    # To not generate and None return to the main code generate an empty dict of dest_params to return
```
rexemin, line 2114: So here you're creating an empty dictionary to maintain consistency of what gets returned inside this function? (always returning a boolean and a dictionary)

rexemin (changes requested):

plugins/modules/zos_copy.py
AndreMarcel99: The function is called one time [here](https://github.com/ansible-collections/ibm_zos_core/blob/a8ddfeabdfd91d392a512c7ba10f5d4a69afa082/plugins/modules/zos_copy.py#L2552)

AndreMarcel99:
Captura de pantalla 2023-05-30 a la(s) 21 39 35

fernandofloresg (approved):

plugins/modules/zos_copy.py
rexemin: I still don't quite understand what this comment means.

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
...
+    type: dict
+    contains:
+      block_size:
+        description:
+          Block size of the dataset.
+        type: int
+        sample: 32760
+      record_format:
+        description:
+          Content data type encoding.
```
rexemin, line 583: Here I'm not sure that the record format of a dataset is the encoding for the content it has.

https://www.ibm.com/docs/en/zos-basic-skills?topic=set-data-record-formats

plugins/modules/zos_copy.py
```diff
@@ -563,6 +563,61 @@
     returned: success
     type: str
     sample: SAMPLE.SEQ.DATA.SET
+dest_created:
+    description: Indicates whether the module created the destination.
+    returned: success and if dest was created by the module.
+    type: bool
+    sample: true
+destination_attributes:
+    description: Attributes of a dest create by the module.
```
rexemin, line 572: Type here: "create" instead of "created". Just pointing it out because this will be part of the collection docs.

rexemin (changes requested):

plugins/modules/zos_copy.py
AndreMarcel99: I guess I miss some created in the last change request sorry.

rexemin: I saw that this part of the documentation was already correct and then you introduced the typo over the last commits. It just a matter of reverting that change. Sorry for hammering on this point, I'm just insisting because this will be part of the documentation for the collection, everything else is pretty good and ready to be merged into `dev`.

AndreMarcel99: Ok, I think is done

rexemin (approved):
Thanks for adding this to the module, and for working quickly on all feedback!

PR #778, opened 2023-05-24
347 new query fields

### LINKED ISSUE #347: [Enhancement][zos_job_query] Include new return fields
##### Enhancement
Module zos_job_query could benefit by providing more fields to the client such as job class, service class, priority, asid, queue, status, and job creation date.

##### USE CASE
System programmer wants to cycle a CICS region based on a developers request. A service ticket will kick of an Ansible playbook to determine if the CICS region is running and if so, cycle it, and determine if it started back up successfully. The ability to interrogate the status and queue fields allows the playbook to know if a region is up, down, or held.

System programmer wants to run a playbook to cycle CICS regions that have run for x number of days in service class y. Using the field job creation date and service class allows them to interrogate the fields to accomplish this task.

##### SUMMARY
Enhance module zos_job_query to provide more fields to the client such as job class, service class, priority, asid, queue, status, and job creation date. There are a large number of fields that can be attained and by adding these fields will give system programmers more flexibility and more use cases when automating.

##### DESIGN
I do not have any at this time but I assume (and I know I shouldn't) that when the job is queried, this information is available it just hasn't been coded to return. So hopefully it would not be too difficult to add the new fields.

richp405:
##### SUMMARY

Added new query fields coming from zoau 1.2.2 or later to the job_query return set

Fixes #347 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

zos_job_query
zos_job_submit (which uses query)


richp405:
There are 2 test runs: The only difference between them is whether test_zos_copy_func is included or not... something is happening with storage, where files are periodically unreachable.  I am not sure why zos_copy tests are being run, since I didn't touch that code.

image

fernandofloresg:
Hi @richp405 Looks like the reason why zos_copy tests are included with the dependency finder is because zos_job_submit is being used in one of the zos_copy tests, you can thank me for that (blame me for it).

There is an option in our pipeline to skip certain tests: SKIP_FUNCTIONAL_TESTS, you could add the zos_copy tests path there.

By looking into the Jenkins logs I see only zos_copy tests where executed hence I request you to run a test skipping them if you think they are not needed.

tests/functional/modules/test_zos_data_set_func.py
fernandofloresg, line 1: Unrelated to the issue but trivial enough to be added in the PR

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -30,7 +30,7 @@ def test_zos_job_query_func(ansible_zos_module):
     results = hosts.all.zos_job_query(job_name="*", owner="*")
     pprint(vars(results))
     for result in results.contacted.values():
-        assert result.get("changed") is False
+        # assert result.get("changed") is False
```
fernandofloresg, line 33: Why was this assertion commented? If we are just querying the result of a just changed should be false right?

richp405: I'm going to re-run without the comment (so the assertion is tested).  I checked the code, and it should always return false, so this should be a trivial check.

richp405: There was a problem with the 'skip tests' feature, so I've fixed that and run just the relevant test functions:

image

It passed with the assertion now un-commented.  Thanks for catching that!

richp405: Note, this passed with the following tests in skip_tests field

functional/modules/test_module_security.py
functional/modules/test_zos_copy_func.py
functional/modules/test_zos_encode_func.py
functional/modules/test_zos_fetch_func.py
functional/modules/test_zos_data_set_func.py

There appears to be something out-of-whack with read/write file systems.

plugins/module_utils/job.py
```diff
@@ -232,13 +240,24 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, duration=
...
             job["ret_code"]["msg_code"] = entry.rc
-            # Why was this set to an empty string?
+
             job["ret_code"]["code"] = None
             if len(entry.rc) > 0:
                 if entry.rc.isdigit():
                     job["ret_code"]["code"] = int(entry.rc)
             job["ret_code"]["msg_text"] = entry.status
 
+            # this section only works on zoau 1.2.3 vvv
```
rexemin, line 250: So I went on a little rabbit hole for this, I just learnt [how Python compares strings when it's not just `=`](https://docs.python.org/3/tutorial/datastructures.html#comparing-sequences-and-other-types).

It turns out that if ZOAU decided to release a version 1.2.10, Python would evaluate `"1.2.10" > "1.2.3"` as `False` because it compares character by character. Testing a bit on the REPL, it seems like you  could compare tuples instead: `ZOAU_API_VERSION > (1, 2, 2)`. This way if `ZOAU_API_VERSION` was something like 1.2.xx, it would be evaluated as bigger.

It's probably not really needed because ZOAU plans to release 1.3.0 relatively soon, but I guess it's good to keep it in mind just in case.

plugins/module_utils/job.py
```diff
@@ -200,6 +205,9 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, duration=
     # jls output: owner=job[0], name=job[1], id=job[2], status=job[3], rc=job[4]
     # e.g.: OMVSADM  HELLO    JOB00126 JCLERR   ?
     # listing(job_id, owner) in 1.2.0 has owner param, 1.1 does not
+    # jls output has expanded in zos 1.2.3 and later: jls -l -v shows headers
```
rexemin, line 208: Sorry for nitpicking, this is just a typo: `zos` instead of `zoau`.

rexemin (approved):
Changes look good. I left two comments, neither of them are a big deal though.

plugins/module_utils/job.py
richp405: good grief... that's a newbie mistake, and I made it!  Good catch!

richp405: I re-ran it and realized this case would work correctly.  The feature is added in 1.2.3, so unless they backport the change, it SHOULD failed on 1.2.anything.  1.2.10 is not > 1.2.2

plugins/module_utils/job.py
richp405: don't be sorry.... details like that look good when the customers don't see them.  I've fixed it.

richp405:
did a re-test to make sure there were no mistakes trailing a comment correction.
image

fernandofloresg (approved):
LGTM!

fernandofloresg:
I'm doing a post-review analysis and we are missing a change log fragment for this. @richp405 @rexemin Lets not forget about that next time. I will write that in a moment.

Issue #788, opened 2023-06-02
[Enabler] Update `ac` with new mounts, ansible 2.15 and targets


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

Update the `ac` command to include the new mounts, targets and ansible 2.15

PR #789, opened 2023-06-02
Update ac command supporting files

ddimatos:
##### SUMMARY
Addresses issue #788 
Updates ac command with added mounts, targets and ansible version.


ddimatos:
You can create a new managed venv with:
```
venv-setup:        $ ac --venv-setup --passsword 123456
                   $ ac --venv-setup
```
This will not delete your old managed venv's it will just add a new one if you want a 2.15 ansible, but do note, that this will make 2.15 the default venv used by `ac` thus if you have custom changes in your venv-2.14 , you will need to manually copy them over to venv-2.15.

fernandofloresg (approved):
Tested in my local env and looks ok

rexemin (approved):
Looks good, thanks for updating the ZOAU mounts too.

ddimatos:
Thanks everyone, just remember @rexemin and @fernandofloresg , these updates won't be applied to older managed venv's , either you have to manually update those or delete everything and rebuild. Although you won't know this as .2.15 will be the latestet and that is a new venv so all the scripts will be there for the new mounts. I do plan to build in a `sync` that will update the older ones each time we update these files but that is on plan for me at least anytime soon.

Issue #635, opened 2023-02-12
[Enabler] [CI/CD] Failing to clean up environtment causes a failure.


ddimatos:
I was trouble shooting the pipeline random issue, we have on occasion a failure when creating a VSAM when submitting JCL, to figure this out i had to disable the automatic job purge so that the joblog would remain behind and I could see the error. The job log is below, you will see `"0IDC3026I DUPLICATE DATA SET NAME ON VOLUME",` which is saying VSAM `USER.PRIVATE.TESTDS` was present when we tried to created it. When I looked at volume 00000 I saw the only thing left behind was the INDEX which means our zos_data_set call with state=absent failed to clean up the VSAM correctly. When I run zos_data_set with state=absent in a playbook the VSAM is cleaned up. I am not sure why on our test system this is inconsistent. 

Below you will find everything to help diagnose and move forward with a plan.
I think the test case can be improved without the need to use the shell module, I understand why we use JCL here so that we can test zos_data_set module, but we may need to add some extra checks to ensure the delete happens correctly.

The test case is (partial):
```
def test_data_set_catalog_and_uncatalog(ansible_zos_module, jcl):
    try:
        hosts = ansible_zos_module
        hosts.all.zos_data_set(
            name=DEFAULT_DATA_SET_NAME, state="cataloged", volumes=DEFAULT_VOLUME
        )
        hosts.all.zos_data_set(name=DEFAULT_DATA_SET_NAME, state="absent")
        hosts.all.file(path=TEMP_PATH, state="directory")
        hosts.all.shell(cmd=ECHO_COMMAND.format(quote(jcl), TEMP_PATH))
        results = hosts.all.zos_job_submit(
            src=TEMP_PATH + "/SAMPLE", location="USS", wait=True, wait_time_s=30
        )
```
This is the command I used to clean up the remaining INDEX:
```
echo "  DELETE USER.PRIVATE.TESTDS.INDEX  FILE(DD1) VVR" | mvscmdauth --pgm=IDCAMS --sysprint=* --dd1=000000,vol --sysin=stdin
```
The above command for some reason did not work in a playbook but did not look further:
```
    - name: Delete VSAM
      ansible.builtin.shell: echo "  DELETE USER.PRIVATE.TESTDS.INDEX  FILE(DD1) VVR" | mvscmdauth --pgm=IDCAMS --sysprint=* --dd1=000000,vol --sysin=stdin
      register: result
```

[...]

Issue #631, opened 2023-02-09
[Documentation] [zos_data_set] Update member doc to explain what type of member is created


ddimatos:
### Documentation issue description

The zos_data_set with MEMBER option creates a member, there is no indication of what type of member is created, looking over the source it is a data type member, recall that in a PDS/E you can have program object members or data members but not both in the same PDS/E. 

The doc not stating what kind of member is being created can lead people in a place where if they try to write a LOADLIB or program object would yield an error. 

```type

    The data set type to be used when creating a data set. (e.g pdse)

    MEMBER expects to be used with an existing partitioned data set.

    Choices are case-insensitive.
    required: False
    type: str
    default: PDS
    choices: KSDS, ESDS, RRDS, LDS, SEQ, PDS, PDSE, LIBRARY, BASIC, LARGE, MEMBER, HFS, ZFS
```

Some verbiage about what type of Member is created and a reminder that the member created is a data member and not suitable for a library. 

### IBM z/OS Ansible core Version

v1.4.0-beta.2

### Ansible module

zos_data_set

ddimatos:
For added clarity, i added 
```
            If I(state=present) and I(type=MEMBER) and the member does not exist in the data set,
            create a member formatted to store data, module completes successfully with I(changed=True).
            Note, a PDSE does not allow a mixture of formats such that there is
            executables (program objects) and data. The member created is formatted to store data,
            not an executable.
```

PR #791, opened 2023-06-03
Update zos_data_set member description created

ddimatos:
##### SUMMARY
Given this change will drive over 700 tests it made sense to collapse all the related zos_data_set into one PR to avoid multiple regression pipelines.

1. Documentation: When a MEMBER is created, the format type of the member was not described, it should be clearly defined so that a user does not try to insert an executable into the data formatted member. Issue #631 

2. Bug: A random issue would occur on the pipeline leaving behind VSAM cluster components be it INDEX or DATA, originally this was thought to be a test case issue and a simple test fix but after reviewing I found it much more complex and took a better part of the week to address. To fix this I had to undo most of what was done in  PR [325](https://github.com/ansible-collections/ibm_zos_core/pull/325) for bugfix #140. Although this was a good attempt, the issue was that only one of the cluster components were being deleted here with a `DELETE USER.PRIVATE.TESTDS.DATA FILE(DD1) VVR` while adding the other for INDEX would have been a suitable fix, the issue such a clean up leaves space unclaimed on the volume and not the correct way to back out a data set from z/OS. Thus introduced a flow of tracking state and uncatalog, catalog, delete to achieve this, see image below for most the flow, the flow has since changed with some added exception actions but mostly the flow is correct for reference. Issue #635 

3. The catalog commands were failing because of formatting, I might have been happening for a long time because of all this other related issues it might not have been apparent, thus now the commands are formatted correctly and tested. Issue #802 
 
4. Small .ac command issue where some files are left behind were cleaned up as part of this PR because I was using `./ac --ac-test` and it was inhibiting my ability to work to not fix it. 

Data set deletion flow.
image

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/changelogs/fragments/791-zos_data_set-update-vsam.yml
+++ b/changelogs/fragments/791-zos_data_set-update-vsam.yml
@@ -7,11 +7,11 @@ trivial:
   now removes the unwanted files.
   (https://github.com/ansible-collections/ibm_zos_core/pull/791)
 bugfixes:
-- zos_copy - Reported a failure caused when `present=absent` for a VSAM
+- zos_data_set - Reported a failure caused when `present=absent` for a VSAM
   data set leaving behind cluster components. Fix introduces a new logical
   flow that will evaluate the volumes, compare it to the provided value and
   if necessary catalog and delete.
-  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791).
 - module_utils - data_set.py - Reported a failure caused when cataloging a
   VSAM data set. Fix now corrects how VSAM data sets are cataloged.
-  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791).
--- a/changelogs/fragments/791-zos_data_set-update-vsam.yml
+++ b/changelogs/fragments/791-zos_data_set-update-vsam.yml
@@ -1,0 +1,17 @@
+trivial:
+- zos_data_set - when a member is created by the module, the format is type
+  data which is not suitable for executables. This change describes the
+  format used when creating member.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+- ac - Reported an issue when functional tests ran leaving behind files. Fix
+  now removes the unwanted files.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+bugfixes:
+- zos_copy - Reported a failure caused when `present=absent` for a VSAM
+  data set leaving behind cluster components. Fix introduces a new logical
+  flow that will evaluate the volumes, compare it to the provided value and
+  if necessary catalog and delete.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
+- module_utils - data_set.py - Reported a failure caused when cataloging a
+  VSAM data set. Fix now corrects how VSAM data sets are cataloged.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
--- a/docs/source/modules/zos_job_query.rst
+++ b/docs/source/modules/zos_job_query.rst
@@ -133,19 +133,31 @@ jobs
 
         [
             {
+                "asid": 0,
+                "creation_datetime": "20230503T121300",
+                "job_class": "K",
                 "job_id": "JOB01427",
                 "job_name": "LINKJOB",
                 "owner": "ADMIN",
-                "ret_code": "null"
+                "priority": 1,
+                "queue_position": 3,
+                "ret_code": "null",
+                "svc_class": "?"
             },
             {
+                "asid": 4,
+                "creation_datetime": "20230503T121400",
+                "job_class": "A",
                 "job_id": "JOB16577",
                 "job_name": "LINKCBL",
                 "owner": "ADMIN",
+                "priority": 0,
+                "queue_position": 0,
                 "ret_code": {
                     "code": "null",
                     "msg": "CANCELED"
-                }
+                },
+                "svc_class": "E"
             }
         ]
 
@@ -231,6 +243,41 @@ jobs
         | **type**: int
 
 
+
+  job_class
+    Letter indicating job class for this job.
+
+    | **type**: str
+    | **sample**: A
+
+  svc_class
+    Character indicating service class for this job.
+
+    | **type**: str
+    | **sample**: C
+
+  priority
+    A numeric indicator of the job priority assigned through JES.
+
+    | **type**: int
+    | **sample**: 4
+
+  asid
+    An identifier created by JES.
+
+    | **type**: int
+
+  creation_datetime
+    Date and time, local to the target system, when the job was created.
+
+    | **type**: str
+    | **sample**: 20230504T141500
+
+  queue_position
+    Integer of the position within the job queue where this jobs resided.
+
+    | **type**: int
+    | **sample**: 3
 
 
 message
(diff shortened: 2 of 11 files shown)


ddimatos:
image
Jenkins:
Job suffix `/job/zosAnsibleAlternate2/41/console`
Log:
`====================== 760 passed in 33901.51s (9:25:01) =======================`
Tests:
```
+ ./dependencyfinder.py -p .. -b dev -s functional/modules/test_module_security.py functional/modules/test_zos_apf_func.py -m
[Pipeline] echo
../tests/functional/modules/test_zos_lineinfile_func.py ../tests/functional/modules/test_zos_fetch_func.py ../tests/functional/modules/test_zos_mvs_raw_func.py ../tests/functional/modules/test_zos_blockinfile_func.py ../tests/functional/modules/test_zos_find_func.py ../tests/functional/modules/test_zos_copy_func.py ../tests/functional/modules/test_zos_data_set_func.py ../tests/unit/test_zos_mvs_raw_unit.py ../tests/functional/modules/test_zos_backup_restore.py ../tests/functional/modules/test_zos_job_submit_func.py ../tests/functional/modules/test_zos_job_output_func.py ../tests/functional/modules/test_zos_job_query_func.py ../tests/functional/modules/test_zos_volume_init_func.py ../tests/functional/modules/test_zos_mount_func.py ../tests/functional/modules/test_zos_encode_func.py
```

plugins/module_utils/data_set.py
```diff
@@ -1067,25 +1111,53 @@ def _catalog_vsam(name, volumes):
...
+                data_set_type_vsam,
+            )
+        else:
+            command = DataSet._VSAM_CATALOG_COMMAND_INDEXED.format(
+                data_set_name,
+                DataSet._build_volume_string_idcams(volumes),
+                data_set_type_vsam,
             )
+
+        command_rc, stdout, stderr = module.run_command(
```
fernandofloresg, line 1140: This is not a deal breaker but why to use module.run_command to execute idcams if we have module_utils/mvs_cmd.idcams?

fernandofloresg: Note: Not requesting to change this

plugins/module_utils/data_set.py
```diff
@@ -13,6 +13,7 @@
 
 __metaclass__ = type
 
+import pprint
```
fernandofloresg, line 16: pprint is an unused import, I guess it was added during the debug process

docs/source/modules/zos_job_query.rst
fernandofloresg, line 1: This doc update was not mentioned in the PR I think. @ddimatos Are you sure this should be included?

docs/source/modules/zos_job_submit.rst
fernandofloresg, line 1: This doc update was not mentioned in the PR I think. @ddimatos Are you sure this should be included?

fernandofloresg (changes requested):
Changes requested for changelog and unused import.

Questions about 2 rst files apparently not related to this PR and use of run_command for mvs_cmd

changelogs/fragments/791-zos_data_set-update-vsam.yml
```diff
@@ -0,0 +1,17 @@
+trivial:
+- zos_data_set - when a member is created by the module, the format is type
+  data which is not suitable for executables. This change describes the
+  format used when creating member.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+- ac - Reported an issue when functional tests ran leaving behind files. Fix
+  now removes the unwanted files.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+bugfixes:
+- zos_copy - Reported a failure caused when `present=absent` for a VSAM
```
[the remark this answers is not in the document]
ddimatos: You are correct, its fixed now.

changelogs/fragments/791-zos_data_set-update-vsam.yml
```diff
@@ -0,0 +1,17 @@
...
+  data set leaving behind cluster components. Fix introduces a new logical
+  flow that will evaluate the volumes, compare it to the provided value and
+  if necessary catalog and delete.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
+- module_utils - data_set.py - Reported a failure caused when cataloging a
+  VSAM data set. Fix now corrects how VSAM data sets are cataloged.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
```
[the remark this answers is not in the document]
ddimatos: You are correct, its fixed now.

changelogs/fragments/791-zos_data_set-update-vsam.yml
```diff
@@ -0,0 +1,17 @@
...
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+- ac - Reported an issue when functional tests ran leaving behind files. Fix
+  now removes the unwanted files.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/791)
+bugfixes:
+- zos_copy - Reported a failure caused when `present=absent` for a VSAM
+  data set leaving behind cluster components. Fix introduces a new logical
+  flow that will evaluate the volumes, compare it to the provided value and
+  if necessary catalog and delete.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/762).
```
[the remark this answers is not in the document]
ddimatos: You are correct, it is fixed now.

plugins/module_utils/data_set.py
ddimatos: You are correct it needs to be standardized into one way of interacting with IDCAMs. I left this untouched because we have a item in plan next quarter to find all these and standardize on one way issue #437 , I would then even go as far as looking at mvs_cmd's usage of run_command because its been a bit[ problematic in terms of heap](https://github.com/ansible/ansible/issues/80379) , we might explore the ZOAU Python APIs instead as well. Good point.

docs/source/modules/zos_job_query.rst
ddimatos: I am ok to remove it, the only reason I added it was because during doc gen for the update zos_data_set this showed up (zos_job_query.rst) and really should have been part of that PR that made the change to include the doc. I figure its better to ensure the doc generates before the release process begins and possibly encounter doc generation issues. When you wait till a release to update docs, the error reporting is not great from sphynx making it hard to solve if there are many. Doing the doc updates with each PR reduces the complexity and release tasks can focus on actually releasing. Let me know, I am good either way but they should have been in the PR that change the module, so I picked it up here.

docs/source/modules/zos_job_submit.rst
ddimatos: @fernandofloresg see above response.

plugins/module_utils/data_set.py
fernandofloresg: Thanks for the info

docs/source/modules/zos_job_query.rst
fernandofloresg: It's ok. Good to know about that rst and makes sense to include them in every PR

fernandofloresg (approved):
Is ok for me now, I get that having this in our dev branch will benefit us immediately. mvs_cmd can be standarized later and unused imports actually are shown get using the latest ansible 2.15 sanity scan so we will clean it :slightly_smiling_face: .

ddimatos:
Imports removed from data_sets.py, thought you were referring to the pprint in the test cases, which is referenced by an unused method. Ran sanity on 2.15 it ran clean.

richp405 (approved):
lgtm

ddimatos:
Also includes issue #803 which was left out of the description.

PR #794, opened 2023-06-05
Bug fix, zos_copy returns an error message when a concurrent copy fails

### LINKED ISSUE #586: [Bug] [zos_copy] Return error message when concurrent copy fails
### Bug description

When `zos_copy` is trying to copy a source dataset into a destination dataset that is already in use by another program (with `DISP=SHR`), the module will report a successful task but nothing will actually be copied into the destination. This bug is related to #357, and so while the implementation of supporting other disposition modes is being explored, the module should at least return an error message when this happens.

This is also related to #559, more information about the issue is found there.

### Playbook verbosity output

_No response_

### Contents of `ansible.cfg`

_No response_

### Contents of the inventory

_No response_

_No response_

### Ansible version

```shell
ansible [core 2.13.5]
  config file = /Users/bpanyar/github/IBMZSoftware/ims-containerization/ansible.cfg
  configured module search path = ['/Users/bpanyar/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/bpanyar/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.10.6 (main, Aug 11 2022, 13:49:25) [Clang 13.1.6 (clang-1316.0.21.2.5)]
  jinja version = 3.1.2
  libyaml = True
```

v1.4.0-beta.2

### IBM ZOAU version

v1.1.1

### z/OS version

_No response_

### Ansible module

zos_copy

AndreMarcel99:
##### SUMMARY
Ensures if the target copy or both the origin and target are on use the module throw an error to the user as the machine would be able to perform the action.

Fixes #586 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
The new function validate_disposition is just a checker for any dataset, and return true or false depends on the state of DSP=SHR is at the moment.

##### ADDITIONAL INFORMATION
The function only is requested if both the target and origin are dataset types or at least the target, the past the name of the dataset to validate with the function that use the opercmd` D GRS,RES=(*,{dataset_name})` to check the status and validate it, throw and error message depend on the case of both on use or only the target. 

```
fatal: [zvm]: FAILED! => {"changed:" false, "msg": "DATASET in lock, unable to wrote in 'SOURCE.ISIN'",
 "rc": null, "stderr": null, "stderr_lines": null, "stdout": null, "stdout_line": null}
```
Captura de pantalla 2023-06-09 a la(s) 13 18 38


plugins/modules/zos_copy.py
```diff
@@ -2447,6 +2478,34 @@ def run_module(module, arg_def):
         )
 
     # ********************************************************************
+    # To validate the source and dest are not lock in a batch process by
+    # the machine and not generate a false positive check the disposition
+    # for try to write in dest and if both src and dest are in lock.
+    # ********************************************************************
+    if src_ds_type != "USS" and dest_ds_type != "USS":
+        source_validation = validate_disposition(src_name)
+        dest_validation = validate_disposition(dest_name)
```
fernandofloresg, line 2487: For readability, since your function validate_disposition() returns a boolean, a better name for variables is something like "source_locked" or "is_source_shr".

Other than that this looks ok to me.

fernandofloresg (changes requested):

fernandofloresg (approved):
LGTM!

richp405 (changes requested):
in zos_copy: 2 error messages (minor), but I think there needs to be another else.  See comments for where.

plugins/modules/zos_copy.py
```diff
@@ -2447,6 +2478,34 @@ def run_module(module, arg_def):
...
+                    src_name, dest_name
+                )
+            )
+        elif not is_dest_lock:
+            module.fail_json(
+                msg="DATASET in lock, unable to wrote in '{0}'".format(
+                    dest_name
+                )
+            )
+    elif dest_ds_type != "USS":
```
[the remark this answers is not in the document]
AndreMarcel99: The case of a dataset in DSP=SHR works for read not for writing and if both are in DSP=SHR came a weird error that lock both datasets and you couldn't access to both datasets for some reason.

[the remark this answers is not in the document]
richp405: that's why I asked.

richp405 (approved):
Thank you.  Looks good!

plugins/modules/zos_copy.py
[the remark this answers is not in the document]
ddimatos: When I was looking over this code, I questioned the reason to check if the src is locked and then went back to my notes which I will share here, I believe (i would have to test) that the issue you saw with the error on the source i fit is locked is accurate. Likely the default since we don't support SHR is that data sets are opened as OLD or MOD, thus an exclusive lock, thus with either OLD/MOD no one can access the data set and hence why you need to also check the source, this is good, thanks @AndreMarcel99 , fiy @richp405 

```
OLD - file must exist and you have exclusive control/hold of the dataset, that is, nobody else can read the file while you are using it. You will read from beginning of dataset and if you write, then it will overwrite on existing data. i.e old data is lost.
SHR - file must exist and all others may READ (multiple user can share the data, such then its a read-only file) it at the same time you are reading it (they must have SHR as well)
MOD - file must exist and you have exclusive control/hold of the dataset, will be adding to the end of the data set, without loosing your old data (i.e append the data to the existing file).
NEW - you are creating the data set for the 1st time.
```

plugins/modules/zos_copy.py
```diff
@@ -2245,6 +2245,37 @@ def normalize_line_endings(src, encoding=None):
     return src
 
 
+def source_locked(dataset_name):
```
ddimatos, line 2248: This is not a blocker but only a comment also to the reviewers @fernandofloresg ,@richp405 , @AndreMarcel99 this function would have been a good candidate for a module_utils entry and not solely accessible to zos_copy. Other modules such as zos_blockinfile, zos_lineinfile , zos_apf all write to data sets (even zos_data_set when it creates members) thus we could use this in a more general and broader sense. Something to note when reviewing. Ill open an issue to move it since we are up against a release point.

plugins/modules/zos_copy.py
```diff
@@ -2454,6 +2485,34 @@ def run_module(module, arg_def):
...
+    # To validate the source and dest are not lock in a batch process by
+    # the machine and not generate a false positive check the disposition
+    # for try to write in dest and if both src and dest are in lock.
+    # ********************************************************************
+    if src_ds_type != "USS" and dest_ds_type != "USS":
+        is_source_lock = source_locked(src_name)
+        is_dest_lock = source_locked(dest_name)
+        if not is_source_lock and not is_dest_lock:
+            module.fail_json(
+                msg="DATASETS in lock, unable to access'{0}' without force and unable to write in'{1}'".format(
```
ddimatos, line 2497: I will check out the code and update the message, it does get the point out to the user but we can be a bit more grammatically correct. @jennyhuang17 fyi

ddimatos (commented):
I will checkout the code and make some updates for the messages and comments I provided so we can expidite the merge.

ddimatos (changes requested):

ddimatos:
I would suggest we do a few things here to ensure this can be done for our 1.7.0-beta.1, go over the comments, take note of the ones that are not halting a merge, open a new git issue for those and we address later. For example moving the function `source_locked` to module_utils can be a new work item or adding additional information such as Jobname to the response to the user, from the screen shot you can see that if we extract the jobname we can also tell users what task is keeping the data set locked, that may be very helpful to users. Of course I would code such to avoid null values, for example look at the screen capture, this information is valuable. 
image

plugins/modules/zos_copy.py
jennyhuang17: Thanks for the heads up @ddimatos . I agree we should be careful with the accuracy and the readability of messages that are going to be seen be users. Although it would be a challenge to review messages in codes, let's figure out a way for this... I'll start with searching for IBM guidelines for message writing and share with the team.

Having said that, can you ask the team to invite me to review when there's doc update or message change?

plugins/modules/zos_copy.py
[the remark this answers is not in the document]
richp405: I had resolved because I got an answer to my question... I thought we could read if dsp=shr, so the source query he had was legit.  The source you put up does the same thing, and looks cleaner to me.

plugins/modules/zos_copy.py
ddimatos: Sure Jenny, we can do that.

plugins/modules/zos_copy.py
```diff
@@ -2501,6 +2532,34 @@ def run_module(module, arg_def):
...
+    # To validate the source and dest are not lock in a batch process by
+    # the machine and not generate a false positive check the disposition
+    # for try to write in dest and if both src and dest are in lock.
+    # ********************************************************************
+    if src_ds_type != "USS" and dest_ds_type != "USS":
+        is_source_lock = data_set_locked(src_name)
+        is_dest_lock = data_set_locked(dest_name)
+        if is_source_lock and is_dest_lock:
+            module.fail_json(
+                msg="DATASETS in lock, unable to access'{0}' and unable to write in'{1}'".format(
```
jennyhuang17, line 2544: Is "DATASETS" a parameter name here? If not, be careful of using all uppercases in a message, as it could sound angry to users. Use the lowercase two-word phrase "data sets":

> msgs="Data sets locked, unable to access '{0}' and unable to write in '{1}'"...

(Just to confirm -- {0} and {1} will be replaced by the source name and the destination name when users see the message, right?)

plugins/modules/zos_copy.py
```diff
@@ -2501,6 +2532,34 @@ def run_module(module, arg_def):
...
+                    src_name, dest_name
+                )
+            )
+        elif is_dest_lock:
+            module.fail_json(
+                msg="DATASET in lock, unable to write in '{0}'".format(
+                    dest_name
```
jennyhuang17, line 2551: Same as above. Change to:

> msg="Data set locked, unable to write in '{0}'"...

plugins/modules/zos_copy.py
```diff
@@ -2501,6 +2532,34 @@ def run_module(module, arg_def):
...
+                )
+            )
+    elif dest_ds_type != "USS":
+        is_dest_lock = data_set_locked(dest_name)
+        if is_dest_lock:
+            module.fail_json(
+                msg="DATASET in lock, unable to write in '{0}'".format(
+                    dest_name
```
jennyhuang17, line 2559: Same as above. Change to:

> msg="Data set locked, unable to write in '{0}'"...

jennyhuang17 (changes requested):
Please see my comments and address the feedback as needed

plugins/modules/zos_copy.py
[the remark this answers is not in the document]
ddimatos: I did some testing, i will put my steps here as i had share in slack as well:

My test:

SSH , setup:
```
dtouch ibmuser.pdse.a
dtouch ibmuser.pdse.b
decho 'hello' 'ibmuser.pdse.a(one)'
decho 'hello' 'ibmuser.pdse.b(one)'
```

ISPF:
1. ISPF 3.4 > ibmuser.pdse.a
2. e next to ibmuser.pdse.a
3. FN F2 (split screen) 
4. ISPF 3.4 > ibmuser.pdse.b
5. e next to ibmuser.pdse.b

**Copy test 1- Both src & dest locked:**
- `cp "//'ibmuser.pdse.a(one)'" "//'ibmuser.pdse.b(onecopy)'"`
- `cp: FSUM6259 target file “//‘ibmuser.pdse.b(onecopy)’“: EDC5061I An error occurred when attempting to define a file to the system.`

**Copy test 2 - src locked, dest not locked:**
- `cp "//'ibmuser.pdse.a(one)'" "//'ibmuser.pdse.b(onecopy2)'"`
- `Success`

**Copy test 3 - src unlocked, dest locked:**
- `cp "//'ibmuser.pdse.a(one)'" "//'ibmuser.pdse.b(onecopy3)'"`
- `cp: FSUM6259 target file “//‘ibmuser.pdse.b(onecopy3)’“: EDC5061I An error occurred when attempting to define a file to the system.`

**Copy test 4 - src unlocked, dest unlocked:**
- `cp "//'ibmuser.pdse.a(one)'" "//'ibmuser.pdse.b(onecopy4)'"`
- `Success`

What this indicates to me is **only** the `dest`  has to be monitored for a lock, not the `src`, not the `src && dest`, just the `dest`:
- There is no issue with a ISPF locking up if both data sets are in disp=shr (locked) and a copy (cp ) is in progress
- There is no need to have the this double check if src_ds_type != "USS" and dest_ds_type != "USS":
- There is no need to monitor src (no one is doing it, i am just stating that its not needed. 

The reason **Copy test 1** failed is because the `dest` is locked, it had nothing to do with both `src && dest` being locked, its the same reason an error in **Copy test 3 failed**, because both these tests locked the `dest`.

The Pull request will need updating. 

Result of ISPF browse to prove the results match the tests:
image
image

ddimatos (changes requested):

plugins/modules/zos_copy.py
```diff
@@ -2245,6 +2245,37 @@ def normalize_line_endings(src, encoding=None):
...
+    result = dict()
+    result["stdout"] = []
+    command_dgrs = "D GRS,RES=(*,{0})".format(dataset_name)
+    response = opercmd.execute(command=command_dgrs)
+    stdout = response.stdout_response
+    if stdout is not None:
+        for out in stdout.split("\n"):
+            if out:
+                result["stdout"].append(out)
+    if len(result["stdout"]) > 4 and "EXC/SHR" in stdout and "SHARE" in stdout:
```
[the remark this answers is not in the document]
AndreMarcel99: The original name was the reason, but already change and now with function call `data_set_locked` apply the logic you mention.

plugins/modules/zos_copy.py
```diff
@@ -2447,6 +2478,34 @@ def run_module(module, arg_def):
...
+        is_dest_lock = source_locked(dest_name)
+        if not is_source_lock and not is_dest_lock:
+            module.fail_json(
+                msg="DATASETS in lock, unable to access'{0}' without force and unable to write in'{1}'".format(
+                    src_name, dest_name
+                )
+            )
+        elif not is_dest_lock:
+            module.fail_json(
+                msg="DATASET in lock, unable to wrote in '{0}'".format(
```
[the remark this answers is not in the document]
ddimatos: I am resolving this as It has been updated.

plugins/modules/zos_copy.py
ddimatos: Resolved

plugins/modules/zos_copy.py
ddimatos: Resolved

ddimatos (approved):
Looks good, and thank you for cooperating on the changes , this looks great @AndreMarcel99

PR #805, opened 2023-06-12
Change the line for the functional one

AndreMarcel99:
##### SUMMARY
Remove unnecessary line and change for functional.


fernandofloresg:
FYI removing ignore_cleanup_errors = True from that test is needed because regression will fail, it was needed in a different version and from doing a git revert it ended up there.

fernandofloresg (approved):

rexemin (approved):
Thanks for catching this so quickly.

PR #810, opened 2023-06-15
Bugfix for #807 - zos_copy ignores encoding for binary files

### LINKED ISSUE #807: [Bug] [zos_copy] Encoding normalization applies to binary files

v1.2.2

### IBM Enterprise Python

v3.10.x

v1.5.0

### ansible-version

v2.14.x

### z/OS version

v2.3

### Ansible module

zos_copy

### Bug description

When copying a binary file from USS, the code that will normalize encoding for USS files gets executed. This code should only apply to text files, never to binary ones, so the module terminates with an error:
```json
{"changed": false, "msg": "An error occurred during encoding: \"iconv: Conversion from codeset \"binary\" to \"UTF-8\": EDC5121I Invalid argument.\n\"", "rc": null, "stderr": null, "stderr_lines": null, "stdout": null, "stdout_lines": null}
```

This error happens if the source file is tagged as `binary` and even when the task specifies `is_binary: true`.

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

rexemin:
##### SUMMARY
Fixes #807. Code introduced in 1.5.0 in `zos_copy` tried to convert the encoding of binary files, resulting in an error when the file was tagged as `binary` on z/OS or invalid bytes for UTF-8 were read.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
zos_copy

##### ADDITIONAL INFORMATION
Added a couple of test cases to validate the handling of binary files.

### DIFF
--- a/changelogs/fragments/810_fix_binary_file_bypass.yml
+++ b/changelogs/fragments/810_fix_binary_file_bypass.yml
@@ -1,5 +1,5 @@
 bugfixes:
 - zos_copy - Encoding normalization used to handle newlines in
-  text files was applied to binary files too. Fix now makes sure
+  text files was applied to binary files too. Fix makes sure
   that binary files bypass this normalization.
   (https://github.com/ansible-collections/ibm_zos_core/pull/810)
--- a/changelogs/fragments/810_fix_binary_file_bypass.yml
+++ b/changelogs/fragments/810_fix_binary_file_bypass.yml
@@ -1,0 +1,5 @@
+bugfixes:
+- zos_copy - Encoding normalization used to handle newlines in
+  text files was applied to binary files too. Fix now makes sure
+  that binary files bypass this normalization.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/810)
--- a/tests/functional/modules/test_zos_copy_func.py
+++ b/tests/functional/modules/test_zos_copy_func.py
@@ -43,6 +43,11 @@ DUMMY DATA ---- LINE 007 ------
 
 DUMMY_DATA_CRLF = b"00000001 DUMMY DATA\r\n00000002 DUMMY DATA\r\n"
 
+# FD is outside of the range of UTF-8, so it should be useful when testing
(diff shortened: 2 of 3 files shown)


rexemin:
Screenshot 2023-06-19 at 11 28 48
Screenshot 2023-06-19 at 11 29 16

richp405 (approved):
Nice, clean, straightforward.

tests/functional/modules/test_zos_copy_func.py
AndreMarcel99, line 1: Personal view is to add a assertion to both cases to verify the content is fine.

changelogs/fragments/810_fix_binary_file_bypass.yml
AndreMarcel99, line 1: Is better: 
Fix makes sure that binary files bypass this normalization

AndreMarcel99 (changes requested):

tests/functional/modules/test_zos_copy_func.py
rexemin: Agree that verifying the content should be in the tests, but because the destinations are binary datasets, the output that we can get from them is going to be finicky, and ZOAU 1.2.2 doesn't allow reading a dataset as a binary stream.

I decided to use the byte `FD` because it should trigger an error whenever the code tries to read it as text.

changelogs/fragments/810_fix_binary_file_bypass.yml
rexemin: Updated the changelog accordingly.

tests/functional/modules/test_zos_copy_func.py
AndreMarcel99: Ok thanks for the answer, so for dev should be there the content.

AndreMarcel99:
Thanks for the answer and the changes.

ddimatos:
Looks good thank you @rexemin

Issue #779, opened 2023-05-25
[Bug] When running the latest release we see a warning PlayContext.verbosity is deprecated. 


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Are the dependencies a supported version?

- [X] The dependencies are supported.

### IBM Z Open Automation Utilities

v1.2.2

### IBM Enterprise Python

v3.9.x

### IBM z/OS Ansible core Version

v1.5.0

### ansible-version

v2.14.x

### z/OS version

v2.5

### Ansible module

zos_copy

### Bug description

It was reported with some details that when the user ran the `zos_copy` module they encountered this warning:
```
[DEPRECATION WARNING]: PlayContext.verbosity is deprecated, use ansible.utils.display.Display.verbosity instead. This feature will be removed in version 2.18. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
````

My assumption is that the `zos_copy` and likely `zos_fetch` plugins which use the ansible display() methods to allow for debug messages in the action plugins needs to be updated according to the deprecation. 

Although this is likely happening with or without verbosity/debug, if you want to see the code in action run the module with DEBUG like so:
```
ANSIBLE_DEBUG=1 ansible-playbook -i inventory <playbook>
```
You will see our messages; that code is likely the source of the warning, there is conditional clauses for various ansible versions that need to probably be updated. 

### Playbook verbosity output.

NA

### Ansible configuration.

```YAML
NA
```

### Contents of the inventory

```YAML
NA
```

### Contents of `group_vars` or `host_vars`

```YAML
NA
```

fernandofloresg:
Hey Demetri, was is the suggested course of action for this? If version < 2.13 keep verbosity and if version > 2.14 add ansible.utils.display.Display().verbosity? 

This is taken from https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_7.html

If so, since I'm using this for zos_unarchive to copy the files I'm seeing this issue and can take over this task.

ddimatos:
Initial prototype working, partial result shown:
```
<ec01150a.vmec.svl.ibm.com> play context verbosity: 4
<ec01150a.vmec.svl.ibm.com> ibm_zos_copy SSH transfer method restored to None
```

Fix is quite easy, detect the ansible-core version, then assign the verbosity to a var and propagate the var's usage over a direct call to `display.verbosity` or `_play_context.verbosity`

```
[...]

PR #814, opened 2023-06-16
Bugfix: Fixes zos_copy and zos_fetch deprecation msg for using _play_context.verbosity

ddimatos:
Same as PR #806 but for release version 1.6.0.

##### SUMMARY
This addresses issue #779 where a deprecation warning would appear when zos_copy or zos_fetch ran. Of the two modules, only the `zos_copy` module  could be automated with a test because its plugin is invoked immediately where as with `zos_fetch` its invoked last and that can not work for testing display.verbosity.

In addition to this, the test case for #616 was addressed here because the original test case could not run properly on local work stations and the pipeline. So a new test case was written to run on both.

The pytest-ansible fixture was updated to support the `zos_copy` adhoc test case.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


rexemin:
Screenshot 2023-06-19 at 20 01 30
Screenshot 2023-06-19 at 20 01 45

ddimatos:
Thank you @rexemin your help getting this through was much appreciated.

AndreMarcel99 (approved):
Ready

changelogs/fragments/814-zos_data_set-update-vsam-copy.yml
```diff
@@ -0,0 +1,7 @@
+bugfixes:changelogs/fragments/814-zos_data_set-update-vsam-copy.yml
```
fernandofloresg, line 1: This will break change log generation

fernandofloresg (changes requested):

changelogs/fragments/814-zos_data_set-update-vsam-copy.yml
ddimatos: All fixed

fernandofloresg (approved):
🦭 of approval.

Issue #802, opened 2023-06-09
[Bug] zos_data_set data_set.py utils cataloging failures


ddimatos:
When cataloging a VSAM data set , some liberties were taken because when a data set is uncatalogued you must know its re-catalog attribute ("LINEAR", "INDEXED", "NONINDEXED", "NUMBERED"). 

The [code](https://github.com/ansible-collections/ibm_zos_core/blob/98c70475ae959caae753bde9282500982a7340d6/plugins/module_utils/data_set.py#L1070) would loop over the types but and tried too on chance find the right type to catalog. If it hit an exception it would try it differently, because of the ordering , an indexed vsam might only have part of its cluster re-cataloged (eg DATA or INDEX) often leaving behind portions of the data set causing issues with catalog. 

The second issue is the command syntax was incorrect in its formatting and would also contribute to these random occurrences.
Below:
```
    _VSAM_CATALOG_COMMAND_NOT_INDEXED = """ DEFINE CLUSTER -
        (NAME('{0}') -
        VOLUMES({1} -
        ) -
        RECATALOG -
        {2}) -
    DATA( -
        NAME('{0}.DATA'))
    """

    _VSAM_CATALOG_COMMAND_INDEXED = """ DEFINE CLUSTER -
        (NAME('{0}') -
        VOLUMES({1} -
        ) -
        RECATALOG -
        {2}) -
    DATA( -
        NAME('{0}.DATA')) -
    INDEX( -
        NAME('{0}.INDEX'))
    """
```

PR #816, opened 2023-06-16
Update zos_data_set member description created

ddimatos:
Same as PR #791 but for v1.6.0 (backport)

##### SUMMARY
Given this change will drive over 700 tests it made sense to collapse all the related zos_data_set into one PR to avoid multiple regression pipelines.

1. Documentation: When a MEMBER is created, the format type of the member was not described, it should be clearly defined so that a user does not try to insert an executable into the data formatted member. Issue #631 

2. Bug: A random issue would occur on the pipeline leaving behind VSAM cluster components be it INDEX or DATA, originally this was thought to be a test case issue and a simple test fix but after reviewing I found it much more complex and took a better part of the week to address. To fix this I had to undo most of what was done in  PR [325](https://github.com/ansible-collections/ibm_zos_core/pull/325) for bugfix #140. Although this was a good attempt, the issue was that only one of the cluster components were being deleted here with a `DELETE USER.PRIVATE.TESTDS.DATA FILE(DD1) VVR` while adding the other for INDEX would have been a suitable fix, the issue such a clean up leaves space unclaimed on the volume and not the correct way to back out a data set from z/OS. Thus introduced a flow of tracking state and uncatalog, catalog, delete to achieve this, see image below for most the flow, the flow has since changed with some added exception actions but mostly the flow is correct for reference. Issue #635 

3. The catalog commands were failing because of formatting, I might have been happening for a long time because of all this other related issues it might not have been apparent, thus now the commands are formatted correctly and tested. Issue #802 
 

Data set deletion flow.
image

##### ISSUE TYPE
- Docs Pull Request

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/source/modules/zos_data_set.rst
+++ b/docs/source/modules/zos_data_set.rst
@@ -56,7 +56,7 @@ state
   If *state=absent* and *volumes* is provided, and the data set is not found in the catalog, the module attempts to perform catalog using supplied *name* and *volumes*. If the attempt to catalog the data set catalog fails, then no action is taken. Module completes successfully with *changed=False*.
 
 
-  If *state=absent* and *volumes* is provided, and the data set is found in the catalog, the module compares the catalog volume attributes to the provided *volumes*. If they volume attributes are different, the cataloged data set will be uncataloged temporarily while the requested data set be deleted is cataloged. The module will catalog the original data set on completion, if the attempts to catalog fail, no action is taken. Module completes successfully with *changed=False*.
+  If *state=absent* and *volumes* is provided, and the data set is found in the catalog, the module compares the catalog volume attributes to the provided *volumes*. If the volume attributes are different, the cataloged data set will be uncataloged temporarily while the requested data set be deleted is cataloged. The module will catalog the original data set on completion, if the attempts to catalog fail, no action is taken. Module completes successfully with *changed=False*.
 
 
   If *state=present* and the data set does not exist on the managed node, create and catalog the data set, module completes successfully with *changed=True*.
--- a/docs/source/modules/zos_data_set.rst
+++ b/docs/source/modules/zos_data_set.rst
@@ -65,6 +65,9 @@ state
   If *state=present* and *replace=False* and the data set is present on the managed node, no action taken, module completes successfully with *changed=False*.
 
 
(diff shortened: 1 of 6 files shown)


rexemin:
Screenshot 2023-06-20 at 8 45 00
Screenshot 2023-06-20 at 8 45 13

docs/source/modules/zos_data_set.rst
```diff
@@ -321,6 +327,9 @@ batch
     If *state=absent* and *volumes* is provided, and the data set is not found in the catalog, the module attempts to perform catalog using supplied *name* and *volumes*. If the attempt to catalog the data set catalog fails, then no action is taken. Module completes successfully with *changed=False*.
 
 
+    If *state=absent* and *volumes* is provided, and the data set is found in the catalog, the module compares the catalog volume attributes to the provided *volumes*. If they volume attributes are different, the cataloged data set will be uncataloged temporarily while the requested data set be deleted is cataloged. The module will catalog the original data set on completion, if the attempts to catalog fail, no action is taken. Module completes successfully with *changed=False*.
```
richp405, line 330: Second sentence: "If they volume" should be "If the volume"

tests/functional/modules/test_zos_data_set_func.py
```diff
@@ -498,7 +505,7 @@ def test_data_member_force_delete(ansible_zos_module):
         for result in results.contacted.values():
             assert result.get("changed") is True
 
-        # add members
+        #add members
```
richp405, line 508: I thought you had to have a space between a comment mark and the comment text.  I think that's a pep8 issue.

richp405 (changes requested):
1 minor typo, and one maybe lint/pep8 issue in a crushed comment.  Otherwise it looks good!

tests/functional/modules/test_zos_data_set_func.py
ddimatos: You do, not sure why the lint missed it, its fixed now.

docs/source/modules/zos_data_set.rst
ddimatos: Thanks, its fixed now.

richp405 (approved):
Thank you, sir.  LGTM!

fernandofloresg (approved):
Reviewed previously for dev branch looks ok to me

PR #821, opened 2023-06-19
439 addf

richp405:
##### SUMMARY
Add F parameter for data set creation, and in validation and inline docs.
Add test that uses F parameter and checks for errors.

Fixes 439

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_data_set, data_set, and DataSet object

##### ADDITIONAL INFORMATION
Eliminates error message for record type='F', and passes value through to zoau.

```paste below

```


richp405:
image

Exclusions (things I didn't change, so the  builds would actually complete:
test_module_security.py,test_zos_blockinfile_func.py,
test_zos_find_func.py,/test_zos_mvs_raw_unit.py,test_zos_volume_init_func.py,test_zos_mvs_raw_func.py,
test_zos_backup_restore.py,test_zos_job_submit_func.py,test_zos_job_query_func.py,test_zos_mount_func.py,
test_zos_copy_func.py,modules/test_zos_fetch_func.py,test_zos_lineinfile_func.py,test_zos_job_output_func.py,
test_zos_encode_func.py,test_zos_apf_func.py

tests/functional/modules/test_zos_data_set_func.py
fernandofloresg, line 1: Looks good to me. This request might be out of scope, but could you instead parametrize this to test creating data sets with different formats? I would assume that test will be in place but didn't find it

fernandofloresg (changes requested):

rexemin (approved):
Looks good, thanks for adding it!

tests/functional/modules/test_zos_data_set_func.py
richp405: It was not in there.  I was hoping to just add 'F', but the other formats were not being tested.  Let me see if I can add them into a parameter loop.

richp405:
Added a create test for each record_type in the test.

image

fernandofloresg (approved):
Great !

PR #826, opened 2023-06-22
696dataint

richp405:
##### SUMMARY
Removal of automated backup... manual backup will still work

Addresses github 696

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_copy


richp405:
image

testing was purely on the zos_copy function.  Backups are disabled unless both backup=True and backup_name is provided.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -534,8 +534,8 @@ def test_copy_dir_to_existing_uss_dir_not_forced(ansible_zos_module):
         for result in copy_result.contacted.values():
             assert result.get("msg") is not None
             assert result.get("changed") is False
-            assert "Error" in result.get("msg")
-            assert "EDC5117I" in result.get("stdout")
+            assert "error" in result.get("msg").lower()
+            # assert "EDC5117I" in result.get("stdout")  # No longer valid, because different utility was used
```
ddimatos, line 538: I find this interesting that `EDC5117I` will not be seen, its ok we don't get that specific but this causes concern to me because that error is a result of mkdir failing, see [here](https://www.ibm.com/docs/en/zos/2.2.0?topic=SSLTBW_2.2.0/com.ibm.zos.v2r2.ceea900/cs00223.html)

The test case name is test_copy_dir_to_existing_uss_dir_not_forced which implies it is going to copy to a existing directory without force and expected to fail, probably by only calling a mkdir() which is why we saw the error. Ideally i would have rather had a check done first to see if the directory exists before calling a mkdir which is what I can hypothesize without following the code path this test is intending to drive. 

I wonder what changed to not make the EDC5117I not appear.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1174,26 +1174,37 @@ def test_backup_uss_file(ansible_zos_module, backup):
     src = "/etc/profile"
     dest = "/tmp/profile"
     backup_name = None
+    if backup:
```
ddimatos, line 1177: We would want to double check this change here after the original logic is reverted back meaning a user can specify backup:true , or backup:true and backup_name:foo

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1571,6 +1582,33 @@ def test_copy_dest_lock(ansible_zos_module):
         hosts.all.zos_data_set(name=DATASET_2, state="absent")
 
 
+@pytest.mark.uss
+@pytest.mark.parametrize("src", [
+    dict(src="/etc/", is_remote=False),
+    dict(src="/etc/", is_remote=True),])
+def test_ensure_copy_directory_does_not_change_permission_on_dest(ansible_zos_module, src):
```
ddimatos, line 1589: Nice to see a test case added but it seems unrelated to the git issue, i just wonder where it came from, was it intentionally added, trying to understand how I should evaluate this.

ddimatos (changes requested):
@richp405 - I think there as a misunderstanding of the issue, it appears we undid more than was needed and negated some of our contractual agreements with the user. 

The issue really was just to undo the emergency logic, the restore_backup and erease_backup logic, we did not intend to change the backup:true and backup_name:"..." logic. 

Feel free to reach out to me, or @rexemin or @fernandofloresg who actually worked on this, but to me that is all I think needed to be done. I wonder if its easier to start from a fresh branch. Happy to discuss this over all call.

richp405:
I'll re-do this on a fresh branch.

PR #828, opened 2023-06-23
Initial commit for load balancer

ddimatos:
##### SUMMARY
This PR while rather large introduces the new concurrent executor utility capable of discovering all functional and unit tests and running individually on a cluster of z/OS managed nodes. As a byproduct of this work, many other updates needed to be made such as integration in AC and all this triggered new capabilities and additional function outside the scope of the original issue. 

To use this and test it you will need to rebuild your managed `ac` venv's with `ac --venv-setup --password 123456`

**For the AC (./ac) tool, this introduces:**
1. `test-dep-find` - Determine which test suites to run given the options selected, drives dependency finder tool.
2. `test-pytest-find` - Get a list of parametizd test cases used by pytest.
3. `venv-stop` (this was fixed as it never was able to shut down a managed venv properly) - shut down a managed AC venv.
4. `test-concurrent`: - Run the conncurrent executor (CE) that can drive test cases to a cluster of hosts
5. Trivial changes:
- Added return code capabilities for `ac-test`
- Added new print type helpers to ease differentiation between stdout and stderr
- Added colorized output, currently using RED, YELLOW, GREEN (although other colors are supported) to visually aid in reading stdout. 
- Enhanced auto doc function - added support for commands that are long and need wrapping and proper indentation

**The scripts directory:**
- This was getting a bit crowed so I introduced new folders and grouped some but not all content such as python modules, requirements-xx.env  and tests for CE. 
- AC and other scripts such as venv.sh were update with new folder paths.
- requirements-xx.env had a few new packages added to it and updated dependencies as regular maintenance such as `ansible-core`, `galaxy-importer`, `ansible-lint`, etc

**For the concurrent executor (ce.py), this introduces:**
- The utility that partially minimally depends on `venv.sh` functions for decrypting the managed node locations
- Auto discovery of a pool of managed z/OS nodes
- Manual override of managed nodes in case you want to bring your own.
- Test discovery (full or partial depending on the options submitted)
- Nearly 21 options to granularly control the CE behavior, whether you want iterative testing, throttled, replays, etc. These options allow CE to be used as a full regression bucket executor or as minimal as 1 test case execution. 
- Log retention, options to manage the logs, whether to send to stdout , html, text and/or include a return code for scripted valiation. 

**pytest' Fixture: `--zinventory-raw`:**
- A  new pytest fixture `--ziventory-raw` which can take a JSON string as an argument to increase the performance the concurrent executors usage of the fixture. 
- This was introduced this time as the original prototype of CE was relying on ./ac --ac-test and the file I/O was hindering performance so a new model was created which uses no I/O and only a JSON string to feed to pytest. 
- This pytest fixture also includes now the `extra_args` option to manage additional options we might not have planned for such as volumes. 

For example:
```
pytest tests/functional/modules/test_zos_apf_func.py::test_add_del --host-pattern=all --zinventory-raw='{"host": "zvm.ibm.com", "user": "ibmuser", "zoau": "/zoau/v1.3.1", "pyz": "/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz", "pythonpath": "/zoau/v1.3.1/lib/3.10", "extra_args":{"volumes":["222222","000000"],"other":"something else"}}'
```
Or
```
pytest tests/functional/modules/test_zos_mount_func.py::test_basic_mount --host-pattern=all --zinventory-raw='{"host": "zvm.ibm.com", "user": "ibmuser", "zoau": "/zoau/v1.3.1", "pyz": "/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz", "pythonpath": "/zoau/v1.3.1/lib/3.10", "extra_args":{"volumes":["222222","000000"],"other":"something else"}}' 
```
The fixture generates a dictionary similar to:
```
{
    "host":"zvm.ibm.com",
    "user":"ibmuser",
    "zoau":"/zoau/v1.3.1",
    "pyz":"/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz",
    "pythonpath": "/zoau/v1.3.1/lib/3.10",
    "extra_args":{
        "volumes":[
            "vol1",
            "vol2"
        ],
        "other": "something else" }
}
```

**Test exercising `volumes_on_systems`:**
```
============================================================================================ 1 passed, 11 warnings in 32.66s =============================================================================================
(venv-2.16) pytest tests/functional/modules/test_zos_mount_func.py::test_basic_mount --host-pattern=all --zinventory-raw='{"host": "ec33025a.vmec.svl.ibm.com", "user": "omvsadm", "zoau": "/zoau/v1.3.1", "pyz": "/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz", "pythonpath": "/zoau/v1.3.1/lib/3.10", "extra_args":{"volumes":["222222","000000"],"myscret":"not so secret if I am showing you"}}'
================================================================================================== test session starts ===================================================================================================
platform darwin -- Python 3.12.4, pytest-8.1.1, pluggy-1.5.0
ansible: 2.16.3
rootdir: /Users/ddimatos/git/gh/ibm_zos_core/tests
configfile: pytest.ini
plugins: ansible-24.1.3, mock-3.14.0
collected 1 item

tests/functional/modules/test_zos_mount_func.py .                                                                                                                                                                  [100%]
============================================================================================ 1 passed, 12 warnings in 51.70s =============================================================================================
```

Test exercising `volumes_with_vvds`:
```
(venv-2.16) pytest tests/functional/modules/test_zos_apf_func.py::test_add_del --host-pattern=all --zinventory-raw='{"host": "ec33025a.vmec.svl.ibm.com", "user": "omvsadm", "zoau": "/zoau/v1.3.1", "pyz": "/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz", "pythonpath": "/zoau/v1.3.1/lib/3.10", "extra_args":{"volumes":["222222","000000"],"myscret":"not so secret if I am showing you"}}'
================================================================================================== test session starts ===================================================================================================
platform darwin -- Python 3.12.4, pytest-8.1.1, pluggy-1.5.0
ansible: 2.16.3
rootdir: /Users/ddimatos/git/gh/ibm_zos_core/tests
configfile: pytest.ini
plugins: ansible-24.1.3, mock-3.14.0
collected 1 item

tests/functional/modules/test_zos_apf_func.py .                                                                                                                                                                    [100%]
============================================================================================ 1 passed, 11 warnings in 36.23s
````

**Stdout and Stderr**
Some parts of CE will write to stdout or stderr, while its a bit difficult to know if this actually works, the concept I applied to validate the  code can be seen with this output below:

```
~ $: python3
>>> import sys
# Print to stderr test before any changes
>>> print("FOO", file=sys.stderr)
FOO
# Print to stdout test before any changes
>>> print("FOO", file=sys.stdout)
FOO

# Create a nothing class to writes nothing to anyplace
>>> class write_nothing():
...     def write(self, s):
...             pass

# Create a new  a file-like object (stream) and point where sys.stderr is pointed
>>> mysys_stderr=sys.stderr
# Use my file stream just to prove a point
>>> print("FOO", file=mysys_stderr)
FOO

# Now that we know my file stream works, lets set it to a class that does nothing and try again.
>>> mysys_stderr=nothing()
# Notice nothing comes out of the print function
>>> print("FOO", file=mysys_stderr)

# Test stdout just for curiosity, works just fine
>>> print("FOO", file=sys.stdout)
FOO

# Restore my file stream `mysys_stderr` and it works fine.
>>> mysys_stderr=sys.stderr
>>> print("FOO", file=mysys_stderr)
FOO
```
##### ISSUE TYPE

### CODE REVIEW

**Review by richp405 — APPROVED** [review]
I didn't detail the load_balance script (not focusing well at the moment).  Everything else looks fine, although I'm surprised about some of the systems you marked as 'prod', since that means that no other builds can safely run when this is distributing.

**Comment by ddimatos** [comment]
> I didn't detail the load_balance script (not focusing well at the moment). Everything else looks fine, although I'm surprised about some of the systems you marked as 'prod', since that means that no other builds can safely run when this is distributing.

Thanks @richp405 , the load balancer can't go into effect until after this Q3 when the test cases are cleaned up, how I will prevent the load balancer from clobbering what the pipeline is using is by making an edit in our pytest fixture to check for a file lock, eg /tmp/balancer_active.txt , the pytest fixter can check that and not run on the z/OS, also why the code has a number of ssh methods to help write this file lock. 

The second way is that the pipelines update and always use the load balancer after we have worked out the issues bugs, but even then multiple balancers running need a way to know and this file lock will help. 

The solution is no fully flushed out and is more a prototype at the moment, where I would put it at about 90% - 95% complete, it still has more work which I can do at will over the next quarter. I do need it checked in so that its not a risk of being lost and periodic updates to it also need to go in ./ac as well. 

Hope that explains why those were marked as production, also I will be ordering about 30 new z/OS targets :)

**Comment by ddimatos** [comment]
@richp405 - for example , today it relies on the dependency finder and divides those tests cases but does not support a user specifying a suite to be divided which is something of value to development nor a way to override with your own set of z/OS instances should they want to load balance a change without all the other test cases.

**Comment by ddimatos** [comment]
@richp405 also commenting the point that no other pipeline could use the z/OS managed nodes if the are in use by the concurrent executor. Things I am working on are to increase virtual storage of each node, and before I even started that, I also validated that they can in fact handle multiple work loads, what I have seen is with multiple connections thus far on machines with lower amounts of virtual storage we tend to get AnsibleConnectionFailures, with the updated machines I am not seeing this, while its not proof its a solution yet, it seems to help.

I ran multiple JCL jobs on our zVM not using Ansible and they were fine, so they can as they are handle more than one ansible task assuming they don't get a connection failure. I did this with FTP.

You cans see below there are 3 jobs submitted at once, actually was more like 10 jobs but only showing 3. 
```
ftp> site
(arguments to SITE command) file=jes
200 SITE command was accepted
ftp> put 'hello.jcl'
ftp> put hello.jcl
200 Port request OK.
125 Sending Job to JES internal reader FIXrecfm 80
250-It is known to JES as JOB00328
250 Transfer completed successfully.
921 bytes sent in 0.00188 seconds (477 kbytes/s)
ftp> put hello.jcl
200 Port request OK.
125 Sending Job to JES internal reader FIXrecfm 80
250-It is known to JES as JOB00329
250 Transfer completed successfully.
921 bytes sent in 0.00173 seconds (520 kbytes/s)
ftp> put hello.jcl
200 Port request OK.
125 Sending Job to JES internal reader FIXrecfm 80
250-It is known to JES as JOB00330
250 Transfer completed successfully.
921 bytes sent in 0.000318 seconds (2.76 Mbytes/s)
```

**Comment by ddimatos** [comment]
Generally this is isolated code but because I introduced a new pytest fixture --zinventory-raw` and altered the behavior or the original `--zinventory` I have been running local pipelines and in this PR I ran a test suite to ensure that the original fixture was not broken by my changes.

The only reason this pipelined failed and will continue is because the changelog fragment check's logic is not ideal, it appears to rely on the branch name and a matching fragment file which is not the pattern we follow, I have an idea how to fix this check, i will do a new PR and added a new `./ac` feature to perform this operation for the pipeline. 

Below is the evidence of the check that is causing my failure:
```
+ ls *827*
+ wc -l
ls: cannot access '*827*': No such file or directory
[Pipeline] echo
Result: Fragment found: 827
NO fragment 827 found.
```
While I definitely had a fragment but not named after my issue, name after my PR.

image

**Comment by ddimatos** [comment]
[concurrent.executor.pptx](https://github.com/user-attachments/files/16518622/concurrent.executor.pptx)

**Comment by ddimatos** [comment]
FYI, the GH actions are using the version of AC that is being checked in , I had change the she-bang to force `bash` that is how I know because I use a couple of redirects `<<<` that are specific to `bash` and `zsh` not the default shel used in gh actions, while not very portable, I will look for an alternative but what I am getting at is the gh actions were a good additional test I was not expecting and now it passes just fine.

**Review by richp405 — APPROVED** [review]
This looks like a great piece of code.  Because of the size, I'll try and run it a few times on my system to vet it more thoroughly.

**Review by richp405 — APPROVED** [review]
Looks like review was re-requested, but no files changed since I reviewed it before.

**Comment by richp405** [comment]
> Generally this is isolated code but because I introduced a new pytest fixture --zinventory-raw`and altered the behavior or the original`--zinventory` I have been running local pipelines and in this PR I ran a test suite to ensure that the original fixture was not broken by my changes.
> 
> The only reason this pipelined failed and will continue is because the changelog fragment check's logic is not ideal, it appears to rely on the branch name and a matching fragment file which is not the pattern we follow, I have an idea how to fix this check, i will do a new PR and added a new `./ac` feature to perform this operation for the pipeline.
> 
> Below is the evidence of the check that is causing my failure:
> 
> ```
> + ls *827*
> + wc -l
> ls: cannot access '*827*': No such file or directory
> [Pipeline] echo
> Result: Fragment found: 827
> NO fragment 827 found.
> ```
> 
> While I definitely had a fragment but not named after my issue, name after my PR.
> 
> image

This is a bug based on my misunderstanding of naming.  If you put the issue number in the field in pipeline, it will pass that check.

**Comment by ddimatos** [comment]
@richp405 thanks, I did not realize we had a new field, for the changelog number, I used it and it worked, while I do have another solution to try that could remove that requirement, it won't be in this PR. 

I was trying to nudge others to review this, as I have another PR I am working that aligns the tool closer to true POSIX standards, support for RHELL and Ubuntu thus the sooner this is merged the easier my other task will be.

File: tests/helpers/ztest.py [hunk 1]
```diff
@@ -47,6 +48,96 @@ def from_yaml_file(cls, path):
...
+        environment_vars.update({'PATH': f"{zoau}/bin:{pyz}/bin:/bin:/usr/sbin:/var/bin"})
+        environment_vars.update({'PYTHONSTDINENCODING': 'cp1047'})
+
+        testvars = dict()
+        testvars.update({'host': host})
+        testvars.update({'user': user})
+        testvars.update({'python_path': f"{pyz}/bin/python3"})
+        testvars.update({'environment': environment_vars})
+
+        if(extra):
```
fernandofloresg, line 134: You might get a NameErrror here if key != `extra_args` bc `extra` is not initialized.

fernandofloresg (commented):
just one comment, not sure if is a request change

tests/helpers/ztest.py
ddimatos: Fixed, thanks.

fernandofloresg (approved):
LGTM

Issue #733, opened 2023-04-13
[Release] [1.6.0] Release tasks


ddimatos:
## Release Requirements Checklist
This checklist is a chronological ordering of the release tasks to complete for an IBM Ansible z/OS core collection. This template is for a GA release that will release to Ansible Automation Platform and Galaxy as well.

By this point the `stagging-v<version>` had been code frozen and releases tasks must only be meta data related, for example, updating galaxy.yml, runtime.yml, generating module doc if needed, release notes etc.

You should be checking off each task as it completes.

## General workflow
This a subsequent release (not a patch) thus these releases tasks will eventually be merged into`main` then back into dev to allow for the meta data updates to propagate as well as the release tag. 
TBD 

### (1) Technical Writer Tasks
- [x] Ensure all technical writer Git issues are resolved for this release

### (2) Pull Request Tasks for `release-tasks-v<version>` branch:
- [x] Update `galaxy.yml`
- [x] Update `meta/runtime.yml`
- [x] Update `meta/ibm_zos_core_meta.yml`
- [x] Update `README.md`
- [x]  Update `.ansible-lint` config , keep it in synch with the `galaxy.yml` `build_ignore` entries. 
- [x] Review copyrights are updated as needed, e.g. current year if changes are made
- [x] Generate module docs (restructured text - RST) and check it into the release branch
- [x] Add a change log summary fragment in the `/changelog` directory for this release:
       - Create a file called "v\<major\>.\<minor\>.\<patch\>_summary.yml", eg `v1.6.0_summary.yml`
       - Place it under `ibm_zos_core/changelogs/fragments/`
       - Add contents (update the date, don't go past the current day):
```
release_summary: |
  Release Date: '2023-006-23'
  This changelog describes all changes made to the modules and plugins included
  in this collection. The release date is the date the changelog is created.
  For additional details such as required dependencies and availability review
  the collections `release notes <https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/release_notes.html>`__
```
- [x] Generate CHANGELOG.rst
    - `antsibull-changelog lint`
    - `antsibull-changelog release` or `antsibull-changelog release -v`
- [x] Create release notes (`docs/source/release_notes.rst`)
- [x] Review all module source imports; ensure sure no new imports are subject to license discrepancies 

### (3) Quality assurance
- [x] Run a Mend scan and check the results into the designated repository
[...]

PR #829, opened 2023-06-24
Release tasks v1.6.0

ddimatos:
##### SUMMARY
This PR does not resolve issue #733 but this PR helps bring #733 closer to completion. 

##### ISSUE TYPE
- Feature Pull Request

Version 1.6.0
=============

New Modules
-----------

- ``zos_volume_init`` - Can initialize volumes or minidisks on target z/OS systems which includes creating a volume label and an entry into the volume table of contents (VTOC).

Minor Changes
-------------

- ``zos_blockinfile`` - Adds an enhancement to allow double quotes within a block.
- ``zos_copy``

      - Updates the behavior of the `mode` option so that permissions are applied to existing directories and contents.
      - Adds an enhancement to option `restore_backup` to track modified members in a data set in the event of an error, restoring them to their previous state without reallocating the data set.
- ``zos_data_set`` - Adds a new option named *force* to enable deletion of a data member in a PDSE that is simultaneously in use by others.
- ``zos_job_query`` - Enables embedded positional wild card placement throughout *job_name* and *job_id* parameters.
- ``zos_lineinfile`` - Adds a new option named *force* to enable modification of a data member in a data set that is simultaneously in use by others.
- ``zos_tso_command`` - Adds a new option named *max_rc* to enable non-zero return codes lower than the specified maximum return as succeeded.
- ``module_utils``

      - job - Adds support for positional wild card placement for `job_name`` and `job_id`.
      - Adds support for import *common.text.converters* over the deprecated *_text* import.

Bugfixes
--------

- ``zos_copy``

      - Fixes a bug where files not encoded in IBM-1047 would trigger an error while computing the record length for a new destination dataset.
      - Fixes a bug where the module would change the mode for a directory when copying in the contents of another directory.
      - Fixes a bug where the incorrect encoding would be used during normalization, particularly when processing newlines in files.
      - Fixes a bug where binary files were not excluded when normalizing data to remove newlines.
      - Fixes a bug where a *_play_context.verbosity* deprecation warning would appear.
- ``zos_fetch`` - Fixes a bug where a *_play_context.verbosity* deprecation warning would appear.
- ``zos_encode`` - Fixes a bug where converted files were not tagged with the new code set afterwards.
- ``zos_find`` - Fixes a bug where the module would stop searching and exit after the first value in a list was not found.
- ``zos_lineinfile``

      - Removes use of Python f-string to ensure support for Python 2.7 on the controller.
      - Fixes a bug where an incorrect error message would be raised when a USS source was not found.
- ``module_utils``

      - data_set - Fixes an failure caused by cataloging a VSAM data set when the data set is not cataloged.
- ``zos_data_set`` - Fixes a bug that will leave VSAM data set cluster components behind when instructed to delete the data set (`present=absent`).
- ``zos_gather_facts`` - Fixes a bug that prevented the module from executing with newer versions of ZOAU.

Availability
------------

* `Automation Hub`_
* `Galaxy`_
* `GitHub`_

Reference
---------

* Supported by `z/OS V2R3`_ or later
* Supported by the `z/OS® shell`_
* Supported by `IBM Open Enterprise SDK for Python`_ `3.9`_ - `3.11`_
* Supported by IBM `Z Open Automation Utilities 1.2.2`_ (or later) but prior to version 1.3.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -25,9 +25,7 @@ exclude_paths:
   - tests/helpers
   - tests/requirements.txt
   - tests/unit
-  - tests/sanity/ignore-2.9.txt
   - tests/sanity/ignore-2.10.txt
-  - tests/sanity/ignore-2.11.txt
   - venv*
 parseable: true
 quiet: false
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,13 +5,13 @@ ibm.ibm_zos_core Release Notes
 .. contents:: Topics
 
 
-v1.6.0-beta.1
-=============
+v1.6.0
+======
 
 Release Summary
 ---------------
 
-Release Date: '2023-04-26'
+Release Date: '2023-06-23'
 This changelog describes all changes made to the modules and plugins included
 in this collection. The release date is the date the changelog is created.
 For additional details such as required dependencies and availability review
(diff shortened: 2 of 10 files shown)


AndreMarcel99 (approved):
All looks clean.

fernandofloresg (approved):
Looks ok to me

ddimatos:
I did reply to your comment @richp405 but also what I did was keep galaxy.yml and .ansible-lint in synch, meaning if we are not including a ignore file in the build (collection) , then we should not be linting it either. Here i added back 2.9 and 2.11 into galaxy.yml thus I added back to ansible-lint as well, hope that makes sense, i just want to keep them in synch with what they are being instructed to skip.

PR #895, opened 2023-07-17
Functional tso command test cases currently can not be run concurrently#880

### LINKED ISSUE #880: [Enabler] [zos_tso_command] Functional test cases currently can not be run concurrently

### Ansible module

_No response_

### Enabler description

The zos_tso_command test cases require sequential execution, thus far these are the ones I have found. These need to be re-written so that they can operate individually without a dependency on one another. I noticed this when I tried to run the load balancer and tests cases failed.

AndreMarcel99:
##### SUMMARY
The change substitute a set of test that depends of a sequential run and summarized in one large test with some of the options tested before and remove hardcoded verifications.

Fixes #880 

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Take logic of 8 tests and validation of each one, that required a sequential run  to just one test in the right sequence.
Captura de pantalla 2023-07-17 a la(s) 12 50 07

`test_zos_tso_command_allocate_listing_delete`
Captura de pantalla 2023-07-17 a la(s) 12 50 17

### DIFF
--- a/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
+++ b/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
@@ -1,4 +1,6 @@
 trivial:
-- zos_tso_command - Test cases were running in a specific sequence to validated.
-  Change summary the sequence into one test case.
+- zos_tso_command - Test suite was set up to run sequentially such that
+  certain tests relied on prior test cases. The new changes combine those
+  inter-dependent test cases into a single test case so that each individual
+  test case can now be run stand-alone.
   (https://github.com/ansible-collections/ibm_zos_core/pull/895)
--- a/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
+++ b/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
@@ -1,4 +1,4 @@
-bugfix:
+trivial:
 - zos_tso_command - Test cases were running in a specific sequence to validated.
   Change summary the sequence into one test case.
   (https://github.com/ansible-collections/ibm_zos_core/pull/895)
--- a/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
+++ b/changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
@@ -1,0 +1,4 @@
+bugfix:
+- zos_tso_command - Test cases were running in a specific sequence to validated.
+  Change summary the sequence into one test case.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/895)
--- a/tests/functional/modules/test_zos_tso_command_func.py
+++ b/tests/functional/modules/test_zos_tso_command_func.py
@@ -15,14 +15,9 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-import os
-import sys
-import warnings
-
 import ansible.constants
 import ansible.errors
 import ansible.utils
-import pytest
 
 DEFAULT_TEMP_DATASET="imstestl.ims1.temp.ps"
 
--- a/tests/functional/modules/test_zos_tso_command_func.py
+++ b/tests/functional/modules/test_zos_tso_command_func.py
@@ -58,7 +58,7 @@ def test_zos_tso_command_allocate_listing_delete(ansible_zos_module):
         for item in result.get("output"):
             assert item.get("rc") == 0
         assert result.get("changed") is True
-    # Validate listds of datasets and validate LISTDS using alias param 'command'
+    # Validate listds of datasets and validate LISTDS using alias param 'command' of auth command
     results = hosts.all.zos_tso_command(commands=["LISTDS '{0}'".format(DEFAULT_TEMP_DATASET)])
     for result in results.contacted.values():
         for item in result.get("output"):
@@ -70,7 +70,7 @@ def test_zos_tso_command_allocate_listing_delete(ansible_zos_module):
         for item in result.get("output"):
             assert item.get("rc") == 0
         assert result.get("changed") is True
-    # Validate LISTCAT comand
+    # Validate LISTCAT comand and an unauth command
     results = hosts.all.zos_tso_command(
         commands=["LISTCAT ENT('{0}')".format(DEFAULT_TEMP_DATASET)]
     )
--- a/tests/functional/modules/test_zos_tso_command_func.py
+++ b/tests/functional/modules/test_zos_tso_command_func.py
@@ -63,13 +63,13 @@ def test_zos_tso_command_allocate_listing_delete(ansible_zos_module):
         for item in result.get("output"):
             assert item.get("rc") == 0
         assert result.get("changed") is True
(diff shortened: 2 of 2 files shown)


richp405 (approved):
Pretty slick merger of tests!

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -146,7 +114,8 @@ def test_zos_tso_command_invalid_command(ansible_zos_module):
     hosts = ansible_zos_module
     results = hosts.all.zos_tso_command(commands=["xxxxxx"])
     for result in results.contacted.values():
-        assert result.get("output")[0].get("rc") == 255
+        for item in result.get("output"):
+            assert item.get("rc") == 255
         assert result.get("changed") is False
```
ketankelkar, line 121: I know you didn't touch this next test `test_zos_tso_command_multiple_commands` but I was wondering if it's possible to verify that 2 commands were run. Currently it looks like the test is only checking that the `rc` is `0`. I'd like to see a check for the correct number of `rc`s coming back (from [docs](https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/modules/zos_tso_command.html#return-values), it appears that for each command there should be a return structure).

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -45,89 +47,53 @@ def test_zos_tso_command_long_command_128_chars(ansible_zos_module):
...
-# the dataset has be deleted.
-def test_zos_tso_command_valid_command_failed_as_has_been_deleted(ansible_zos_module):
-    hosts = ansible_zos_module
-    results = hosts.all.zos_tso_command(commands=["delete 'imstestl.ims1.temp.ps'"])
+    # Validate already was remove
+    results = hosts.all.zos_tso_command(commands=["delete '{0}'".format(DEFAULT_TEMP_DATASET)])
     for result in results.contacted.values():
-        assert result.get("output")[0].get("rc") == 8
+        for item in result.get("output"):
+            assert item.get("rc") == 8
```
ketankelkar, line 96: A little nit-picking here, I think it would be beneficial to add a comment explaining why the expected rc is 8. Maybe something like:

"# Expect the tso_command to fail here because the previous command will have already deleted the data set"

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -45,89 +47,53 @@ def test_zos_tso_command_long_command_128_chars(ansible_zos_module):
...
+            assert item.get("rc") == 0
         assert result.get("changed") is True
-
-
-# The failure path test
-# the dataset has be deleted.
-def test_zos_tso_command_valid_command_failed_as_has_been_deleted(ansible_zos_module):
-    hosts = ansible_zos_module
-    results = hosts.all.zos_tso_command(commands=["delete 'imstestl.ims1.temp.ps'"])
+    # Validate already was remove
```
ketankelkar, line 92: consider rewording this comment to:
"Validate data set was removed by previous call"

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -45,89 +47,53 @@ def test_zos_tso_command_long_command_128_chars(ansible_zos_module):
...
+            assert item.get("rc") == 0
         assert result.get("changed") is True
-
-
-# The positive path test
-# tests that single command works as well with alias
-def test_zos_tso_command_auth_command_listds_using_alias(ansible_zos_module):
-    hosts = ansible_zos_module
-    results = hosts.all.zos_tso_command(command=["LISTDS 'imstestl.ims1.temp.ps'"])
+    # Validate listds auth of a dataset with commands
```
ketankelkar, line 66: this is outside my scope of knowledge, but I'm not seeing where 'auth' and not-'auth' is invoked in the test case. To be fair, I don't see where/how it was invoked in the tests prior to your changes either. I think this is a good opportunity to add testing for that functionality too

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -45,89 +47,53 @@ def test_zos_tso_command_long_command_128_chars(ansible_zos_module):
...
+            assert item.get("rc") == 0
         assert result.get("changed") is True
-
-
-# The positive path test
-# tests that alias "command" works
-def test_zos_tso_single_command_auth_command_listds_using_alias(ansible_zos_module):
-    hosts = ansible_zos_module
-    results = hosts.all.zos_tso_command(command="LISTDS 'imstestl.ims1.temp.ps'")
+    # Validate listds auth single command
```
ketankelkar, line 72: 'single command' is a bit unclear. Maybe reword this to:
"Validate LISTDS using alias param 'command'"

changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
```diff
@@ -0,0 +1,4 @@
+bugfix:
```
ketankelkar, line 1: i'd recategorize this to 'trivial' which will skip including this fragment in the changelog since these changes are to the tests and not to the module/module_utils, from a user pespective, there's no change.

ketankelkar (changes requested):
Overall changes look great! Just a few minor edits to comments and a couple of requests to expand test coverage. It might be more appropriate to record the requests as a separate issue and implement them later instead of as part of this PR depending on the level of effort (i'm not sure).

changelogs/fragments/880-Functional_tso_command_test_cases_currently_can_not_be_run_concurrently.yml
```diff
@@ -0,0 +1,4 @@
+trivial:
+- zos_tso_command - Test cases were running in a specific sequence to validated.
+  Change summary the sequence into one test case.
```
ketankelkar, line 3: Perhaps reword the message to something like:
- zos_tso_command - Test suite was set up to run sequentially such that certain tests relied on prior test cases. The new changes combine those inter-dependent test cases into a single test case so that each individual test case can now be run stand-alone.

tests/functional/modules/test_zos_tso_command_func.py
AndreMarcel99: I already adapt the comment above to say what command is auth and which not

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -45,89 +42,54 @@ def test_zos_tso_command_long_command_128_chars(ansible_zos_module):
...
-        assert result.get("output")[0].get("rc") == 0
+        for item in result.get("output"):
+            assert item.get("rc") == 0
         assert result.get("changed") is True
-
-
-# The positive path test
-def test_zos_tso_command_unauth_command_listcat(ansible_zos_module):
-    hosts = ansible_zos_module
+    # Validate LISTCAT comand and an unauth command
```
ketankelkar, line 73: missing an 'm' in 'command'

ketankelkar (changes requested):
thanks for making the requested changes, glad we got the auth stuff cleared up offline, and also thanks for adding the extra check for the two-command test. Just suggesting two very tiny edits and this should be good to go. Remember to re-run the pipeline and paste the results before merging

ketankelkar:
clean pipeline run 
image

ketankelkar (approved):
lgtm!

PR #915, opened 2023-07-26
Staging v1.7.0 beta.1

ddimatos:
##### SUMMARY
Release 1.7.0-beta.1 tasks, addresses issue #879 
This PR includes all 1.7.0-beta.1 content ready to merge into main and later be tagged and released. See issue #879 for more details.


ddimatos:
I noticed as I was generating doc that the changed value coming back from zos_lineinfile is changed:<int>, yet doc describes it a boolean, that is because its supposed to be a boolean but zoau is sending it back as an int, so ill create an issue to correct that but don't want to change the code right now on a release boundary, so I will update the doc temporarily.

{"cmd": "dsedhelper -d -c IBM-1047 /D160882/d ANSIBLE.LISTCAT.INPUT ", "found": 3, "changed": 1}

image

ddimatos:
For pipeline regression, a single pipeline was not used, instead 28 individual pipelines were used, one for each test suite, this content was collected and added to the release folder, there is no image to share here, please review the release folder.

richp405 (approved):
Only issue, which I don't think matters, is in docs/scripts/*.sh, the copyright dates are in 2020, not 2023.  Looks like a great release overall!

ddimatos:
> Only issue, which I don't think matters, is in docs/scripts/*.sh, the copyright dates are in 2020, not 2023. Looks like a great release overall!

Thanks @richp405 - I updated them

Issue #695, opened 2023-03-27
[Enabler] [pipeline] Add ansible-lint to the pipeline


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

Ansible-lint is a requirement in June 2023, we need this in place to ensure we are able to release.

ddimatos:
./ac has `ac --ac-lint` , I think the only thing needed is to ensure a RC is returned to STDOUT from ac and remaining graphical output be sent to stderr (not ideal) but verbosity need not be considered here which is what we would send to stderr. eg
```
# --------------
# Running ....
# --------------
``` 

^^^ can all go to stderr

ddimatos:
You might need to have a build either from the pipeline or use `./ac --ac-build`

ddimatos:
@richp405 please shadow @AndreMarcel99 on this item, have him understand our pipeline code a bit and walk him through a change. This should be pretty easy since the config is in the root project now, remember, not all releases will have the config so you may need to check first that `.ansible-lint` is present first. If we end up having to release a fix to an older version we can add the `.ansbile-lint` then at that time. 

Also, we have a work item for Q3.4 to address the output (below) , we are allowed to have this till Q1 2024, so you will want to ignore anything that looks like this:
```
meta/runtime.yml:1: meta-runtime[unsupported-version]: requires_ansible key must be set to a supported version.
tests/sanity/ignore-2.11.txt:16: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-choices-do-not-match-spec at line 16, which is not a permitted ignore.
tests/sanity/ignore-2.11.txt:23: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-default-does-not-match-spec at line 23, which is not a permitted ignore.
tests/sanity/ignore-2.11.txt:24: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-type-does-not-match-spec at line 24, which is not a permitted ignore.
tests/sanity/ignore-2.11.txt:26: sanity[cannot-ignore]: Ignore file contains validate-modules:parameter-type-not-in-doc at line 26, which is not a permitted ignore.
tests/sanity/ignore-2.11.txt:27: sanity[cannot-ignore]: Ignore file contains validate-modules:undocumented-parameter at line 27, which is not a permitted ignore.
[...]

PR #813, opened 2023-06-15
Add ansible-lint tooling added

ddimatos:
##### SUMMARY
This addresses issue #809 partially. There is still the remaining CI/CD work to be done that is not managed in this repository. We can continue to track the CI/CD work for ansible-lint using this git issue #809 but can consider this issue complete for now. 

##### ISSUE TYPE
- Feature Pull Request

PR #932, opened 2023-08-15
Testlint

richp405:
##### SUMMARY
This is the 2 changes needed for ansible-lint to work on the dev branch

This is the end of ticket 695: 
https://github.com/ansible-collections/ibm_zos_core/issues/695

##### ISSUE TYPE
- Feature Pull Request: Add ansible-lint scanning to pipeline

##### COMPONENT NAME
Pipeline: allowing Ansible-lint to run without tripping over the sanity/ignore files.
Also updates requirements to 2.11, since 2.9 is now out-of-support.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -26,6 +26,7 @@ exclude_paths:
   - tests/requirements.txt
   - tests/unit
   - tests/sanity/ignore-2.10.txt
+  - tests/sanity/ignore*.txt
   - venv*
 parseable: true
 quiet: false
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,2 +1,2 @@
 ---
-requires_ansible: '>=2.9.0'
+requires_ansible: '>=2.11.0'


rexemin:
Changes look straightforward, just one question: in the issue Demetri mentioned that at some point we'll have to work on fixing the ignores because the requirements for collections will change in 2024, but I couldn't find it, do you know what issue will that be? Or am I understanding that part wrong?

richp405:
> Changes look straightforward, just one question: in the issue Demetri mentioned that at some point we'll have to work on fixing the ignores because the requirements for collections will change in 2024, but I couldn't find it, do you know what issue will that be? Or am I understanding that part wrong?

@rexemin I don't remember that conversation, but he would be right.  I'm "hiding" the error so we can check.  The error (there are at least 100) is that things in the sanity/ignore files are not allowed to be ignored.  I'll look in some threads to see if there are any other approaches.

richp405:
> > Changes look straightforward, just one question: in the issue Demetri mentioned that at some point we'll have to work on fixing the ignores because the requirements for collections will change in 2024, but I couldn't find it, do you know what issue will that be? Or am I understanding that part wrong?
> 
> @rexemin I don't remember that conversation, but he would be right. I'm "hiding" the error so we can check. The error (there are at least 100) is that things in the sanity/ignore files are not allowed to be ignored. I'll look in some threads to see if there are any other approaches.

Interesting... I see comments in a 'merged' doc: https://github.com/ansible-collections/ibm_zos_core/pull/813/files. My concern is the entry is listed as merged, but that's NOT what was in /dev/ yesterday... it shows only 2.9 ignore and a dependency on 2.9...the 813 issue showed ignore on 2.9, 2.10 and 2.11, and a dependency on 2.12.  Not sure where to go with this.

richp405:
Since we have the actual fix work assigned for Q4, I'll close this PR.

PR #934, opened 2023-08-16
Remove conditional unnecessary

AndreMarcel99:
##### SUMMARY
Fix conditionals for next issues in runs of pipeline

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
The test is the one about backups and to ensure removes the dataset at the end, it was required the assignation of a dataset name with or without a member, one of the options of backup is None, the conditional it was for case None to ensures the delete of dataset but generate and error because variable not assigned of case None.

### DIFF
--- a/changelogs/fragments/934-Remove-conditional-unnecessary.yml
+++ b/changelogs/fragments/934-Remove-conditional-unnecessary.yml
@@ -1,0 +1,2 @@
+- trivial:
+  - zos_blockinfile - remove test conditional unnecessary (https://github.com/ansible-collections/ibm_zos_core/pull/934).
--- a/tests/functional/modules/test_zos_blockinfile_func.py
+++ b/tests/functional/modules/test_zos_blockinfile_func.py
@@ -1221,9 +1221,6 @@ def test_ds_block_insertafter_eof_with_backup(ansible_zos_module, dstype, backup
             if backup_name:
                 backup_ds_name = result.get("backup_name")
                 assert backup_ds_name is not None
-            else:
-                backup_ds_name = result.get("backup_name")
-                assert backup_ds_name is not None
         results = hosts.all.shell(cmd="cat \"//'{0}'\" ".format(params["path"]))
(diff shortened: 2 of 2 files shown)


rexemin:
Is there a corresponding issue for this PR? Or is this something you did when working through some other item?

AndreMarcel99:
When I was cleaning the blockinfile I put this bad code

richp405 (approved):
Pretty straightrforward change.  Looks good from here.

fernandofloresg:
Why is this failing? 
What I understand from this code is that the module params include backup=True. If I recall correctly, is backup=True and backup name is None the module will create a temporary data set and use it as backup, hence, it should return a backup name.


If this code was to be removed then how will we assert the fact that when user sets backup=True the module will return the backup name of the automatically created data set? 

If this test is failing, wouldn't it be because our code is not providing the expected data set name and hence, our code has a bug?

AndreMarcel99:
You're right @fernandofloresg and just need to remove the conditional, I forgot that create the temporary dataset required to be remove by space.
Thanks.
Captura de pantalla 2023-08-17 a la(s) 12 03 01

rexemin (approved):
Looks good, thanks for addressing @fernandofloresg's comment. I figure there isn't really a necessity to run the pipeline for this change.

fernandofloresg:
Just before I perform the merge. What happened here was to remove an unnecessary if/else block. Just to be sure, previous version was not failing right?  Is this just a code enhancement?

fernandofloresg:
I guess if it failed one time in many, was it because of environment issues?

AndreMarcel99:
My guess is yes because it was trying to access to backup_name from the result and it was the option NONE as backup name

fernandofloresg:
I guess not, if it failed once after a lot of attempts maybe there was some instability and didn't return the proper backup name. Anyways this change simplifies the test case and will catch any issue with backup_name.

Issue #903, opened 2023-07-19
[Bug] zos_job_submit action plugin leaves AnsiballZ_tempfile.py in ansible temp dir


stewartfrancis:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Are the dependencies a supported version?

- [X] The dependencies are supported.

### IBM Z Open Automation Utilities

v1.2.3

### IBM Enterprise Python

v3.10.x

### IBM z/OS Ansible core Version

v1.6.0

### ansible-version

v2.14.x

### z/OS version

v2.5

### Ansible module

zos_job_submit

### Bug description
Update: See discussion `slack.com/archives/CSLEJ8VGV/p1689608848640829`

Submitting a `LOCAL` JCL file from the controller to a managed z/OS node leaves an `AnsiballZ_tempfile.py` temporary file in a temporary subdirectory in `~/.ansible/tmp`

This is as a result of the `zos_job_submit` action module invoking the `ansible.builtin.tempfile` module.  This does not happen when executing the `tempfile` module natively on z/OS.  This appears to be caused by the temporary directory used by the `tempfile` module being different from the temporary directories used by other modules executed by the `zos_job_submit` action plugin.  That means that all the other `AnsiballZ_*.py` files are cleaned up by the `rm` command seen in the verbose log, but the `tempfile` `AnsiballZ_tempfile.py` file is left behind in its directory which is never deleted.

This is a pretty serious problem for our operator collection as we're using `zos_job_submit` in an environment with a user directory with a very restricted amount of space, so these small temporary files are eventually causing us to run out of space when we end up with hundreds of them remaining.

Sample playbook:
```yaml
---
- name: Test
  hosts: all
  gather_facts: false

  tasks:
    - name: Asdf
      environment: "{{ z_environment_vars }}"
      ibm.ibm_zos_core.zos_job_submit:
        src: test.jcl
        location: LOCAL
        wait: true
        max_rc: 0
      register: job_submit_response
```

Sample JCL (can be anything):
```
//ZTESTDS JOB REGION=0M,MSGLEVEL=(1,1),MSGCLASS=R
//DEFTRACE EXEC PGM=IEFBR14,MEMLIMIT=0M
//DD1      DD DISP=(NEW,DELETE,DELETE),
//         SPACE=(CYL,(25)),UNIT=SYSDA,
//         DCB=(RECFM=VB,LRECL=4092,BLKSIZE=4096),
//         DSN=IBMUSER.TEST
[...]

PR #951, opened 2023-08-23
Add fix for change copy built-in for zos_copy and remove remain files…

### LINKED ISSUE #926: [Enhancement] [zos_job_submit] Update the zos_job_submit plugin to use zos_copy

### Ansible module

zos_job_submit

When zos_job_submit was written, zos_copy had not been developed and the zos_job_submit then relied on a few community modules such as file and and copy. While this is not an issue it could be given the recent work and exploration of how ansible is becoming more strict with UTF8 responses and IBM Enterprise changing how it treats files opened as binary, its time this action plugin be reconsidered for an update. 

Currently the action plugin manages creating a tmp file on the managed node to copy JCL into from the controller to the managed node which actually has an issue open #903 , I propose we use zos_copy to perform these operations which are part of its core function, zos_copy can create new files on the managed node as well as transfer JCL from the controller to the managed node. You should consider file tags and permissions as well as encoding when updating the plugin.

AndreMarcel99:
##### SUMMARY
Change the action plugin of zos_job_submit module that was calling Ansible build int copy now uses zos_copy. Change the remove function to ensure Ansible/temp folder goes empty by files created by action plugin.

Fixes #926 and #903 

##### ISSUE TYPE
- Bugfix Pull Request
- Enabler Pull Request

##### COMPONENT NAME
Change the call of plugin of Ansible build in copy to zos_copy module, adjust the code to ensure works properly and add validation to remove temporal files that module use.

##### ADDITIONAL INFORMATION
Import action module of zos_copy an prepare the change on the environment to works properly and get the right files with the jcl, lastly add validations to ensure clean temporary files creates to works with local files.

Captura de pantalla 2023-08-29 a la(s) 14 09 21
Captura de pantalla 2023-08-29 a la(s) 14 09 54

### DIFF
--- a/changelogs/fragments/951-Change-copy-for-zos-copy-and-remove-temporary-files.yml
+++ b/changelogs/fragments/951-Change-copy-for-zos-copy-and-remove-temporary-files.yml
@@ -1,0 +1,7 @@
+bugfixes:
+  - zos_job_submit: Temporary files were created in tmp directory.
+      Fix now ensure the deletion of files every time the module run.
+        (https://github.com/ansible-collections/ibm_zos_core/pull/951)
+minor_changes:
+  - zos_job_submit: Change action plugin call from copy to zos_copy.
+      (https://github.com/ansible-collections/ibm_zos_core/pull/951)
--- a/changelogs/fragments/951-Change-copy-for-zos-copy-and-remove-temporary-files.yml
+++ b/changelogs/fragments/951-Change-copy-for-zos-copy-and-remove-temporary-files.yml
@@ -1,6 +1,6 @@
 bugfixes:
   - zos_job_submit: Temporary files were created in tmp directory.
-      Fix now ensure the deletion of files every time the module run.
+      Fix now ensures the deletion of files every time the module run.
         (https://github.com/ansible-collections/ibm_zos_core/pull/951)
 minor_changes:
   - zos_job_submit: Change action plugin call from copy to zos_copy.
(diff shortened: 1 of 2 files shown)


richp405 (approved):
looks okay from here

changelogs/fragments/951-Change-copy-for-zos-copy-and-remove-temporary-files.yml
```diff
@@ -0,0 +1,7 @@
+bugfixes:
+  - zos_job_submit: Temporary files were created in tmp directory.
+      Fix now ensure the deletion of files every time the module run.
```
fernandofloresg, line 3: change "ensure" to "ensures"

plugins/action/zos_job_submit.py
```diff
@@ -148,16 +154,20 @@ def run(self, tmp=None, task_vars=None):
...
+            copy_task.args = copy_module_args
+            zos_copy_action_module = ZosCopyActionModule(task=copy_task,
+                                                         connection=self._connection,
+                                                         play_context=self._play_context,
+                                                         loader=self._loader,
+                                                         templar=self._templar,
+                                                         shared_loader_obj=self._shared_loader_obj)
+            result.update(zos_copy_action_module.run(task_vars=task_vars))
+            module_args["src"] = dest_path
             result.update(
```
fernandofloresg, line 171: Please add a validation when zos_copy action fails, this will avoid some headaches when debugging. Take a look into zos_unarchive, the code checks result.get('msg') I believe, is enough just adding a message that copying to remote failed.

fernandofloresg (changes requested):
All ok, no new test cases are needed since this changes an already implemented functionality behavior under the hood.

AndreMarcel99:
Captura de pantalla 2023-09-04 a la(s) 12 22 31
Captura de pantalla 2023-09-04 a la(s) 12 22 47

fernandofloresg (approved):
looks good now

PR #966, opened 2023-09-11
Add python 3.11-3 to ac mount tables

fernandofloresg:
##### SUMMARY

Trivial pull request, add python 3.11-3 into mounts.env.

No regression test is needed

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

ac tool


AndreMarcel99 (approved):

ketankelkar (approved):
LGTM, one consideration is to always name the _latest_ python patch releases 3.xx and then the older patches/GAs as 3.xx-yy (eg 3.11-ga) so that `allpython/3.9`, `allpython/3.10`, and `allpython/3.11` always point to the latest, while the older versions are just one configuration step away

fernandofloresg:
> 

Might be, but this is dependent on how we configure our pipelines, this current config follows their convention

PR #968, opened 2023-09-14
Initial commit to add LIBRARY to choices and docs

### LINKED ISSUE #920: [Bug] [zos_copy] Missing 'library' for dest_data_set sub option type.

Module zos_copy is missing the type library that enables and supports an executable. Although a PDS/E will suffice for an executable, this sets the DSNTYPE , thus its simply a doc issue, the code can already support it. 

Refer to zos_data_set module type to additional examples. 

The zos_copy documentation below that needs updating:
```
  dest_data_set:
    description:
      - Data set attributes to customize a C(dest) data set to be copied into.
    required: false
    type: dict
    suboptions:
      type:
        description:
          - Organization of the destination
        type: str
        required: true
        choices:
          - KSDS
          - ESDS
          - RRDS
          - LDS
          - SEQ
          - PDS
          - PDSE
          - MEMBER
          - BASIC
 ```

You will also need to update the choices module spec , eg
```
            dest_data_set=dict(
                type='dict',
                required=False,
                options=dict(
                    type=dict(
                        type='str',
                        choices=['BASIC', 'KSDS', 'ESDS', 'RRDS',
                                 'LDS', 'SEQ', 'PDS', 'PDSE', 'MEMBER'],
                        required=True,
                    ),
```

v1.7.0-beta.1

### Ansible module

zos_copy

### PR DESCRIPTION
**Pull request by richp405**
##### SUMMARY
Change to zos_copy:
Added LIBRARY type to dest_dataset/suboptions as a valid choice
Updated in file documentation to reflect the option

Fixes #920 

##### ISSUE TYPE
- Bugfix Pull Request (missing option)
- Docs Pull Request (added option)

##### COMPONENT NAME
zos_copy.py

### DIFF
--- a/changelogs/fragments/920-bug-add-library-feature-documentation.yml
+++ b/changelogs/fragments/920-bug-add-library-feature-documentation.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - zos_copy: Add 'LIBRARY' option as a destination dataset suboption.
+      Update documentation to show this option.
+        (https://github.com/ansible-collections/ibm_zos_core/pull/968)
--- a/changelogs/fragments/920-bug-add-library-feature-documentation.yml
+++ b/changelogs/fragments/920-bug-add-library-feature-documentation.yml
@@ -1,4 +1,0 @@
-bugfixes:
-  - zos_copy: Add 'LIBRARY' option as a destination dataset suboption.
-      Update documentation to show this option.
-        (https://github.com/ansible-collections/ibm_zos_core/pull/968)
--- a/changelogs/fragments/920-zos-copy-add-library-choice.yml
+++ b/changelogs/fragments/920-zos-copy-add-library-choice.yml
@@ -1,0 +1,4 @@
+bugfixes:
+- zos_copy - Update option limit to include LIBRARY as dest_dataset/suboption value.
+    Documentation updated to reflect this change.
+      (https://github.com/ansible-collections/ibm_zos_core/pull/).
--- a/changelogs/fragments/920-zos-copy-add-library-choice.yml
+++ b/changelogs/fragments/920-zos-copy-add-library-choice.yml
@@ -1,4 +1,4 @@
 bugfixes:
 - zos_copy - Update option limit to include LIBRARY as dest_dataset/suboption value.
     Documentation updated to reflect this change.
-      (https://github.com/ansible-collections/ibm_zos_core/pull/).
+      (https://github.com/ansible-collections/ibm_zos_core/pull/968).
--- a/docs/source/modules/zos_copy.rst
+++ b/docs/source/modules/zos_copy.rst
@@ -74,7 +74,7 @@ dest
 
   If ``dest`` is a nonexistent data set, it will be created following the process outlined here and in the ``volume`` option.
 
-  If ``dest`` is a nonexistent data set, the attributes assigned will depend on the type of ``src``. If ``src`` is a USS file, ``dest`` will have a Fixed Block (FB) record format and the remaining attributes will be computed. If ``src`` is binary, ``dest`` will have a Fixed Block (FB) record format with a record length of 80, block size of 32760, and the remaining attributes will be computed.
+  If ``dest`` is a nonexistent data set, the attributes assigned will depend on the type of ``src``. If ``src`` is a USS file, ``dest`` will have a Fixed Block (FB) record format and the remaining attributes will be computed. If *is_binary=true*, ``dest`` will have a Fixed Block (FB) record format with a record length of 80, block size of 32760, and the remaining attributes will be computed. If *executable=true*,``dest`` will have an Undefined (U) record format with a record length of 0, block size of 32760, and the remaining attributes will be computed.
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Review by ketankelkar — COMMENTED** [review]
LGTM - curious to know what kind of validation can be done (both docs-wise and functionally). I don't anticipate any hiccups

File: changelogs/fragments/920-bug-add-library-feature-documentation.yml
Comment (fernandofloresg) [inline changelogs/fragments/920-bug-add-library-feature-documentation.yml:1]: There are two changelog fragments for this PR, please keep only one

**Review by fernandofloresg — CHANGES_REQUESTED** [review]
Seems really straightforward, can we add a test for it ?

File: changelogs/fragments/920-bug-add-library-feature-documentation.yml
Comment (richp405) [reply changelogs/fragments/920-bug-add-library-feature-documentation.yml:1]: oops!  I've removed the bug version

**Comment by richp405** [comment]
> Seems really straightforward, can we add a test for it ?

@fernandofloresg I'm not sure.. there don't appear to be tests to the other destination subtypes, so I'm not sure if we should add another test here.

**Comment by ddimatos** [comment]
> > Seems really straightforward, can we add a test for it ?
> 
> @fernandofloresg I'm not sure.. there don't appear to be tests to the other destination subtypes, so I'm not sure if we should add another test here.

@fernandofloresg @richp405 , LIBRARY ultimately maps to PDSE in the lower layers of the code, testing LIBRARY is tested by zos_data_set since its the shared module_utility used to create these data sets, the only thing that might need validation is that our module correctly pass along the option till the shared code. 

To actually test a library,[ the easiest but fairly costly test case would be to replicate the load module test](https://github.com/ansible-collections/ibm_zos_core/blob/516df020117ad0854f788f416b7c9bd225c7d876/tests/functional/modules/test_zos_copy_func.py#L2537) and instead of using a PDSE to write the load module into, write it into a type LIBRARY. We could also have @ketankelkar see if he could add a LIBRARY test when they write the alias tests as there will be new load module tests. 

Low level code I mentioned:
```
if (!argstrnocasecmp(remainder, TYPE_PDSE) || !argstrnocasecmp(remainder, TYPE_LIBRARY)) {
		setDynParm(dynParms, Dsntype, __DSNT_LIBRARY);
		if (!dynParms[Dsorg].used) {
			setDynParm(dynParms, Dsorg, __DSORG_PO);
		}
```

**Comment by ddimatos** [comment]
I generated and pushed up the updated doc to reflect this change directly into the branch, you can see the *.rst and screen capture below:
image

**Comment by ddimatos** [comment]
> LGTM - curious to know what kind of validation can be done (both docs-wise and functionally). I don't anticipate any hiccups

@ketankelkar , i generated doc for this PR , saw that was missing, short of a manual playbook test the only other type of check/test we can do is to write a test case using LIBRARY which is call the module_utils and tested there but maybe we can tweak a load module test to use a LIBRARY data set instead of a PDSE (they are the same).

**Comment by ketankelkar** [comment]
I found this test case near the very bottom: [test_dest_data_set_parameters](https://github.com/ansible-collections/ibm_zos_core/blob/516df020117ad0854f788f416b7c9bd225c7d876/tests/functional/modules/test_zos_copy_func.py#L3401), maybe it can be augmented with other data set types too, it looks like it's currently only testing sequential data sets.

I'd be happy to add it into the loadlib test cases too! I'll pull in your changes

**Review by ddimatos — APPROVED** [review]
@richp405 This looks good to me, I commented on a few of the discussions about testing this, I think this can be tested by the alias function which is creating a number of new executable tests which are already expensive in time, don't think we need a test simply just for Library when we can test library with alias support. Thank you.

**Comment by richp405** [comment]
@fernandofloresg Based on the conversation above (DD and Ketan)... are you okay with that as a test for this?

**Comment by ketankelkar** [comment]
I updated my branch (`enhancement/423/zos_copy-add-data-set-member-alias-support`) with the latest `dev` and then only modified the two lines this PR changed up in zos_copy.py 
here's the git diff:
image

I modified the test cases with loadlibs and switched the `type: pdse` to `type: library`. 
e.g.:
image
image

And all the test cases with loadlibs passed
image

Note- i created a new pytest "mark" called loadlibs and marked all the test cases I found with loadlibs.

**Comment by ketankelkar** [comment]
my comment above was a bit misguided, i modified the `test_copy_pds_loadlib_to_pds_loadlib` test case to run with and without pre-allocated destination data sets. In the variation without pre-allocated data sets, I invoke `zos_copy` with the `dest_data_set` option and pass in 'LIBRARY' as a value.

```python
        if not is_created:
            # dest data set does not exist, specify it in dest_dataset param.
            # copy src loadlib to dest library pds w/o aliases
            copy_res = hosts.all.zos_copy(
                src="{0}".format(src_lib),
                dest="{0}".format(dest_lib),
                remote_src=True,
                executable=True,
                aliases=False,
                dest_data_set={
                    'type': "LIBRARY",
                    'record_format': "U",
                    'record_length': 0,
                    'block_size': 32760,
                    'space_primary': 2,
                    'space_type': "M",
                }
            )
            # copy src loadlib to dest library pds w aliases
            copy_res_aliases = hosts.all.zos_copy(
                src="{0}".format(src_lib),
                dest="{0}".format(dest_lib_aliases),
                remote_src=True,
                executable=True,
                aliases=True,
                dest_data_set={
                    'type': "LIBRARY",
                    'record_format': "U",
                    'record_length': 0,
                    'block_size': 32760,
                    'space_primary': 2,
                    'space_type': "M",
                }
            )
```

As the image shows, both variations of this test case pass:
image

ddimatos:
@ketankelkar thanks so much, looks like you cover the LIBRARY test in your alias branch work which just came up for review today.

fernandofloresg (approved):
is ok for me @richp405

PR #969, opened 2023-09-14
Simplify loaldlib test cases

### LINKED ISSUE #958: [Enabler] [zos_copy] Improve test case - test_copy_pds_loadlib_member_to_pds_loadlib_member

### Ansible module

zos_copy

Update, simplify and add comments/docs to the `test_copy_pds_loadlib_member_to_pds_loadlib_member` test case in zos_copy.

Something to keep in mind when simplifying/modularizing is that each test case still needs to be independent and self-contained. Any helper functions broken away from this code block cannot do any testing unless they are designed to be thread-safe (expect parallel runs of individual test cases).

Comments by @ddimatos 
>its a good first attempt but it can be easier to follow and maybe more modular so we can reuse parts of it, this way it won’t block the pr and it can be new work. you can pull some of my comments from [here](https://github.com/ansible-collections/ibm_zos_core/pull/804/files#r1282706832) maybe enter the gh issue number as a comment afterwards [here](https://github.com/ansible-collections/ibm_zos_core/pull/804/files#r1282706832)

Some of the comments mentioned above:
>I know this test case was there from before and it was expanded, I did struggle to follow all it was doing, often when i see a test case like this and I am editing it, i will add comments as I learn it so its easier for other to later maintain and reviewers to follow. I know this not part of this code change as it existed before but just how objects are named is confusing, for example member = "HELLOSRC" is not really source, it is the actual program that gets created by the compile and link done inlink_loadlib_from_cobol, "HELLOPGM" might be a better name, this made hard to follow, there is no action here just commenting on how it was hard to follow from even before you edited it.

and

>Took me some time but I finally understood why there is a second zos_copy, although it could have been its own test case given how complicated this one test is, what is being done is more or less a round trip copy, in the first zos_copy, a copy of an executable from one PDSE(member) to another PDSE(member) in the next zos_copy, a copy of PDSE(member) to USS, this is to test the round trip of the data and execution bits.

see internal slack message 
> C043LSMTQ2H/p1693309666543319?thread_ts=1693245020.939149&cid=C043LSMTQ2H

AndreMarcel99:
##### SUMMARY
Simplify test case for executable.

Fixes #958 

##### ISSUE TYPE
- Enhancement

##### COMPONENT NAME
Generate two functions that will works as helpers to ensure the creation of executables files to remove unnecessary an repetitive case for test case.

##### ADDITIONAL INFORMATION
Generate two functions that will works as helpers to ensure the creation of executables files to remove unnecessary an repetitive case for test case, also by dividing the cases by test ensuring proper understanding and calls of functions.

Captura de pantalla 2023-09-14 a la(s) 15 55 04
Captura de pantalla 2023-09-14 a la(s) 15 55 13

### DIFF
--- a/changelogs/fragments/969-Simplify_loadlib_test_cases.yml
+++ b/changelogs/fragments/969-Simplify_loadlib_test_cases.yml
@@ -1,0 +1,2 @@
+trivial:
+- zos_copy - Divide large test case for loadlibs and simplify functions.
--- a/changelogs/fragments/969-Simplify_loadlib_test_cases.yml
+++ b/changelogs/fragments/969-Simplify_loadlib_test_cases.yml
@@ -1,2 +1,3 @@
 trivial:
 - zos_copy - Divide large test case for loadlibs and simplify functions.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/969)
--- a/tests/functional/modules/test_zos_copy_func.py
+++ b/tests/functional/modules/test_zos_copy_func.py
@@ -342,7 +342,7 @@ def generate_executable_ds(hosts, src, dest, cobol):
     assert rc == 0
     exec_res = hosts.all.shell(cmd=cmd.format(member, src))
     for result in exec_res.contacted.values():
-            assert result.get("rc") == 0
+        assert result.get("rc") == 0
 
 def generate_executable_uss(hosts, src, src_jcl_call):
     hosts.all.zos_copy(content=hello_world, dest=src, force=True)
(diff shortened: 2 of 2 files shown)


ketankelkar (commented):
This LGTM, only question I have is would this be safe to run with the load balancer since some of the data sets are hard-coded already? Maybe it's worth adding a check to the helper functions to test if the executable/loadlib already exists or not and if not, then dont bother re-generating, just return

richp405 (approved):
Clean and straightforward.  Nice!

ketankelkar (approved):

PR #970, opened 2023-09-14
Switched job.py to use the version_checker module for zoau versioning

richp405:
##### SUMMARY
Modified job.py routine to use zoau_version_checker to check version of zoau

should resolve 949

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
job.py needs to know the zoau version so it can determine which variables to check for in the return object.

##### ADDITIONAL INFORMATION
It appears (still checking) that the __init__.py file is not maintained, as 1.2.5 is showing version=1.2.4.


richp405:
Since zoau replied the issue is in a few versions of their build, I'll switch job.py... will test this before marking the ticket ready for review.

richp405:
Update: Zoau team is making the change in 1.2.5, so we don't need to switch job.py.  Closing the ticket as cancelled.

Issue #943, opened 2023-08-22
[Enabler] [zos_operator] Implement new wait argument


storfibm:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Ansible module

zos_operator, zos_operator_action_query

### Enhancement or feature description

ZOAU's opercmd in 1.2.5 has received a new toggle switch `wait`, which when active will cause opercmd to wait up until the specified `timeout` duration. Otherwise, opercmd will return sooner with a response if only the `timeout` variable is used. The z/OS Core Collection should support this mechanism to ensure consistent behavior.

**Acceptance Criteria**
Given the z/OS Core Collection
When a user wants to wait the duration of a specified timeout
Then zos_operator and zos_operation_action_query accomodate the behavior

**Requirements**
-Integrate new opercmd `wait` option within opercmd.execute() into zos_operator and zos_operator_action_query
-Dependency on ZOAU 1.2.5

**Additional context**
https://jsw.ibm.com/browse/NAZARE-10054

PR #976, opened 2023-09-21
Enhancement/943/add wait zos operator and query

richp405:
##### SUMMARY
Enhancement to zos_operator, adding the 'wait' function back in with a new definition
Enhancement to zos_operator_action_query, adding both wait and wait_time_s

Addresses #943 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_operator
zos_operator_action_query

##### ADDITIONAL INFORMATION
The timing system in zoau has changed to add new options:
 - wait_time_s is the maximum number of seconds to wait for a response

- If wait is True, zoau will always wait the maximum amount of time
- If wait is False, zoau will return after data is received
 
If wait is not included, it defaults to False.

- name: Execute operator command to show jobs, always waiting 5 seconds for response
  zos_operator:
    cmd: 'd a,all'
    wait_time_s: 5
    wait: true

- name: Execute operator command to show jobs, waiting up to 7 seconds for response
  zos_operator:
    cmd: 'd a,all'
    wait_time_s: 7

### DIFF
--- a/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
+++ b/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
@@ -1,6 +1,8 @@
 minor_changes:
   - zos_operator: Changed system to call 'wait=true' parameter to zoau call.
+      Requires zoau 1.2.5 or later.
       (https://github.com/ansible-collections/ibm_zos_core/pull/976)
-  - zos_operator_action_query: Add wait_time_s parameter in the operator_action_query.
+  - zos_operator_action_query: Add a max delay of 5 seconds on each part of the operator_action_query.
+      Requires zoau 1.2.5 or later.
       (https://github.com/ansible-collections/ibm_zos_core/pull/976)
 
--- a/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
+++ b/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
@@ -1,6 +1,6 @@
 minor_changes:
-  - zos_operator: Added the 'wait' parameter back in to use the new -w operator.
+  - zos_operator: Changed system to call 'wait=true' parameter to zoau call.
       (https://github.com/ansible-collections/ibm_zos_core/pull/976)
-  - zos_operator_action_query: Add wait_time_s and 'wait' parameters in the operator_action_query.
+  - zos_operator_action_query: Add wait_time_s parameter in the operator_action_query.
       (https://github.com/ansible-collections/ibm_zos_core/pull/976)
 
--- a/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
+++ b/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
@@ -1,4 +1,4 @@
-enhancements:
+minor_changes:
   - zos_operator: Added the 'wait' parameter back in to use the new -w operator.
       (https://github.com/ansible-collections/ibm_zos_core/pull/976)
   - zos_operator_action_query: Add wait_time_s and 'wait' parameters in the operator_action_query.
--- a/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
+++ b/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
@@ -1,5 +1,6 @@
 enhancements:
   - zos_operator: Added the 'wait' parameter back in to use the new -w operator.
-      (https://github.com/ansible-collections/ibm_zos_core/pull/xxx)
+      (https://github.com/ansible-collections/ibm_zos_core/pull/976)
   - zos_operator_action_query: Add wait_time_s and 'wait' parameters in the operator_action_query.
-      (https://github.com/ansible-collections/ibm_zos_core/pull/xxx)
+      (https://github.com/ansible-collections/ibm_zos_core/pull/976)
+
--- a/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
+++ b/changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
@@ -1,0 +1,5 @@
+enhancements:
+  - zos_operator: Added the 'wait' parameter back in to use the new -w operator.
+      (https://github.com/ansible-collections/ibm_zos_core/pull/xxx)
+  - zos_operator_action_query: Add wait_time_s and 'wait' parameters in the operator_action_query.
+      (https://github.com/ansible-collections/ibm_zos_core/pull/xxx)
--- a/plugins/modules/zos_operator.py
+++ b/plugins/modules/zos_operator.py
@@ -57,14 +57,11 @@ options:
     default: 1
   wait:
     description:
-      - Configuring wait used by the L(zos_operator,./zos_operator.html) module
-        has been deprecated and will be removed in a future ibm.ibm_zos_core
-        collection.
-      - Setting this option will yield no change, it is deprecated.
-      - Review option I(wait_time_s) to instruct operator commands to wait.
+      - Setting this option will tell opercmd to wait the full wait_time, instead
+        of returning on first data received
     type: bool
     required: false
-    default: true
+    default: false
 """
 
 EXAMPLES = r"""
@@ -81,12 +78,13 @@ EXAMPLES = r"""
   zos_operator:
     cmd: "\\$PJ(*)"
 
-- name: Execute operator command to show jobs, waiting up to 5 seconds for response
+- name: Execute operator command to show jobs, always waiting 8 seconds for response
   zos_operator:
     cmd: 'd a,all'
     wait_time_s: 5
+    wait: true
 
-- name: Execute operator command to show jobs, always waiting 7 seconds for response
+- name: Execute operator command to show jobs, waiting up to 7 seconds for response
   zos_operator:
     cmd: 'd a,all'
     wait_time_s: 7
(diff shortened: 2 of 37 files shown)


richp405:
image

richp405:
After discussion, I've updated the code to check for zoau 1.2.5 or later, and will re-request reviews when the ansible tests have passed.

richp405:
image

AndreMarcel99 (approved):
Clean and direct

changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
```diff
@@ -0,0 +1,6 @@
+enhancements:
```
fernandofloresg, line 1: enhancements is not a valid tag for changelogs, rather use minor_changes or major_changes. I think minor_changes will be ok.

plugins/modules/zos_operator.py
```diff
@@ -286,6 +289,23 @@ def run_operator_command(params):
     wait_s = params.get("wait_time_s")
     cmdtxt = params.get("cmd")
 
+    zv = ZOAU_API_VERSION.split(".")
```
fernandofloresg, line 292: Can we instead use module_utils/zoau_version_checker instead? If any changes occur in the future then we would just need to change in one place.

plugins/modules/zos_operator_action_query.py
```diff
@@ -251,7 +292,33 @@ def run_module():
     try:
         new_params = parse_params(module.params)
 
-        cmd_result_a = execute_command("d r,a,s")
+        kwargs = {}
+
+        wait_s = new_params.get("wait_time_s")
+
+        zv = ZOAU_API_VERSION.split(".")
```
fernandofloresg, line 299: same as with zos_operator, would prefer using module_utils/zoau_version_checker

fernandofloresg (dismissed):
Looks good to me, just a couple of comments

changelogs/fragments/943-enhance-Add-wait-zos-operator-and-query.yml
richp405: changed

plugins/modules/zos_operator_action_query.py
richp405: We will neeed to talk about this... there's a large issue with the zoau_version _checker

plugins/modules/zos_operator.py
richp405: I'd like to create a new ticket for that change, based on Wednesday's conversation.  It works, but I'll make the change so that zoau_version_checker calls the new code, and then this can use zoau_version_checker.

plugins/modules/zos_operator_action_query.py
richp405: I'd like to create a new ticket for that change, based on Wednesday's conversation. It works, but I'll make the change so that zoau_version_checker calls the new code, and then this can use zoau_version_checker.

ddimatos:
I have not reviewed the code yet but wonder would like to understand the value of `wait` in addition to `wait_time_s`, in some of the recent updates we removed `wait` and this does not align to decision.

So the opercmd docs are [here](https://www.ibm.com/docs/en/zoau/1.2.x?topic=reference-opercmd) for reference , my question is how more value do really get from having `wait`, if `wait` is true it will wait the full time of `wait_time_s` , does that mean if `wait` is false it will return sooner than `wait_time_s` , eg i set a really long `wait_time_s`, say 60 seconds and the command finishes in 30 seconds, it returns? Is there a problem with this, in the past I think consoles could write to a registry more than once , would those repeated writes not be captured since wait is true?

I feel like if someone says wait 60, you might just want to always honor that, i see the value in returning early but only if we can be sure that returning early does not truncate anything, in that case we could always set the value to true and avoid exposing it.

richp405:
@ddimatos This has been a back-and-forth.  Customers don't want a big delay all the time, and in some cases they don't want to miss data.  In the zoau slack channel, p1696597067029359 is a new entry, through support, asking about this.  Ironically, this one is complaining that 1.2.4's 1-second delay is too long.

The return, as I understand if from Anthony, is once something is received and stopped, so wait=false means a 60 second delay could return in 2-3 seconds if output was received.

ddimatos:
@richp405 thanks for that, while I agree no one wants to wait longer then needed especially in a user interactive situation, eg using the the ZOAU cmd line interface. With an Ansible experience where users are more interested in success over performance, they probably don't mind the extra wait time (wait=true - meaning we default to this and don't expose it) else what they face is failing tasks that might return to soon where responses are chunks. 

Ansible and ZOAU have two very difference audiences, I want to be sure we are always addressing the Ansible audience when exposing ZOAU options.

storfibm:
Based on the discussion above, I think the new wait option needs to be integrated silently. Don't expose a new option in the module. Updated #943 to drop the option requirement. Just driving -T in the future alone, however, may cause early returns and truncation issues.

richp405:
Rebuilt, removing option for wait, and passing through wait_arg=true (-w).

image

plugins/modules/zos_operator.py
```diff
@@ -286,6 +273,18 @@ def run_operator_command(params):
...
+    getit = False
+    if zv[0] > "1":
```
ddimatos, line 278: I feel like this version check is going to result in a bug, so zoau 1.2.5.0 which is checked by this line of code `elif zv[0] == "1" and zv[1] == "2" and zv[2] > "4":` (which is correct) , but notice this is an if/elif so the very first `if zv[0] > "1":` would evaluate to true if the version of zoau is less than `1.2.5.0`.

For example, if the version is 1.2.4.0, the first `if` will yield true and skip the last one `elif zv[0] == "1" and zv[1] == "2" and zv[2] > "4":` , all you need is the last check, so you can remove:
```
    if zv[0] > "1":
        getit = True
    elif zv[0] == "1" and zv[1] > "2":
        getit = True
```

also could you use something a bit more meaningful then `getit` as a variable name please, thank you.

plugins/modules/zos_operator.py
ddimatos: Please share the issue here so we can resolve this conversation, thanks @richp405

plugins/modules/zos_operator_action_query.py
```diff
@@ -251,7 +279,28 @@ def run_module():
...
+        getit = False
+        if zv[0] > "1":
```
ddimatos, line 288: I feel like this version check is going to result in a bug, so zoau 1.2.5.0 which is checked by this line of code elif zv[0] == "1" and zv[1] == "2" and zv[2] > "4": (which is correct) , but notice this is an if/elif so the very first if zv[0] > "1": would evaluate to true if the version of zoau is less than 1.2.5.0.

For example, if the version is 1.2.4.0, the first if will yield true and skip the last one elif zv[0] == "1" and zv[1] == "2" and zv[2] > "4": , all you need is the last check, so you can remove:

    if zv[0] > "1":
        getit = True
    elif zv[0] == "1" and zv[1] > "2":
        getit = True

also could you use something a bit more meaningful then getit as a variable name please, thank you.

ddimatos (changes requested):
Hi @richp405 thanks for the updates and speedy turn around, as I was reading the source I found what I think might be a bug in both modules, more so please don't rush to make changes just yet, I have some concerns about adding wait_time_s to `zos_operation_action_query` and trying to understand where the original request came from and more so why.

richp405:
Okay, variables renamed, calls and inputs shuffled, and through the smoke test:
image

plugins/modules/zos_operator.py
richp405: #1018

plugins/modules/zos_operator_action_query.py
richp405: #1018

plugins/modules/zos_operator.py
richp405: Git #1018 is the new issue, which shouldn't take a day to clear up.

ddimatos: Thanks @richp405 - i actually read the code incorrectly, my bad, I read the ">" as ">=", thanks for var update :)

plugins/modules/zos_operator_action_query.py
ddimatos: Thank you.

ddimatos (approved):
@richp405 - this looks good and thank you for the the added changes per the review comments, your great to work with, thanks so much Rich.

fernandofloresg (approved):
My comments around the zoau version checker are being addressed and the rest of this looks good to me.

richp405:
@ddimatos One last question: should we add version detection here, to decide if default delay is passed as 5 (old style) or 500 (for 1.3 and later?).

ddimatos:
@richp405 I did think about this and while I did not look their python source, my guess is the timeout is using shell interface option `-T` which various if you look at the [noteworthy changes here](https://www.ibm.com/docs/en/zoau/1.2.x?topic=whats-new-noteworthy#version-1-2-4-5) and search on `timeout` you will see the behavior changed from one release to another. Since `-w` is the new option in ZOAU 1.2.5 which forces the full time specified in -T , i don't think we are using -w (obviously through the python interface but correct me if i am wrong) such that we need to do version detection around this code. I may have misunderstood your question though.

In 1.2.5 in appears to wait up to the time but reads as if it will return early UNLESS -w is specified. 
```
New -w argument added to toggle behavior when waiting for command response output. Specifying this argument causes opercmd to wait the entire timeout period before returning.
    The default timeout value has been changed to 1 second. This is the maximum amount of time that opercmd will wait before returning, regardless if it received any command response.
```

In 1.2.4.1 Python accepts the shell interface default , no change on the shell

In 1.2.4 it appears to wait the full time.
```
 It now waits the exact amount of time specified in the TIMEOUT argument while collecting messages. This allows it to collect more complete message responses from commands that produce multiple single line messages.
```

Short answer, i don't think so, but correct me if I don't understand your question.

Issue #933, opened 2023-08-16
[Enabler] Generate Class to Manage Dataset Name


AndreMarcel99:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

By running the pipeline sometimes the errors with the enabler of volumes #625 generates an error because of the tracks on disposal is fully because some datasets aren't deleted and also to avoid repetition on name for every data set.

Issue #726, opened 2023-04-07
[Bug] [zos_apf] Test case occasionally fails


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Ansible module

zos_apf

### Enabler description

This is not a bug in Ansible Core's code but what I noticed is the occasional test case failure, I have raised this with ZOAU and looks like its going to be a 3Q work item thus we need to ensure we make time to validate. 

I think this is a timing issue, I think some where a data set is being listed maybe with `dls` and its taking to long or the code is non blocking thus moves on not waiting for the result of the data set to come back because this is intermittent.

There are many more details in zoau tracker: `9678`
Update: it might also be addressed by tracker `9679`.

Error:
```
    {'contacted': {'ec33017a.vmec.svl.ibm.com': {'_ansible_no_log': False,
    'changed': False,
    'invocation': {'module_args': {'batch': None,
                                   'force_dynamic': True,
                                   'library': 'APFTEST.P0000228.T0670444',
                                   'operation': None,
                                   'persistent': {'backup': False,
                                                  'backup_name': None,
                                                  'data_set_name': 'APFTEST.PRST.P6777445.T0562383',
                                                  'marker': '/* '
                                                            '{mark} '
                                                            'BLOCK '
                                                            '*/'},
                                   'sms': False,
                                   'state': 'present',
                                   'tmp_hlq': None,
                                   'volume': '000000'}},
    'rc': 8,
    'stderr': 'BGYSC4709E Was not '
              'able to find dataset '
              'APFTEST.P0000228.T0670444 '
              'in volume 000000.\n',
    'stderr_lines': ['BGYSC4709E Was '
                     'not able to '
                     'find dataset '
                     'APFTEST.P0000228.T0670444 '
                     'in volume '
                     '000000.'],
    'stdout': '',
    'stdout_lines': []}}
    }
```

I was able to recreate this in Python:
```
>>> from zoautil_py import zsystem, datasets
>>> opt='add'
>>> dsname='APFTEST.P0000228.T0670444'
>>> volume='000000'
>>> sms = False
>>> forceDynamic=True
>>> persist = {'addDataset': 'APFTEST.PRST.P6777445.T0562383', 'marker': "/* {mark} BLOCK */"}
>>> zsystem.apf(opt=opt, dsname=dsname, volume=volume, sms=sms, forceDynamic=forceDynamic, persistent=persist).stderr_response
[...]

PR #977, opened 2023-09-21
[v1.9.0] Enabler/625/remove hard coded vol user and datasets names from test cases

### LINKED ISSUE #625: [Enabler] Remove hard coded vol user from test cases
### Enabler description

This is easier said than done but in an effort to continue to ensure test cases are portable we need a strategy to make our volume selection dynamic. When I pointed the functional tests at our native system the zos_copy functional tests failed I believe [here](https://github.com/ansible-collections/ibm_zos_core/blob/20bdea300d13ff67b83563b065cb74ba24203dcd/tests/functional/modules/test_zos_copy_func.py#L1961) as volume `000000` does not exist and it seems to have hung.

We also need to validate this scenario with our code such what happens when a user passes an invalid volume. If it turns out its a bug lets create a separate bug report. 

```
@pytest.mark.seq
@pytest.mark.pdse
@pytest.mark.parametrize("src_type", ["seq", "pds", "pdse"])
def test_copy_data_set_to_volume(ansible_zos_module, src_type):
    hosts = ansible_zos_module
    source = "USER.TEST.FUNCTEST.SRC"
    dest = "USER.TEST.FUNCTEST.DEST"

    try:
        hosts.all.zos_data_set(name=source, type=src_type, state='present')
        copy_res = hosts.all.zos_copy(
            src=source,
            dest=dest,
            remote_src=True,
            volume='000000'
        )

        for cp in copy_res.contacted.values():
            assert cp.get('msg') is None
            assert cp.get('changed') is True
            assert cp.get('dest') == dest

        check_vol = hosts.all.shell(
            cmd="tsocmd \"LISTDS '{0}'\"".format(dest),
            executable=SHELL_EXECUTABLE,
        )

        for cv in check_vol.contacted.values():
            assert cv.get('rc') == 0
            assert "000000" in cv.get('stdout')
    finally:
        hosts.all.zos_data_set(name=source, state='absent')
        hosts.all.zos_data_set(name=dest, state='absent')
```

```
{"changed": false, "message": "", "names": ["USER.TEST.FUNCTEST.SRC"], "invocation": {"module_args": {"name": "USER.TEST.FUNCTEST.SRC", "type": "seq", "state": "present", "space_type": "M", "replace": false, "batch": null, "space_primary": null, "space_secondary": null, "record_format": null, "sms_management_class": null, "sms_storage_class": null, "sms_data_class": null, "block_size": null, "directory_blocks": null, "record_length": null, "key_offset": null, "key_length": null, "volumes": null, "tmp_hlq": null}}}\n', b"OpenSSH_8.6p1, LibreSSL 3.3.6\r\ndebug1: Reading configuration data /Users/ddimatos/.ssh/config\r\ndebug1: Reading configuration data /etc/ssh/ssh_config\r\ndebug1: /etc/ssh/ssh_config line 21: include /etc/ssh/ssh_config.d/* matched no files\r\ndebug1: /etc/ssh/ssh_config line 54: Applying options for *\r\ndebug2: resolve_canonicalize: hostname <sanitize IP> is address\r\ndebug3: expanded UserKnownHostsFile '~/.ssh/known_hosts' -> '/Users/ddimatos/.ssh/known_hosts'\r\ndebug3: expanded UserKnownHostsFile '~/.ssh/known_hosts2' -> '/Users/ddimatos/.ssh/known_hosts2'\r\ndebug1: Authenticator provider $SSH_SK_PROVIDER did not resolve; disabling\r\ndebug1: auto-mux: Trying existing master\r\ndebug2: fd 3 setting O_NONBLOCK\r\ndebug2: mux_client_hello_exchange: master version 4\r\ndebug3: mux_client_forwards: request forwardings: 0 local, 0 remote\r\ndebug3: mux_client_request_session: entering\r\ndebug3: mux_client_request_alive: entering\r\ndebug3: mux_client_request_alive: done pid = 31545\r\ndebug3: mux_client_request_session: session request sent\r\ndebug1: mux_client_request_session: master session id: 2\r\ndebug3: mux_client_read_packet: read header failed: Broken pipe\r\ndebug2: Received exit status from master 0\r\n")
```
```
<Sanitize IP> (0, b'\n{"changed": false, "message": "", "names": ["USER.TEST.FUNCTEST.SRC"], "invocation": {"module_args": {"name": "USER.TEST.FUNCTEST.SRC", "type": "seq", "state": "present", "space_type": "M", "replace": false, "batch": null, "space_primary": null, "space_secondary": null, "record_format": null, "sms_management_class": null, "sms_storage_class": null, "sms_data_class": null, "block_size": null, "directory_blocks": null, "record_length": null, "key_offset": null, "key_length": null, "volumes": null, "tmp_hlq": null}}}\n', b"OpenSSH_8.6p1, LibreSSL 3.3.6\r\ndebug1: Reading configuration data /Users/ddimatos/.ssh/config\r\ndebug1: Reading configuration data /etc/ssh/ssh_config\r\ndebug1: /etc/ssh/ssh_config line 21: include /etc/ssh/ssh_config.d/* matched no files\r\ndebug1: /etc/ssh/ssh_config line 54: Applying options for *\r\ndebug2: resolve_canonicalize: hostname <sanitize ip>  is address\r\ndebug3: expanded UserKnownHostsFile '~/.ssh/known_hosts' -> '/Users/ddimatos/.ssh/known_hosts'\r\ndebug3: expanded UserKnownHostsFile '~/.ssh/known_hosts2' -> '/Users/ddimatos/.ssh/known_hosts2'\r\ndebug1: Authenticator provider $SSH_SK_PROVIDER did not resolve; disabling\r\ndebug1: auto-mux: Trying existing master\r\ndebug2: fd 3 setting O_NONBLOCK\r\ndebug2: mux_client_hello_exchange: master version 4\r\ndebug3: mux_client_forwards: request forwardings: 0 local, 0 remote\r\ndebug3: mux_client_request_session: entering\r\ndebug3: mux_client_request_alive: entering\r\ndebug3: mux_client_request_alive: done pid = 31545\r\ndebug3: mux_client_request_session: session request sent\r\ndebug1: mux_client_request_session: master session id: 2\r\ndebug3: mux_client_read_packet: read header failed: Broken pipe\r\ndebug2: Received exit status from master 0\r\n")
META: ran handlers
```

### Ansible module

zos_copy

AndreMarcel99:
##### SUMMARY
Rewrite test suites as zos_apf and zos_encode to ensure atomicity by test.
Change all hardcoded content in all test suites of datasets names and values and change by the dynamic functions to get unique values or volumes to use by system to running tests.
Add function to get aleatory and valid names for datasets names as fixture.
Add a class and fixture to manage the volumes to use by system and do not fail in entrance of system.
All the new functions added expected to conclude in a correct running of parallel in al tests

Fixes #625 and #933 
##### COMPONENT NAME
ls_volumes - Wrapper class to manage volumes
get_volumes - fixture by session to get volumes are on use by system
get_dataset - fixture to get a unique name of dataset to avoid error dataset already on the system

##### ADDITIONAL INFORMATION
Class volumes run one time by session start to get volumes on the system to use and every test that needs to ask to the class that manage the name and if is on use or not, as values of the class Volumes and ensures not repeat the volumes in a test with two volumes if is need give volume 0 (000000) as is the one with more tracks to use it. One the test finish call the in_use to free, the last way is for the systems in parallel.

The fixture get_dataset works for test case calling first and aleatory high level qualifier of size that can be default 8 but some JCL found to extend so could change as parameter of the function and to ensure the unique mvstmp call a unique name base on the date and time with specifications.

Captura de pantalla 2023-09-21 a la(s) 11 20 39

There are some test cases that required PDS already on the system and also USS as tmp/etc that will keep behavior did not change the structure, but are few ones.

[Pipeline runs](https://ibm.box.com/s/44ab6b0u3f3s56apl6cmsx5rmdosi9cf)


tests/helpers/volumes.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, name):
+        self.name = name
+        self.in_use = False
+
+    def use(self):
+        self.in_use = True
+
+    def free(self):
+        self.in_use = False
+
```
richp405, line 32: Question: should you add a __repr__ or a __str__ function, so that the content and volume of the class can be shown?  Not a requirement, but a suggestion.

richp405 (changes requested):
Please check that look for stdout gathering.

ddimatos:
I started to review this @AndreMarcel99 and decided to hold off because on first pass I noticed the conftest.py (fixture) was modified and while I have not read all the code, it would be helpful for me to have a mindset of why this was modified over simply creating a new python class. Lets say we move away from the fixture (pytest-ansible) we lose this data set and volume code ,  I know its not hard to copy and paste it elsewhere but it seems like we did not need this to be in a fixture, can you help me understand if there was a good reason for that? The only reason that comes to mind is that you wanted concurrency safety here, which I still don't know you will have by putting in a fixture. 

Second thing, this does not seem to be a design patter of fixtures unless i am missing something but the function within the function seems a bit unnecessary. 

image

AndreMarcel99:
@ddimatos The reason for get_dataset() inside a fixture as a function is to be callable as many time as needed in a test, because if were just a fixture with no function to return just return one name even if is call 20 times. But yes all can be move, the main reason to had as a fixture is for the access to the connection.

tests/functional/modules/test_zos_apf_func.py
```diff
@@ -75,62 +44,48 @@
 /*ENDAPFLIST*/"""
 
 
-def run_shell_cmd(hosts, cmdStr):
+def get_stdout(hosts, cmdStr):
     results = hosts.all.shell(cmd=cmdStr)
-    pprint(vars(results))
     for result in results.contacted.values():
         out = result.get("stdout")
```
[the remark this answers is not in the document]
AndreMarcel99: No the result.get("stdout") catch the full output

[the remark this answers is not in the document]
fernandofloresg: As it is should be ok @richp405, that is because even if that is under a "for" statement it is only one result fetched.

tests/helpers/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+import string
+import random
+import re
+
+# Function or test to ensure random names of datasets
+    # Is need the hosts to call the shell in every test and for some cases of
+    # long datasets names that will generate problems with jcl the hlq size can
+    # change
+def get_dataset(hosts, hlq_size=8):
```
fernandofloresg, line 13: Couple of thoughts about this function.

1- get_dataset name is just too general for me, would prefer "get_random_data_set_name" or "get_tmp_ds_name", or something similar that indicates to me when reading this, that the output will be a temporary data set name that most likely will be a unique name.

2- Better use data set as opposed to dataset.

fernandofloresg: Forgot to mention, Use a docstring (https://peps.python.org/pep-0257/) for this function, this will be used a lot and having those details you commented above the function will be pretty useful.

tests/helpers/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+    letters =  string.ascii_uppercase
+    hlq =  ''.join(random.choice(letters)for iteration in range(hlq_size))
+    # Ensure the hlq is correct for the dataset naming conventions, until then continue working
+    while not re.fullmatch(
+    r"^(?:[A-Z$#@]{1}[A-Z0-9$#@-]{0,7})",
+            hlq,
+            re.IGNORECASE,
+        ):
+        hlq =  ''.join(random.choice(letters)for iteration in range(hlq_size))
+    # Get the second part of the name with the ocmand mvstmp by time is give
```
fernandofloresg, line 24: typo in "ocmand" to "command"

tests/helpers/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+
+# Function or test to ensure random names of datasets
+    # Is need the hosts to call the shell in every test and for some cases of
+    # long datasets names that will generate problems with jcl the hlq size can
+    # change
+def get_dataset(hosts, hlq_size=8):
+    # Generate the first random hlq of size pass as parameter
+    letters =  string.ascii_uppercase
+    hlq =  ''.join(random.choice(letters)for iteration in range(hlq_size))
+    # Ensure the hlq is correct for the dataset naming conventions, until then continue working
```
fernandofloresg, line 17: change "Ensure the hlq is correct for the dataset naming conventions, until then continue working" to
"Generate a random HLQ and verify if is valid, if not, repeat the process"

tests/helpers/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+    while not re.fullmatch(
```
fernandofloresg, line 18: Having a while in here could theorically generate a loop. Add a counter into it to remove that possibility.

tests/helpers/dataset.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import absolute_import, division, print_function
```
fernandofloresg, line 1: Missing copyright info here.

tests/helpers/volumes.py
```diff
@@ -0,0 +1,57 @@
...
+# limitations under the License.
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import pytest
+
+class Volume:
+# Class volume to save the information with volume name and a flag of use
```
fernandofloresg, line 21: Change this comments to instead use docstrings.

Also phrasing is a bit confusing, I would recommend:

"
Volume handler class which manages volume assignment for tests, volume status retrieval from the system and keeps track of current volumes in use by the test session. 
"

fernandofloresg: Hold on, this is for below class instead.  In this case is better to say something like "Volume class represents a volume on the z system, it tracks if the status of the volume with respect to the current test session."

tests/helpers/volumes.py
fernandofloresg: This might be useful for test logging which we will definitely need to improve since now both data set names and volume assignation will be dynamic for the test suite.

tests/helpers/volumes.py
```diff
@@ -0,0 +1,57 @@
...
+    def __str__(self):
+        return f'The volume {self.name} is in {self.in_use} in use'
+
+    def use(self):
+        self.in_use = True
+
+    def free(self):
+        self.in_use = False
+
+class ls_Volume:
```
fernandofloresg, line 36: ls_Volume is not a representative name for a class, based on the actions performed by this class a better name will be VolumeHandler or VolumeManager

tests/helpers/volumes.py
```diff
@@ -0,0 +1,57 @@
...
+# Class to call fixture get_volumes and filled the array and manage volumes
+# instances of are in use or not
+    def __init__(self, *get_volumes):
+        self.volume = get_volumes
```
fernandofloresg, line 40: self.volume is a list, is better to change the variable name to volume_list but in this case I will recommend converting the output of get_volumes to a set or dict, that way we can avoid having to do a for loop every time we want to access an specific volume.

tests/helpers/volumes.py
```diff
@@ -0,0 +1,57 @@
...
+    for volume in ls_vols.volume:
+        if not (volume.in_use):
+            volume.use()
+            return volume.name
+    print("Not more volumes in disposal return volume 000000")
+    return "000000"
+
+def free_vol(vol, ls_vols):
+    # Check from the array the volume is already free for other test to use
+    for volume in ls_vols.volume:
```
fernandofloresg, line 55: We can reduce the for loop using a dict instead as ls_vols.volume

tests/helpers/volumes.py
```diff
@@ -0,0 +1,57 @@
...
+
+class ls_Volume:
+# Class to call fixture get_volumes and filled the array and manage volumes
+# instances of are in use or not
+    def __init__(self, *get_volumes):
+        self.volume = get_volumes
+
+
+def get_available_vol(ls_vols):
+    # Check in the list of volumes one on use or not, also send a default
```
fernandofloresg, line 44: Use docstrings instead of simple comments

tests/functional/modules/test_zos_tso_command_func.py
```diff
@@ -18,8 +18,10 @@
 import ansible.constants
 import ansible.errors
 import ansible.utils
+from ibm_zos_core.tests.helpers.dataset import (
+    get_dataset)
 
-DEFAULT_TEMP_DATASET="imstestl.ims1.temp.ps"
+#DEFAULT_TEMP_DATASET="imstestl.ims1.temp.ps"
```
fernandofloresg, line 24: You can remove this instead of commenting it

fernandofloresg (changes requested):
Added some comments, will continue reviewing this tomorrow and focus on fixture use vs using a class to fetch the volumes

tests/conftest.py
```diff
@@ -84,6 +89,40 @@ def ansible_zos_module(request, z_python_interpreter):
     except Exception:
         pass
 
+    # Call of the class by the class ls_Volume (volumes.py file) as many times needed
+    # one time the array is filled
+@pytest.fixture(scope="session")
+def get_volumes(ansible_zos_module):
+    # Call the pytest-ansible plugin to execute the command d u,dasd,online
```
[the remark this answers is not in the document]
fernandofloresg: code-wise I would like to see something like this:


```suggestion
    
   Try:
    VolumeHandler.get_available_volumes()
Except VolumeRetrievalError:
    // Add some code to log "Error retrieving list of available volumes in the system"
    VolumeHandler.set_default_volume()
    // Add some code to log "Setting default volume to {0}".format(VolumeHandler.DEFAULT_VOLUME)
```

fernandofloresg, line 96: Use a docstring instead

tests/functional/modules/test_zos_archive_func.py
```diff
@@ -352,12 +351,15 @@ def test_uss_archive_remove_targets(ansible_zos_module, format):
 def test_mvs_archive_single_dataset(ansible_zos_module, format, data_set, record_length, record_format):
     try:
         hosts = ansible_zos_module
+        DATASET = get_dataset(hosts)
```
fernandofloresg, line 354: I know I wrote this and now I see it's my fault, but there are things we can correct here :).
Is there a reason for the variable names to be kept in uppercase? Recommend reading https://peps.python.org/pep-0008/#function-and-variable-names and I think is better to keep them as lowercase.

 I was using data_set.get("name") in various places, which is not a descriptive name and although having a variable name DATASET and MVS_DEST_ARCHIVE are more descriptive why not use src_data_set and dest_data_set or archive_data_set?

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -15,19 +15,16 @@
...
 
-VOLUME = "222222"
-VOLUME2 = "222222"
-VOLUME_TO_BACKUP = VOLUME
-BIG_VOLUME = "DSHRL1"
-BIG_VOLUME2 = "DSHRL2"
-DATA_SET_NAME = "USER.PRIVATE.TESTDS"
-DATA_SET_NAME2 = "USER.PRIVATE.TESTDS2"
-DATA_SET_PATTERN = "USER.PRIVATE.*"
+#DATA_SET_NAME = "USER.PRIVATE.TESTDS"
```
fernandofloresg, line 25: We can remove this instead of commenting

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -199,6 +196,7 @@ def assert_file_does_not_exist(hosts, path):
 )
 def test_backup_of_data_set(ansible_zos_module, backup_name, overwrite, recover):
     hosts = ansible_zos_module
+    DATA_SET_NAME = get_dataset(hosts)
```
fernandofloresg, line 199: name variable is ok, just change it to lowercase

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -444,8 +440,6 @@ def test_backup_and_restore_of_multiple_data_sets(ansible_zos_module, data_set_i
             hlq=NEW_HLQ,
         )
         assert_module_did_not_fail(results)
-        assert_data_set_exists(hosts, DATA_SET_RESTORE_LOCATION)
```
fernandofloresg, line 447: Also why is this assertion removed? I understand the current test case intents to verify that the data set exists after restore

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -470,7 +464,7 @@ def test_backup_and_restore_exclude_from_pattern(ansible_zos_module):
         )
         results = hosts.all.zos_backup_restore(
             operation="backup",
-            data_sets=dict(include=DATA_SET_PATTERN, exclude=DATA_SET_NAME2),
+            data_sets=dict(include=DATA_SET_NAME, exclude=DATA_SET_NAME2),
```
fernandofloresg, line 467: This change also affects the test case, that is because in this part we are looking to match a pattern. Fix should be simple, instead of using the complete data set name just strip from right until the next dot and then add "*", that should be your pattern to test.

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -483,8 +477,6 @@ def test_backup_and_restore_exclude_from_pattern(ansible_zos_module):
             hlq=NEW_HLQ,
         )
         assert_module_did_not_fail(results)
-        assert_data_set_exists(hosts, DATA_SET_RESTORE_LOCATION)
-        assert_data_set_does_not_exist(hosts, DATA_SET_RESTORE_LOCATION2)
```
fernandofloresg, line 487: Also I don't see the reason to remove these assertions from the test.

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -411,17 +407,17 @@ def test_backup_and_restore_of_data_set_when_restore_location_exists(
...
     [
-        [DATA_SET_NAME, DATA_SET_NAME2],
-        DATA_SET_PATTERN,
+        ["USER.PRIVATE.TESTDS", "USER.PRIVATE.TESTDS2"],
+        "USER.PRIVATE.*",
     ],
 )
 def test_backup_and_restore_of_multiple_data_sets(ansible_zos_module, data_set_include):
     hosts = ansible_zos_module
+    DATA_SET_NAME = "USER.PRIVATE.TESTDS"
```
fernandofloresg, line 416: Why do we keep hardcoded content in this specific test case? If is to make sure those don't exist we can use zos_data_set replace option right?

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -20,6 +22,7 @@
 
 __metaclass__ = type
 
+DEFAULT_HLQ = "USER.PRIVATE"
```
fernandofloresg, line 25: this is also a hard coded content right? Maybe we can include in the same data set name generator a function that generates random valid HLQs.

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -939,9 +942,8 @@ def test_ds_block_insertafter_regex(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(insertafter="ZOAU_ROOT=", block="ZOAU_ROOT=/mvsutil-develop_dsed\nZOAU_HOME=\\$ZOAU_ROOT\nZOAU_DIR=\\$ZOAU_ROOT", state="present")
-    test_name = "DST1"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
fernandofloresg, line 945: oh messing with getting information from the stack in this tests I don't think is the way to go. Since is just used to create a temp file why not let tempfile name it instead?

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -1157,7 +1150,7 @@ def test_ds_tmp_hlq_option(ansible_zos_module):
             hlq = result.get("stdout")
         if len(hlq) > 8:
             hlq = hlq[:8]
-        ds_full_name = hlq + "." + test_name.upper() + "." + ds_type
+        ds_full_name = hlq + "." + get_dataset(hosts)
```
fernandofloresg, line 1153: I think get_dataset currently gets you a ds name with an HLQ, what happens if get_dataset returns a data set name with the maximum name size allowed?  If we are doing this is better to also have a function that generates ds_names without HLQ.

That way in your data set name generator you can have:

1- Function to generate a random valid HLQ.
2- Function to generate a random valid ds name without HLQ.
3- Function that essentially uses 1 and 2 to generate a complete random valid ds name

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -1435,6 +1425,7 @@ def test_ds_not_supported(ansible_zos_module, dstype):
 def test_ds_block_insertafter_regex_fail(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
+    DEFAULT_DATA_SET_NAME = get_dataset(hosts)
```
fernandofloresg, line 1428: better to use lowercase now that we are doing a refactor

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1960,8 +1966,9 @@ def test_ensure_copy_file_does_not_change_permission_on_dest(ansible_zos_module,
 @pytest.mark.seq
 @pytest.mark.parametrize("ds_type", ["PDS", "PDSE", "SEQ"])
 def test_copy_dest_lock(ansible_zos_module, ds_type):
-    DATASET_1 = "USER.PRIVATE.TESTDS"
-    DATASET_2 = "ADMI.PRIVATE.TESTDS"
+    hosts = ansible_zos_module
+    DATASET_1 = get_dataset(hosts)
+    DATASET_2 = get_dataset(hosts)
```
fernandofloresg, line 1971: Same comment about lowercase, I think I will stop commenting this and just let you change that in other test cases.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -2949,19 +2955,20 @@ def test_copy_pds_to_existing_pds(ansible_zos_module, args):
         hosts.all.zos_data_set(name=src, state="absent")
         hosts.all.zos_data_set(name=dest, state="absent")
 
-
+#Special case for call a program
```
fernandofloresg, line 2958: I didn't get this comment, can you explain?

fernandofloresg (changes requested):
Will continue commenting in a next review, just doing a checkpoint

richp405 (approved):
Fernando has several good readability comments.  I approve assuming you will respond to his comments.

tests/functional/modules/test_zos_copy_func.py
AndreMarcel99: With the addition of hlq size is fixed

tests/conftest.py
ddimatos: Its resolved

tests/helpers/volumes.py
ddimatos: The code has since been changed, best guess is this is addressed when the code now has `all_volumes = hosts.all.zos_operator(cmd="d u,dasd,online,,65536")`

tests/helpers/volumes.py
ddimatos: Its been renamed to `Volume`, thus resolved

tests/helpers/volumes.py
ddimatos: I can see a __str__ function to dump the object , thus resolved.
```
   def __str__(self):
        return f'The volume {self.name} is in {self.in_use} in use'
```

tests/helpers/volumes.py
ddimatos: I updated it, resolved.

tests/helpers/dataset.py
ddimatos: Resolved

tests/helpers/dataset.py
ddimatos: Resolved

tests/helpers/dataset.py
ddimatos: resolved

tests/helpers/dataset.py
ddimatos: Resolved

tests/functional/modules/test_zos_apf_func.py
```diff
@@ -75,62 +44,48 @@
...
-    run_shell_cmd(hosts, cmdStr)
-    return prstds
-
-
-def persistds_delele(hosts, ds):
-    cmdStr = "drm {0}".format(ds)
-    run_shell_cmd(hosts, cmdStr)
-
-
 def set_test_env(hosts, test_info):
```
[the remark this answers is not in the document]
ddimatos: Resolved

tests/functional/modules/test_zos_apf_func.py
```diff
@@ -75,62 +44,48 @@
 /*ENDAPFLIST*/"""
 
 
-def run_shell_cmd(hosts, cmdStr):
+def get_stdout(hosts, cmdStr):
```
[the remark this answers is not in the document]
ddimatos: Resolved

changelogs/fragments/977-remove-hard-coded-vols-and-datasets.yml
```diff
@@ -0,0 +1,19 @@
...
+  - zos_fetch - Remove hard coded volumes and datasets names in test.
+  - zos_find - Remove hard coded volumes in test.
+  - zos_job_query - Remove hard coded volumes and datasets names in test.
+  - zos_job_submit - Remove hard coded volumes and datasets names in test.
+  - zos_lineinfile - Remove hard coded volumes and datasets names in test.
+  - zos_mount - Remove hard coded volumes and datasets names in test.
+  - zos_mvs_raw - Remove hard coded volumes and datasets names in test.
+  - zos_tso_command - Remove hard coded volumes and datasets names in test.
+  - zos_unarchive - Remove hard coded volumes and datasets names in test.
+  - volumes - Class to manage volumes in the system.
```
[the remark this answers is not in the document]
ddimatos: Resolved

changelogs/fragments/977-remove-hard-coded-vols-and-datasets.yml
```diff
@@ -0,0 +1,19 @@
+trivial:
+  - zos_apf - Test suite was hard to follow. Change made easy to follow.
```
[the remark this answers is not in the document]
ddimatos: Resolved

ddimatos (changes requested):
In this review, I focused on reducing the existing noise, meaning the unresolved comments, I went through all the comments, validated the changes and closed them. In doing so I found about 6 or so compile errors, I assume it was some refactoring but no regression had been run after. I did correct the compile errors, I also addressed some of the comments that were unresolved such as lower case'ing some vars, changing other var names, I did not run regression yet, but I also plan to do a review of my own now. 

Please do see one unresolved comment I left for you @AndreMarcel99 
https://github.com/ansible-collections/ibm_zos_core/pull/977/files#r1374865263

tests/functional/modules/test_zos_fetch_func.py
```diff
@@ -291,7 +296,8 @@ def test_fetch_vsam_data_set(ansible_zos_module):
             None
             os.remove(dest_path)
         hosts.all.file(path=USS_FILE, state="absent")
-        hosts.all.file(path=TEMP_JCL_PATH, state="absent")
+        hosts.all.file(path=temp_jcl_path, state="absent")
+        volumes.free_vol(volume_1)
```
ddimatos, line 300: Should free_vol bet in an finally block if its important to free?

tests/functional/modules/test_zos_find_func.py
```diff
@@ -298,6 +303,8 @@ def test_find_vsam_in_volume(ansible_zos_module):
             batch=[dict(name=i, state='absent') for i in VSAM_NAMES]
         )
         hosts.all.zos_data_set(name=alternate_vsam, state='absent')
+        volumes.free_vol(volume_1)
```
ddimatos, line 306: Should these be in a finally block in case the test fails and they don't get freed?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -550,9 +554,8 @@ def test_ds_line_insertafter_regex(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(insertafter="ZOAU_ROOT=", line="ZOAU_ROOT=/mvsutil-develop_dsed", state="present")
-    test_name = "DST1"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 557: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -573,9 +576,8 @@ def test_ds_line_insertbefore_regex(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(insertbefore="ZOAU_ROOT=", line="unset ZOAU_ROOT", state="present")
-    test_name = "DST2"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 579: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -596,9 +598,8 @@ def test_ds_line_insertafter_eof(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(insertafter="EOF", line="export 'ZOAU_ROOT'", state="present")
-    test_name = "DST3"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 601: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -618,9 +619,8 @@ def test_ds_line_insertbefore_bof(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(insertbefore="BOF", line="# this is file is for setting env vars", state="present")
-    test_name = "DST4"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 622: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -641,9 +641,8 @@ def test_ds_line_replace_match_insertafter_ignore(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="ZOAU_ROOT=", insertafter="PATH=", line="ZOAU_ROOT=/mvsutil-develop_dsed", state="present")
-    test_name = "DST5"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 644: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -664,9 +663,8 @@ def test_ds_line_replace_match_insertbefore_ignore(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="ZOAU_ROOT=", insertbefore="PATH=", line="unset ZOAU_ROOT", state="present")
-    test_name = "DST6"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 666: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -687,9 +685,8 @@ def test_ds_line_replace_nomatch_insertafter_match(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="abcxyz", insertafter="ZOAU_ROOT=", line="ZOAU_ROOT=/mvsutil-develop_dsed", state="present")
-    test_name = "DST7"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 688: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -710,9 +707,8 @@ def test_ds_line_replace_nomatch_insertbefore_match(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="abcxyz", insertbefore="ZOAU_ROOT=", line="unset ZOAU_ROOT", state="present")
-    test_name = "DST8"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 710: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -733,9 +729,8 @@ def test_ds_line_replace_nomatch_insertafter_nomatch(ansible_zos_module, dstype)
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="abcxyz", insertafter="xyzijk", line="ZOAU_ROOT=/mvsutil-develop_dsed", state="present")
-    test_name = "DST9"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 732: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -756,9 +751,8 @@ def test_ds_line_replace_nomatch_insertbefore_nomatch(ansible_zos_module, dstype
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="abcxyz", insertbefore="xyzijk", line="unset ZOAU_ROOT", state="present")
-    test_name = "DST10"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 754: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -779,9 +773,8 @@ def test_ds_line_absent(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(regexp="ZOAU_ROOT=", line="", state="absent")
-    test_name = "DST11"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 776: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -974,9 +958,8 @@ def test_ds_line_does_not_insert_repeated(ansible_zos_module, dstype):
     hosts = ansible_zos_module
     ds_type = dstype
     params = dict(line='ZOAU_ROOT=/usr/lpp/zoautil/v100', state="present")
-    test_name = "DST15"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 961: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -1034,9 +1017,8 @@ def test_ds_encoding(ansible_zos_module, encoding, dstype):
     insert_data = "Insert this string"
     params = dict(insertafter="SIMPLE", line=insert_data, state="present")
     params["encoding"] = encoding
-    test_name = "DST13"
-    temp_file = "/tmp/{0}".format(test_name)
-    ds_name = test_name.upper() + "." + ds_type
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 1020: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -18,6 +18,10 @@
 import pytest
 import inspect
 
+from ibm_zos_core.tests.helpers.dataset import (
+    get_tmp_ds_name,
+    get_random_hlq)
+
 __metaclass__ = type
 
 DEFAULT_DATA_SET_NAME = "USER.PRIVATE.TESTDS"
```
ddimatos, line 27: This seems to be not in use anymore, all the functions override this, maybe remove it and lower case DEFAULT_DATA_SET_NAME to default_data_set_name , given lower case is usually local and upper case global.

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -803,17 +796,11 @@ def test_ds_tmp_hlq_option(ansible_zos_module):
     ds_type = "SEQ"
     kwargs = dict(backup_name=r"TMPHLQ\..")
     params = dict(insertafter="EOF", line="export ZOAU_ROOT", state="present", backup=True, tmp_hlq="TMPHLQ")
-    test_name = "DST12"
-    temp_file = "/tmp/zos_lineinfile/" + test_name
+    temp_file = "/tmp/{0}".format(inspect.stack()[0][3])
```
ddimatos, line 799: Same as earlier, do we need to inspect the stack for a temp file?

tests/functional/modules/test_zos_mvs_raw_func.py
```diff
@@ -51,6 +51,7 @@ def test_failing_name_format(ansible_zos_module):
 def test_disposition_new(ansible_zos_module):
     try:
         hosts = ansible_zos_module
+        DEFAULT_DATA_SET = get_tmp_ds_name(hosts)
```
ddimatos, line 54: Lower case DEFAULT_DATA_SET, use upper case only for global (top level) variables.

I won't repeat this , but do so for the entire class please, thank you.

tests/functional/modules/test_zos_mvs_raw_func.py
```diff
@@ -1693,7 +1741,7 @@ def test_concatenation_fail_with_unsupported_dd_type(ansible_zos_module):
                             ),
                             dict(
                                 dd_data_set=dict(
-                                    data_set_name=DEFAULT_DATA_SET,
+                                    data_set_name="USER.PRIVATE.TEST",
```
ddimatos, line 1744: We don't have to figure this out now, but I would like to undertand why "USER.PRIVATE.TEST" remains hard coded ,  I may have to review the code.

tests/functional/modules/test_zos_unarchive_func.py
```diff
@@ -369,12 +366,16 @@ def test_uss_unarchive_copy_to_remote(ansible_zos_module):
 def test_mvs_unarchive_single_data_set(ansible_zos_module, format, data_set, record_length, record_format):
     try:
         hosts = ansible_zos_module
+        MVS_DEST_ARCHIVE = get_tmp_ds_name(hosts)
+        DATASET = get_tmp_ds_name(hosts, 4)[0:29]
```
ddimatos, line 370: Use lower case for local variables, there are several here.

ddimatos, line 370: I would like to understand why we need to truncate data set names, so long as it stays shorter than 44 (sum of all qualifiers) should not be an issue. Regarding `DATASET = get_tmp_ds_name(hosts, 4)[0:29]`, this happens in a few other places a well.

ddimatos (changes requested):
@AndreMarcel99 - this looks like a lot of comments but really its just 4 - 6 comments constantly being repeated. The volume suggestion I made can wait, we can create a new issue for that, it will b considerable work. 

Thank you, if you want we can go over this on a call.

tests/helpers/dataset.py
```diff
@@ -0,0 +1,54 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+import pytest
+import string
+import random
+import re
+
+def get_tmp_ds_name(hosts, hlq_size=8, hlq=None):
```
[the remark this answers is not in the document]
ddimatos: You can consider this resolved, i recalled the behavior for mvstmp incorrectly, it actually states in the doc that it only generates a name. The code was correct, but I understand you will probably improve the performance by not making a call as you shared with me.

tests/functional/modules/test_zos_apf_func.py
```diff
@@ -74,63 +43,45 @@
...
 
 
-def set_test_env(hosts, test_info):
-    # results = hosts.all.zos_data_set(name=ds, type="SEQ")
+def test_add_del(ansible_zos_module):
+    hosts = ansible_zos_module
+    test_info = dict(library="", state="present", force_dynamic=True)
     cmdStr = "mvstmp APFTEST"
-    ds = run_shell_cmd(hosts, cmdStr)[:25]
+    results = hosts.all.shell(cmd=cmdStr)
```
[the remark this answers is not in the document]
ddimatos: I understand now that the call to mvstmp with APFTEST was so that the code could honor the static dictionaries. Feel free to resolve this if you don't have time to change things.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -4319,18 +4321,23 @@ def test_backup_pds(ansible_zos_module, args):
...
+def test_copy_data_set_to_volume(ansible_zos_module, volumes_on_systems, src_type):
     hosts = ansible_zos_module
-    source = "USER.TEST.FUNCTEST.SRC"
-    dest = "USER.TEST.FUNCTEST.DEST"
-
+    source = get_tmp_ds_name(hosts)
+    dest = get_tmp_ds_name(hosts)
+    volumes = Volume_Handler(volumes_on_systems)
+    volume_1 = volumes.get_available_vol()
+    if volume_1 == "SCR03":
```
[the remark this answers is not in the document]
ddimatos: Actually my comment to say this breaks portability is not true, this should not impact it, feel free to respond to this, but i understand SCCR03 is s small volume and unstable.

tests/functional/modules/test_zos_mvs_raw_func.py
AndreMarcel99: Hard but fixed

tests/functional/modules/test_zos_apf_func.py
```diff
@@ -200,65 +178,106 @@ def test_add_del_persist(ansible_zos_module):
...
-    for result in results.contacted.values():
-        assert result.get("rc") == 0
-    del_exptd = del_expected.replace(" ", "")
-    cmdStr = "cat \"//'{0}'\" ".format(test_info['persistent']['data_set_name'])
-    actual = run_shell_cmd(hosts, cmdStr).replace(" ", "")
-    assert actual == del_exptd
-    for item in test_info['batch']:
-        clean_test_env(hosts, item)
-    persistds_delele(hosts, test_info['persistent']['data_set_name'])
+#def test_batch_add_del(ansible_zos_module):
```
AndreMarcel99, line 225: Test comment related to [issue](https://github.com/ansible-collections/ibm_zos_core/issues/726)

tests/functional/modules/test_zos_encode_func.py
```diff
@@ -77,25 +77,60 @@
...
+    if ds_type == "KSDS":
+        params["key_length"] = key_length
+        params["key_offset"] = key_offset
+
+    hosts.all.zos_data_set(**params)
+
+    if add_data:
+        record_src = "/tmp/zos_copy_vsam_record"
+
+        hosts.all.zos_copy(content=VSAM_RECORDS, dest=record_src)
```
fernandofloresg, line 110: I think I mentioned this in another PR, but we want to avoid introducing new dependencies between modules in the test cases, we can use `hosts.all.shell` to do this with `echo` instead of using `zos_copy`

fernandofloresg: Is better to fix this now instead of having to do another round of development before merging into `dev`

tests/functional/modules/test_zos_encode_func.py
```diff
@@ -455,11 +500,22 @@ def test_uss_encoding_conversion_mvs_pds_to_uss_path(ansible_zos_module):
...
         hosts.all.file(path=USS_DEST_PATH, state="absent")
 
 
 def test_uss_encoding_conversion_mvs_ps_to_mvs_pds_member(ansible_zos_module):
     hosts = ansible_zos_module
+    MVS_PDS = get_tmp_ds_name()
+    MVS_PDS_MEMBER = MVS_PDS + '(MEM)'
+    MVS_PS = get_tmp_ds_name()
+    hosts.all.zos_data_set(name=MVS_PS, state="present", type="seq")
+    hosts.all.zos_copy(content=TEST_DATA, dest=MVS_PS)
```
fernandofloresg, line 514: I think I mentioned this in another PR, but we want to avoid introducing new dependencies between modules in the test cases, we can use hosts.all.shell to do this with echo instead of using zos_copy.

Is better to remove this now instead of doing another round of development before merging into `dev`.

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -17,10 +17,12 @@
 import re
 import pytest
 import inspect
+import os
```
fernandofloresg, line 20: GH is marking `import os` as a new line, but I don't see it being used in the tests, if not used please remove.

tests/functional/modules/test_zos_mvs_raw_func.py
fernandofloresg: but why is it hard to remove? what is the failure? what did you tried and why a new disposition on other parameter affects this?

fernandofloresg (changes requested):
A couple of changes requested, don't mean pick on you or something, I just think there are still questions unanswered and don't want to include new dependencies between modules in the test cases.

tests/functional/modules/test_zos_apf_func.py
ddimatos: Good points, seems safe to resolve , we have an open issue for it for 1.3 so we will uncomment it.

ddimatos (approved):
I have only one open question at this point in the review, basically I think I understand why the data set was hard coded; if am correct, to change this is a bit out of scope now and best we create a backlog entry for it. See the comment here: https://github.com/ansible-collections/ibm_zos_core/pull/977/files#r1451691637

tests/functional/modules/test_zos_encode_func.py
```diff
@@ -15,12 +15,14 @@
...
+from ibm_zos_core.tests.helpers.dataset import (
+    get_dataset)
 
 __metaclass__ = type
 
-USS_FILE = "/tmp/encode.data"
+USS_FILE = "/tmp/encode_data"
 USS_NONE_FILE = "/tmp/none"
-USS_DEST_FILE = "/tmp/converted.data"
+USS_DEST_FILE = "/tmp/converted_data"
```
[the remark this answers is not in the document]
AndreMarcel99: Is another task to change all the USS files

tests/functional/modules/test_zos_mvs_raw_func.py
```diff
@@ -1474,6 +1516,7 @@ def test_concatenation_with_data_set_member(ansible_zos_module):
 def test_concatenation_with_unix_dd_and_response(ansible_zos_module):
```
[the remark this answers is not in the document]
AndreMarcel99: It was one test with parameters for datasets and other for USS just added the difference to avoid be the same name

tests/functional/modules/test_zos_data_set_func.py
```diff
@@ -147,17 +148,21 @@ def print_results(results):
 @pytest.mark.parametrize(
     "jcl",
     [PDS_CREATE_JCL, KSDS_CREATE_JCL, RRDS_CREATE_JCL, ESDS_CREATE_JCL, LDS_CREATE_JCL],
+    ids=['PDS_CREATE_JCL', 'KSDS_CREATE_JCL', 'RRDS_CREATE_JCL', 'ESDS_CREATE_JCL', 'LDS_CREATE_JCL']
```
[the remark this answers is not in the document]
AndreMarcel99: Is for debugging test cases, without JCL the test just show name(full test of PDS_CREATE_JCL) but now is just the ID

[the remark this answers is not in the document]
fernandofloresg: Solved

tests/functional/modules/test_zos_encode_func.py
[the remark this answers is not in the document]
fernandofloresg: We agreed on removing hard coded content for USS in another occasion

tests/functional/modules/test_zos_mvs_raw_func.py
fernandofloresg: should be ok

tests/functional/modules/test_zos_lineinfile_func.py
fernandofloresg: done

tests/functional/modules/test_zos_encode_func.py
fernandofloresg: done

tests/functional/modules/test_zos_encode_func.py
fernandofloresg: done

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1960,19 +1962,20 @@ def test_ensure_copy_file_does_not_change_permission_on_dest(ansible_zos_module,
...
     else:
-        src_data_set = DATASET_1
-        dest_data_set = DATASET_2
+        src_data_set = data_set_1
+        dest_data_set = data_set_2
     try:
         hosts = ansible_zos_module
-        hosts.all.zos_data_set(name=DATASET_1, state="present", type=ds_type, replace=True)
-        hosts.all.zos_data_set(name=DATASET_2, state="present", type=ds_type, replace=True)
+        hosts.all.zos_data_set(name=data_set_1, state="present", type="pdse", replace=True)
```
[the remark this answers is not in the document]
fernandofloresg: done

fernandofloresg (approved):

PR #978, opened 2023-09-22
Modified versions in bug issue template to avoid users picking a non-existing version

fernandofloresg:
##### SUMMARY

Since some users have picked wrong versions that are not yet available we agreed to change the bug options to only latest released and "to release", so I kept v1.8.0-beta.1 in case a user founds and issue in dev branch.

didn't write a changelog fragment entry because is not a change to the codebase or testing suite.

##### ISSUE TYPE

- Docs Pull Request


AndreMarcel99 (approved):

richp405 (approved):
Thank you... that will help!

PR #980, opened 2023-09-26
Added latest to allow member copy when disp=shr

### LINKED ISSUE #735: [Enhancement] [zos_copy] Support disposition share (DSP=SHR) / copy force

### Ansible module

zos_copy

Presently, `zos_copy` does not support interactions with data sets that are already in use by other processes (DSP=SHR). Copy operations that involve such dispositions will error out even when users know that it is safe to perform the overwrite. This effort seeks to introduce the ability to proceed with operations via a force flag or similar, still giving the user the choice.

**Dependency**
Depends on ZOAU's upcoming dcp force option

**Acceptance Criteria**
Given a z/OS core collection user
When zos_copy is invoked
Then user has the ability to force the operation to proceed even when other processes have a lock

**Requirements**
-Incorporate functionality into `force` behavior and provide appropriate feedback
-Provide sufficient warnings within the zos_copy module and other documentation collateral
-Works with sequential, PDS/E and member data set operations

fernandofloresg:
##### SUMMARY

zos_copy does not support interactions with data sets that are already in use by other processes (DSP=SHR). With these changes using force flag now user can issue copies for these data sets.

Fixes #735 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_copy

### DIFF
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -158,6 +158,22 @@ options:
     type: bool
     default: false
     required: false
+  force_lock:
+    description:
+      - By default, when c(dest) is a MVS data set and is being used by another
+        process with DISP=SHR or DISP=OLD the module will fail; use C(force_lock)
+        to bypass this check and continue with copy.
+      - If set to C(true) and destination is a MVS data set opened by another
+        process with DISP=SHR then zos_copy will copy using DISP=SHR.
+      - Using C(force_lock) uses operations that are subject to race conditions
+        and can lead to data loss, use with caution.
+      - If a dataset member has aliases, and is not a program
+        object, copying that member to a dataset that is in use will result in
+        the aliases not being preserved in the target dataset. When this scenario
+        occurs the module will fail.
+    type: bool
+    default: false
+    required: false
   ignore_sftp_stderr:
     description:
       - During data transfer through SFTP, the module fails if the SFTP command
@@ -786,7 +802,7 @@ class CopyHandler(object):
         is_binary=False,
         executable=False,
         backup_name=None,
-        force=False,
+        force_lock=False,
     ):
         """Utility class to handle copying data between two targets
 
@@ -806,7 +822,7 @@ class CopyHandler(object):
         self.is_binary = is_binary
         self.executable = executable
         self.backup_name = backup_name
-        self.force = force
+        self.force_lock = force_lock
 
     def run_command(self, cmd, **kwargs):
         """ Wrapper for AnsibleModule.run_command """
@@ -838,7 +854,7 @@ class CopyHandler(object):
         if self.is_binary:
             copy_args["options"] = "-B"
 
-        if self.force:
+        if self.force_lock:
             copy_args["options"] += " -f"
 
         response = datasets._copy(new_src, dest, None, **copy_args)
@@ -860,7 +876,7 @@ class CopyHandler(object):
             src {str} -- The name of the source VSAM
             dest {str} -- The name of the destination VSAM
         """
-        out_dsp = "shr" if self.force else "old"
+        out_dsp = "shr" if self.force_lock else "old"
         dds = {"OUT": "{0},{1}".format(dest.upper(), out_dsp)}
         repro_cmd = """  REPRO -
         INDATASET('{0}') -
@@ -1417,7 +1433,7 @@ class PDSECopyHandler(CopyHandler):
         is_binary=False,
         executable=False,
         backup_name=None,
-        force=False,
+        force_lock=False,
     ):
         """ Utility class to handle copying to partitioned data sets or
         partitioned data set members.
@@ -1436,7 +1452,7 @@ class PDSECopyHandler(CopyHandler):
             is_binary=is_binary,
             executable=executable,
             backup_name=backup_name,
-            force=force,
+            force_lock=force_lock,
         )
 
     def copy_to_pdse(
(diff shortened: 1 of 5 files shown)


fernandofloresg:
Screenshot 2023-10-02 at 4 12 12 PM


Screenshot 2023-10-02 at 4 12 42 PM


Test files run:
+ python3 -m pytest -vvv -x --durations=0 --host-pattern=all -Z=/var/lib/jenkins/workspace/zosAnsible-Fernando/ibm_zos_core/configuration.yml ../tests/functional/modules/test_zos_encode_func.py ../tests/functional/modules/test_zos_copy_func.py ../tests/functional/modules/test_zos_archive_func.py ../tests/functional/modules/test_zos_lineinfile_func.py ../tests/functional/modules/test_zos_fetch_func.py ../tests/functional/modules/test_zos_blockinfile_func.py ../tests/functional/modules/test_zos_data_set_func.py ../tests/functional/modules/test_zos_mount_func.py

plugins/modules/zos_copy.py
```diff
@@ -778,7 +794,8 @@ def __init__(
         module,
         is_binary=False,
         executable=False,
-        backup_name=None
+        backup_name=None,
+        force_lock=False,
```
rexemin, line 798: I think it would be helpful if you added a description in the docstring for the `force_lock` argument, at least on this class.

rexemin (commented):
Looks good, thanks for working on this. I just have one question: I noticed you added support for `force_lock` for all types of data sets, but only actually use that support when copying to a partitioned data set, why is that? Or am I missing something?

fernandofloresg:
> Looks good, thanks for working on this. I just have one question: I noticed you added support for `force_lock` for all types of data sets, but only actually use that support when copying to a partitioned data set, why is that? Or am I missing something?

Does it? Is also included in the _copy_to_seq and _copy_to_vsam_ methods, and the test case creates a sequential data set as well. I think that is sufficient for those cases right?

plugins/modules/zos_copy.py
fernandofloresg: Agree, done

rexemin:
> > Looks good, thanks for working on this. I just have one question: I noticed you added support for `force_lock` for all types of data sets, but only actually use that support when copying to a partitioned data set, why is that? Or am I missing something?
> 
> Does it? Is also included in the _copy_to_seq and _copy_to_vsam_ methods, and the test case creates a sequential data set as well. I think that is sufficient for those cases right?

Sorry, what I mean is that you updated both `copy_to_seq` and `copy_to_vsam` to support the new option, but on [line 2,618](https://github.com/ansible-collections/ibm_zos_core/blob/enhancement/735/copy-disp-shr/plugins/modules/zos_copy.py#L2618) you didn't pass the value of `force_lock` when instantiating `copy_handler`, which in turn is the object that calls the copy functions.

```python
    copy_handler = CopyHandler(
        module,
        is_binary=is_binary,
        executable=executable,
        backup_name=backup_name
    )
```

fernandofloresg:
> > > Looks good, thanks for working on this. I just have one question: I noticed you added support for `force_lock` for all types of data sets, but only actually use that support when copying to a partitioned data set, why is that? Or am I missing something?
> > 
> > 
> > Does it? Is also included in the _copy_to_seq and _copy_to_vsam_ methods, and the test case creates a sequential data set as well. I think that is sufficient for those cases right?
> 
> Sorry, what I mean is that you updated both `copy_to_seq` and `copy_to_vsam` to support the new option, but on [line 2,618](https://github.com/ansible-collections/ibm_zos_core/blob/enhancement/735/copy-disp-shr/plugins/modules/zos_copy.py#L2618) you didn't pass the value of `force_lock` when instantiating `copy_handler`, which in turn is the object that calls the copy functions.
> 
> ```python
>     copy_handler = CopyHandler(
>         module,
>         is_binary=is_binary,
>         executable=executable,
>         backup_name=backup_name
>     )
> ```

You're right. Will change that and test. In the meantime will add the do not merge tag.

fernandofloresg:
Screenshot 2023-10-04 at 6 38 22 AM


Done

rexemin (approved):
Thanks for working so quickly on the things I mentioned, looks really good!

plugins/modules/zos_copy.py
```diff
@@ -151,6 +151,22 @@
...
+    description:
+      - By default, when c(dest) is a MVS data set and is being used by another
+        process with DISP=SHR or DISP=OLD the module will fail. Use C(force_lock)
+        to bypass this check and continue with copy.
+      - If set to C(true) and destination is a MVS data set opened by another
+        process then zos_copy will try to copy using DISP=SHR.
+      - Using C(force_lock) uses operations that are subject to race conditions
+        and can lead to data loss, use with caution.
+      - If a dataset member has aliases, and is not a program
+        object, copying that member to a dataset that is in use will result in
```
ddimatos, line 164: This is minor, but `dataset` should be `data set` , couple of places in the `force_lock` description.

Not a necessary change and if you update, i don't think you need to run regression.

ddimatos (approved):
@fernandofloresg this looks great, really clean and well written. I had one minor comment that I did not consider necessary to change about the use of `dataset` vs `data set`. Thanks so much, so many folks are waiting on this.

plugins/modules/zos_copy.py
fernandofloresg: Agree, I copied this from ZOAU docs and didn't pay attention

Issue #621, opened 2023-01-24
[Bug] zos_copy at 1.4.0 or above fails if src dataset has ALIAS. 1.3.6 works fine. 


adn1107:
### Bug description

TASK [ims_dataset : Copy System Datasets] **************************************
Monday 23 January 2023  14:46:08 -0800 (0:00:40.033)       0:05:01.496 ******** 
failed: [wazi] (item={'src': 'DFSF10.SDFSRESL', 'dest': 'IMSTEST.IMWZ.SDFSRESL'}) => {"ansible_loop_var": "item", "changed": false, "item": {"dest": "IMSTEST.IMWZ.SDFSRESL", "src": "DFSF10.SDFSRESL"}, "msg": "Unable to copy data set member DFSF10.SDFSRESL to data set member IMSTEST.IMWZ.SDFSRESL(ALIAS(DBFDBAC1)", "rc": 1, "stderr": "BGYSC1205E LISTCAT processing failed with pattern 66342.\nBGYSC1205E LISTCAT processing failed with pattern 50397965.\n", "stderr_lines": null, "stdout": "cp: FSUM6258 cannot open file \"//'DFSF10.SDFSRESL(ALIAS(DBFDBAC1)'\": EDC5047I An invalid file name was specified as a function parameter.\n", "stdout_lines": null}

### Ansible version

```shell
ansible --version
ansible [core 2.13.5]
  config file = None
  configured module search path = ['/Users/adn/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/adn/workspaces/nazare/venv/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/adn/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/adn/workspaces/nazare/venv/bin/ansible
  python version = 3.9.6 (default, Sep 26 2022, 11:37:49) [Clang 14.0.0 (clang-1400.0.29.202)]
  jinja version = 3.1.2
  libyaml = True
```

### IBM z/OS Ansible core Version

v1.4.0

### IBM ZOAU version

v1.2.2

### Ansible module

zos_copy

ddimatos:
Hi @adn1107 - are you looping over a PDS or PDSE and assuming one of the members is an ALIAS, is that how you encountered this? 

Anything more you might have like the playbook or verbose output will help expedite this.

ddimatos:
[...]

PR #1014, opened 2023-10-07
Enhancement/423/zos copy add data set member alias support

### LINKED ISSUE #423: [Enhancement] [zos_copy]  Support Data Set member Aliases
### Enhancement description

This is an enhancement because aliases were not intentionally supported.

# Enhancement

The `zos_copy` module does not copy any aliases, take for example the snippet below:
```
zos_copy:
 src: SYS2.MQMM001.SCSQSNLE(*)
 dest: SYS2.MQM.LPALIB.ANS
 remote_src: true
```

Where the data set  `SYS2.MQMM001.SCSQSNLE` when browsed has:
```
Name Prompt Alias-of Size
_________ CSQNL28E 00000A28
_________ CSQ3ECMX CSQNL28E 00000A28
```

In this case, only the member would be copied to the dest and not the alias.

## Source Problem
`zos_copy` uses ZOAU to list the members with source similar to `members = list(map(str.strip, datasets.list_members(src)))`  only no Aliases are returned which results in `zos_copy` not knowing about the alias and passing the listing `zos_copy` handler `copy_to_member` which ultimately invokes `datasets._copy(new_src, dest)` on a list does not include an alias.

# Notes
- The issue originally was in internal archive `C01E25VTZ3L/p1622058177044000`
- There are a number of ways to handle this, requesting ZOAU `mls` and `dls` support reading aliases
- An alias is like a *nix symlink, we would not want to copy an alias such that it results in a physical copy, that would defeat the purpose. We would want to ensure/check if something is an alias and make another alias, this should either be done by zos_copy or zoau methods (ensure we don't make another physical copy)
- Some ideas if we want to do this solely in zos_copy is to remove the `datasets.list_members(src) `call and insert our own commands or REXX that would check if a member is an Alias before doing a copy, basically separating members from aliases and then for the Aliases we could run a command like so in TSO to create the Alias for the `zos_copy` function `RENAME 'IMSTESTL.DIMATO.JCLCOPY(DATA)' 'IMSTESTL.DIMATO.JCLCOPY(DATAX)' ALIAS` and to list the members and Aliases `LISTDS 'IMSTESTL.DIMATO.JCLCOPY' MEMBERS`

image

# Comments
## From @adn1107 
IEBCOPY COPYGROUP option [can be used to copy aliases](https://www.ibm.com/docs/en/zos/2.1.0?topic=program-copying-objects-copygrp-copygroup-statements) 

It is recommended that you use the COPYGRP or COPYGROUP statement to copy PDS load modules or PDSE program objects and their aliases from or to a PDSE data set. Program objects can have aliases that are longer than eight characters. Using the COPYGRP or COPYGROUP statement will ensure that these longer aliases are copied along with their member.

Use the COPYGRP or COPYGROUP statement to begin a group copy, unload, or load. A group consists of a member and all of its aliases. COPYGRP or COPYGROUP treats the group as a single entity.

COPYGRP can be used to copy a data set when either the input data set or the output data set, or both, are PDSE:
PDSE to PDSE
PDSE to PDS
PDS to PDSE
COPYGROUP can be used to copy a data set when either the input data set or the output data set, or both, are PDS or PDSE:
PDSE to PDSE
PDSE to PDS
PDS to PDSE
PDS to PDS
For unloading groups:
Using COPYGRP: PDSE to PS
Using COPYGROUP: PDSE to PS, or PDS to PS
For loading groups:
Using COPYGRP: PS to PDSE
Using COPYGROUP: PS to PDSE, or PS to PDS
If neither data set is a PDSE, the request is treated the as a COPY operation subject to the syntax requirements of COPYGRP.

When using the COPYGRP or COPYGROUP statement:

    All aliases in a group will be copied with the member or neither the aliases or the member in a group will be copied.
    There can be only one INDD per copy operation.
    You can use the SELECT statement to selectively copy members. Either the member name or an alias can be specified to copy the member and all of its aliases.
    Do not indicate replace (R) on the SELECT statement.
    The EXCLUDE statement is not supported.

### Ansible module

zos_copy

ketankelkar:
##### SUMMARY
Introduces a new option `aliases` in the zos_copy module to preserve aliases when copying executable libraries (MVS or USS) to partitioned data sets.

Fixes #423 
see supporting issue #1015 

This also addresses - #621 
##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/changelogs/fragments/1014-zos-copy-add-data-set-member-aliases.yml
+++ b/changelogs/fragments/1014-zos-copy-add-data-set-member-aliases.yml
@@ -1,0 +1,5 @@
+minor_changes:
+- zos_copy - introduces a new option 'aliases' to enable preservation of member aliases
+  when copying data to partitioned data sets (PDS) destinations from USS or other PDS sources.
+  Copying aliases of text based members to/from USS is not supported.
+  (https://github.com/ansible-collections/ibm_zos_core/pull/1014)
--- a/docs/source/modules/zos_copy.rst
+++ b/docs/source/modules/zos_copy.rst
@@ -171,6 +171,17 @@ executable
   | **type**: bool
 
 
+aliases
+  If set to ``true``, indicates that any aliases found in the source (USS file, USS dir, PDS/E library or member) are to be preserved during the copy operation.
+
+  Aliases are implicitly preserved when libraries are copied over to USS destinations. That is, when ``executable=True`` and ``dest`` is a USS file or directory, this option will be ignored.
+
+  Copying of aliases for text-based data sets from USS sources or to USS destinations is not currently supported.
+
+  | **required**: False
+  | **type**: bool
+
+
 local_follow
   This flag indicates that any existing filesystem links in the source tree should be followed.
 
@@ -687,12 +698,21 @@ Examples
          record_format: VB
          record_length: 150
 
-   - name: Copy a Program Object on remote system to a new PDSE member MYCOBOL.
+   - name: Copy a Program Object and its aliases on a remote system to a new PDSE member MYCOBOL
      zos_copy:
        src: HLQ.COBOLSRC.PDSE(TESTPGM)
        dest: HLQ.NEW.PDSE(MYCOBOL)
        remote_src: true
        executable: true
+       aliases: true
+
+       - name: Copy a Load Library from a USS directory /home/loadlib to a new PDSE
+     zos_copy:
+       src: '/home/loadlib/'
+       dest: HLQ.LOADLIB.NEW
+       remote_src: true
+       executable: true
+       aliases: true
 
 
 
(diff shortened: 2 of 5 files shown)


richp405 (changes requested):
I believe this needs a changelog/fragment for approval.

rexemin (commented):
Looks good. As Rich said, it needs a changelog fragment

richp405 (approved):
Thanks for the changelog.

fernandofloresg (approved):
Looks good to me, great that you cleaned the test cases and implicitly tested the newly added option "LIBRARY".

ketankelkar:
above commits were comments/docs, re-ran that portion of the pipeline for validation - 
Screenshot 2023-10-11 at 1 55 36 PM

PR #1019, opened 2023-10-10
Staging v1.7.0 merge to main

ddimatos:
##### SUMMARY
Staging 1.7.0 tasks are for the version 1.7.0 release, this addresses issue #888.
This PR includes all 1.7.0 content ready to merge into main and later be tagged and released. 
Note, there has been no code change from 1.7.0-beta.2 to promote 1.7.0, all changes are meta and doc related only.

See issue #888 for more details.

PR #1023, opened 2023-10-13
Staging v1.7.0 merge to main (#1019)

ddimatos:
This merges the meta data changes made in version 1.7.0 (`git cherry-pick  4726116`) for issue #888 
From this PR [here](https://github.com/ansible-collections/ibm_zos_core/commit/4726116eac771a3db633400b2a3fd932a5d652f3) 
We need to have these changes in our `dev` branch will be Q4 1.9.0-beta.1, the metadata changes and doc changes must travel be pushed upstream to `dev`.

* Galaxy 1.7 updates
* Update meta runtime to support ansible-core 2.14 or later
* Update ibm_zos_core_meta.yml with updated version
* Update readme to align to supported ansible versions and new urls
* Added additional sanity ignore files to the exclude list
* Added additional sanity ignore files to the exclude list for ansible-lint.
* Update copyright yrs for source files that were overlooked
* Remove requirements from module doc, rely on offerings minimum requirements, also zoau 1.2.1 never was supported
* Add changelog summary for 1.7
* Adding generated antsibull-changelog release changelog and artifacts
* Remove v1.7.0_summary, its no longer needed
* Update release notes for ac 1.7.0
* Remove unsupported collection versions requiring a version of zoau that is EOS

### DIFF
--- a/plugins/action/zos_copy.py
+++ b/plugins/action/zos_copy.py
@@ -1,4 +1,4 @@
-# Copyright (c) IBM Corporation 2019, 2020, 2021, 2022
+# Copyright (c) IBM Corporation 2019-2023
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -25,7 +25,11 @@ exclude_paths:
   - tests/helpers
   - tests/requirements.txt
   - tests/unit
+  - tests/sanity/ignore-2.9.txt
   - tests/sanity/ignore-2.10.txt
+  - tests/sanity/ignore-2.11.txt
+  - tests/sanity/ignore-2.12.txt
+  - tests/sanity/ignore-2.13.txt
   - venv*
 parseable: true
 quiet: false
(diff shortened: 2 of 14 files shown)


plugins/action/zos_copy.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) IBM Corporation 2019, 2020, 2021, 2022
+# Copyright (c) IBM Corporation 2019-2023
```
richp405, line 1: I don't think this is a biggie, but it is inconsistent... other changes have year, year, year, etc.  This has year-year.  Minor nit-pick

richp405 (approved):
One minor note about inconsistent copyright years... otherwise, it looks fine.

plugins/action/zos_copy.py
ddimatos: @richp405 - I have not followed up with legal but the reason i compressed it down from the comma delimited is because a range can be used for sequential yrs.

My understanding is a range in  software programs that are constantly being revised and updated incrementally can be used like in the change I pushed. The ‘start’ year should be the date of the oldest content still contained in the work, with the ‘end’ year being the last year that the work was revised.

What I am looking into is what if we skip a year after a range, where with commas its easy to identify the year of the change but will use up space, once we have entered into a hyphenated model, we only focus on the start and end dates and not in between.

AndreMarcel99 (approved):

PR #1027, opened 2023-10-16
main change to version checker for consolidation

### LINKED ISSUE #1018: [Enhancement] Consolidate zoau version checks

### Ansible module

zos_operator

Currently there are 4 routines that check to zoau version: zos_operator, zos_operator_action_query, utils/job.py and and utils/zoau_version_checker.py

We need to consolidate these for both speed and consistency, so the code is more maintainable in the future.  

The routines need to be replaced with the fastest, accurate check, and the calls need to point to this new routine.

richp405:
##### SUMMARY
Consolidate the different version checks to use one routine

Fixes #1018

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- zos_operator
- zos_operator_action_query
- utils/job.py and 
- utils/zoau_version_checker.py

### DIFF
--- a/plugins/module_utils/zoau_version_checker.py
+++ b/plugins/module_utils/zoau_version_checker.py
@@ -102,19 +102,6 @@ def get_zoau_version_str():
                                provision for a 4th level eg "v1.2.0.1".
 
     """
-    # zoaversion_out = subprocess.run(
-    #    'zoaversion', shell=True, capture_output=True, check=False
-    # )
-    # version_list = (
-    #    zoaversion_out
-    #    .stdout
-    #    .decode('UTF-8')
-    #    .strip()
-    #    .split(' ')[3][1:]
-    #    .split('.')
-    #)
-    zoaversion_out = ZOAU_API_VERSION.split('.')
-
     version_list = (
         ZOAU_API_VERSION.split('.')
     )
--- a/plugins/module_utils/zoau_version_checker.py
+++ b/plugins/module_utils/zoau_version_checker.py
@@ -13,8 +13,12 @@
 
 from __future__ import absolute_import, division, print_function
 
-import subprocess
+# import subprocess
 
+try:
+    from zoautil_py import ZOAU_API_VERSION
+except Exception:
+    ZOAU_API_VERSION = "1.2.0"
 
 __metaclass__ = type
 
@@ -96,16 +100,23 @@ def get_zoau_version_str():
     Returns:
         { [int, int, int] } -- ZOAU version found in format [#,#,#]. There is a
                                provision for a 4th level eg "v1.2.0.1".
+
     """
-    zoaversion_out = subprocess.run(
-        'zoaversion', shell=True, capture_output=True, check=False
-    )
+    # zoaversion_out = subprocess.run(
+    #    'zoaversion', shell=True, capture_output=True, check=False
+    # )
+    # version_list = (
+    #    zoaversion_out
+    #    .stdout
+    #    .decode('UTF-8')
+    #    .strip()
+    #    .split(' ')[3][1:]
+    #    .split('.')
+    #)
+    zoaversion_out = ZOAU_API_VERSION.split('.')
+
     version_list = (
-        zoaversion_out
-        .stdout
-        .decode('UTF-8')
-        .strip()
-        .split(' ')[3][1:]
-        .split('.')
+        ZOAU_API_VERSION.split('.')
     )
+
     return version_list
--- a/tests/functional/modules/test_zos_job_submit_func.py
+++ b/tests/functional/modules/test_zos_job_submit_func.py
@@ -529,7 +529,12 @@ def test_job_submit_max_rc(ansible_zos_module, args):
                 #Expecting: - "The job return code 8 was non-zero in the job output, this job has failed"
                 #           - Consider using module zos_job_query to poll for a long running job or
                 #             increase option \\'wait_times_s` to a value greater than 10.",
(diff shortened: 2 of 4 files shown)


richp405:
Note: I merged in dev and one of the tests blew up.  Finally got that fixed.

image

fernandofloresg (approved):
I like that this removes the subprocess necessity, you could have remove the import line instead of commenting but is ok for me.

plugins/module_utils/zoau_version_checker.py
```diff
@@ -96,16 +100,10 @@ def get_zoau_version_str():
...
-        'zoaversion', shell=True, capture_output=True, check=False
-    )
     version_list = (
-        zoaversion_out
-        .stdout
-        .decode('UTF-8')
-        .strip()
-        .split(' ')[3][1:]
-        .split('.')
+        ZOAU_API_VERSION.split('.')
```
ddimatos, line 106: @ketankelkar - can you comment on why you had the decode(utf-8) initially here that is being removed. I imagine the shell call to zoauversion is going to print out in the native locale which is going to be a EBCDIC code page, so wondering if there is any concern of this change. @richp405 feel free to also comment.

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -529,7 +529,12 @@ def test_job_submit_max_rc(ansible_zos_module, args):
                 #Expecting: - "The job return code 8 was non-zero in the job output, this job has failed"
                 #           - Consider using module zos_job_query to poll for a long running job or
                 #             increase option \\'wait_times_s` to a value greater than 10.",
-                if result.get('duration') >= args["wait_time_s"]:
+                if result.get('duration'):
+                    duration = result.get('duration')
```
ddimatos, line 533: I don't really recall what comes back from get duration but i doubt its ever a non-zero value which leads me to question the `duration = 0` else stmt, will we mask the second part of this test by setting it to 0? 

Here:  assert re.search(r'non-zero', repr(result.get("msg")))

Comes down to that if `duration = result.get('duration')` can return non-zero values which I don't think it does, we might need the duration to remain undefined so it can hit the non-zero else noted above.

ddimatos (changes requested):

plugins/module_utils/zoau_version_checker.py
ddimatos: Oh never mind, you are using the python api variable `ZOAU_API_VERSION`, sorry about that, i read to fast, it is nice subproccess is removed.

ddimatos (approved):

ddimatos (changes requested):
@richp405 - I have 2 comments/concerns:
1) First being about the test case change defaulting to a 0, vs it remaining none which might have been the intention to trigger a non-zero error. 
2) Using python var `ZOAU_VERSION` might cause stack traces for users who try to use zoau versions less than `1.2.1` which is where I see the var `ZOAU_VERSION` was made available. Granted we do state the minimum versions of ZOAU but don't have any tooling to pre-check this so if a users happens to use 1.1.1 with the latest code which is not supported they may end up with a stack trace.

tests/functional/modules/test_zos_job_submit_func.py
richp405: duration is how long (in seconds) a job took to run.  So, if the job runs over, fails before start, or fails to complete, there are several pathways where duration is not returned.

If duration is undefined, the test_zos_job_submit_func routine crashes unit testing, because it is comparing a NonType to an integer, which python doesn't like to do.

I put the value of 0 in there, so that 'duration' is always defined, even if it isn't returned.

richp405:
> @richp405 - I have 2 comments/concerns:
> 
> 1. First being about the test case change defaulting to a 0, vs it remaining none which might have been the intention to trigger a non-zero error.
> 2. Using python var `ZOAU_VERSION` might cause stack traces for users who try to use zoau versions less than `1.2.1` which is where I see the var `ZOAU_VERSION` was made available. Granted we do state the minimum versions of ZOAU but don't have any tooling to pre-check this so if a users happens to use 1.1.1 with the latest code which is not supported they may end up with a stack trace.

I think you marked #2 (about version) as resolved.  We avoid the stack trace by doing try import/except, so there should always be a value when version checker is imported.

tests/functional/modules/test_zos_job_submit_func.py
ddimatos: We went through this today, likely we need to correct zos_job_submit, I see your point and can consider this resolved, thanks @richp405

ddimatos:
> > @richp405 - I have 2 comments/concerns:
> > 
> > 1. First being about the test case change defaulting to a 0, vs it remaining none which might have been the intention to trigger a non-zero error.
> > 2. Using python var `ZOAU_VERSION` might cause stack traces for users who try to use zoau versions less than `1.2.1` which is where I see the var `ZOAU_VERSION` was made available. Granted we do state the minimum versions of ZOAU but don't have any tooling to pre-check this so if a users happens to use 1.1.1 with the latest code which is not supported they may end up with a stack trace.
> 
> I think you marked #2 (about version) as resolved. We avoid the stack trace by doing try import/except, so there should always be a value when version checker is imported.

Thanks Rich, I missed the diff, it must have been there from another commit and I thought this was the ZOAU variable not the one you initialized, thanks again Rich.

ddimatos (approved):
Thanks @richp405 , looks good

Issue #835, opened 2023-07-04
[Enhancement] [Epic] [zos_copy] Preserve American Standards Association (ASA) text file control characters


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Ansible module

zos_copy

### Enhancement or feature description

This issue represents all work needing to be completed to ensure ASA support. Some of the items in the list below could be considered no longer needed or cancelled as development and testing progresses.

[ASA control characters](https://www.ibm.com/docs/en/zos/2.5.0?topic=output-using-asa-text-files) are ' ' (blank), '+', '-', and '0 thru '9', and 'A' thru 'C' one-byte field at the beginning of each record for data sets allocated with an `A` as in VBA and FBA.

1. [x] Research ASA, generate, inspect and ensure reproduction such that a test can can be created.
     - #883 
3. [x] Evaluate existing support for ASA in commands and programs.
     - #884 
5. [x] Evaluate `zos_copy` capabilities when interacting with ASA.
     - #885
7. [ ] Propose and deliver solution, for example maybe a new option such as `asa_text`.
     - #886 

Documents:
1) https://en.wikipedia.org/wiki/ASA_carriage_control_characters
2) https://www.ibm.com/docs/en/zos/2.4.0?topic=descriptions-cp-copy-file
3) https://www.ibm.com/docs/en/zos/2.5.0?topic=output-using-asa-text-files
4) https://www.ibm.com/docs/en/zos/2.5.0?topic=sets-control-characters

PR #1028, opened 2023-10-16
[Enhancement] [zos_copy] Add support for ASA control chars

rexemin:
##### SUMMARY
Implements #835. 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
zos_copy

### DIFF
--- a/plugins/module_utils/copy.py
+++ b/plugins/module_utils/copy.py
@@ -22,6 +22,9 @@ from ansible_collections.ibm.ibm_zos_core.plugins.module_utils.ansible_module im
 from ansible_collections.ibm.ibm_zos_core.plugins.module_utils.better_arg_parser import (
     BetterArgParser,
 )
+from ansible_collections.ibm.ibm_zos_core.plugins.module_utils import (
+    data_set
+)
 
 if PY3:
     from shlex import quote
@@ -110,7 +113,7 @@ def copy_ps2uss(src, dest, is_binary=False):
     return rc, out, err
 
 
-def copy_pds2uss(src, dest, is_binary=False):
+def copy_pds2uss(src, dest, is_binary=False, asa_text=False):
     """Copy the whole PDS(E) to a uss path
 
     Arguments:
@@ -119,6 +122,8 @@ def copy_pds2uss(src, dest, is_binary=False):
 
     Keyword Arguments:
         is_binary: {bool} -- Whether the file to be copied contains binary data
+        asa_text: {bool} -- Whether the file to be copied contains ASA control
+            characters
 
     Raises:
         USSCmdExecError: When any exception is raised during the conversion.
@@ -130,12 +135,22 @@ def copy_pds2uss(src, dest, is_binary=False):
     module = AnsibleModuleHelper(argument_spec={})
     src = _validate_data_set_name(src)
     dest = _validate_path(dest)
+
     cp_pds2uss = "cp -U -F rec \"//'{0}'\" {1}".format(src, quote(dest))
-    if is_binary:
+
+    # When dealing with ASA control chars, each record follows a
+    # different format than what '-F rec' means, so we remove it
+    # to allow the system to leave the control chars in the
+    # destination.
+    if asa_text:
+        cp_pds2uss = cp_pds2uss.replace("-F rec", "", 1)
+    elif is_binary:
         cp_pds2uss = cp_pds2uss.replace("rec", "bin", 1)
+
     rc, out, err = module.run_command(cp_pds2uss)
     if rc:
         raise USSCmdExecError(cp_pds2uss, rc, out, err)
+
     return rc, out, err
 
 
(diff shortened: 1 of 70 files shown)


plugins/module_utils/copy.py
```diff
@@ -216,6 +228,87 @@ def copy_vsam_ps(src, dest):
...
+        dest: {str} -- The MVS destination data set or member
+
+    Returns:
+        boolean -- The return code after the copy command executed successfully
+        str -- The stdout after the copy command executed successfully
+        str -- The stderr after the copy command executed successfully
+    """
+    module = AnsibleModuleHelper(argument_spec={})
+
+    oget_cmd = 'tsocmd "OGET \'{0}\' \'{1}\'"'.format(src, dest)
```
ddimatos, line 245: This is not a real issue, back in the beginning we had issues with `tsocmd` , likely from the Rocket Python z port, can't recall the issue. I would like to offline discuss the use of OGET also wondering if you thought there was any value in using module_utils -> mvs_cmd -> ikjeft01(...) to invoke OGET, it drives mvscmd which can also have limitations with temporary data sets, no action here but just wanted to have a discussion.

ddimatos, line 245: Some of the other new copy routines perform a validation on src and dest, this method `copy_asa_uss2mvs` does not perform the below, is it needed for our security scans? 

```
    src = _validate_data_set_name(src)
    dest = _validate_path(dest)
```

plugins/module_utils/copy.py
```diff
@@ -216,6 +228,87 @@ def copy_vsam_ps(src, dest):
...
+        str -- The stderr after the copy command executed successfully
+    """
+    from os import path
+    from zoautil_py import datasets
+
+    module = AnsibleModuleHelper(argument_spec={})
+    src = _validate_data_set_name(src)
+    dest = _validate_path(dest)
+
+    for member in datasets.list_members(src):
```
ddimatos, line 292: When iterating and encountering a non-zero RC and failing, do we leave a partially or any PDSE on the managed z/OS node? I recall we would have some clean up in a module failure but did not follow the code all the way, maybe you know off hand.

plugins/module_utils/data_set.py
```diff
@@ -321,6 +323,10 @@ def allocate_model_data_set(ds_name, model, vol=None):
             alloc_cmd = """{0} -
             VOLUME({1})""".format(alloc_cmd, vol.upper())
 
+        if asa_text:
+            alloc_cmd = """{0} -
+            RECFM(F,B,A)""".format(alloc_cmd)
```
ddimatos, line 328: I thought that when a `dest` does not exists where `zos_copy` will create it, doesn't `zos_copy` also honor the volume option for destinations that are created by `zos_copy` , maybe I am remembering it incorrectly.

plugins/modules/zos_copy.py
```diff
@@ -1793,21 +1870,37 @@ def create_seq_dataset_from_file(
...
+    adjust_record_format = False
 
     # When src is a binary file, the module will use default attributes
     # for the data set, such as a record format of "VB".
     if not is_binary:
         record_format = "FB"
-        record_length = get_file_record_length(file)
+        if not record_length:
+            record_length = get_file_record_length(file)
+            adjust_record_format = True
```
ddimatos, line 1892: Is there any issue with `record_format = "FB"` where ASA needs FBA , I saw other code updates to FBA so maybe those later overwrite this `FB` but wanted to highlight the point.

ddimatos (changes requested):
@rexemin - this looks really nice, its clean, very nice work. I left a few comments, we can talk about some of them offline and resolve. The only other thing I noticed is if  you wanted to save your self a boolean line 1892 I think could have:
```
if asa_text:
        # Adding one byte more to the record length to account for the
        # control character at the start of each line.
        record_length += 1
```

plugins/module_utils/data_set.py
rexemin: It does honor the volume specified by the user. The record format option doesn't overwrite the volume one, it's appending to the string

plugins/module_utils/copy.py
rexemin: We leave an incomplete PDSE in case of failure, we removed the code that would try to clean up because it was giving us more problems than it fixed

plugins/modules/zos_copy.py
rexemin: The record format gets overwritten inside `get_data_set_attributes` if the module is copying ASA text files.

richp405 (approved):
I didn't see anything beyond what Demetri already commented on.

plugins/module_utils/copy.py
fernandofloresg: No is not, executed Appscan and Mend security scans and there are no issues, we currently have no validations for mvs data set paths.

fernandofloresg (approved):
Talked about my questions with Ivan, mainly around support with executable: true for which he added a validation and around record format + 1.

plugins/module_utils/copy.py
ddimatos: I see the complexity that can come with this so I opened a research issue, I don't think it should be further addressed at this point , thank you for the update, here is the issue. https://github.com/ansible-collections/ibm_zos_core/issues/1033

plugins/module_utils/data_set.py
ddimatos: Ah, yup i see the append now, my bad and why I should not do late night reviews, thanks for explaining.

plugins/modules/zos_copy.py
ddimatos: Thanks @rexemin , I should have followed the code but was aiming to get a quick turn around on this PR, I do see it updated there.

ddimatos:
As of 10/18 1:45 PST all but one comment are now resolved an I know Ivan is looking at the [comment](https://github.com/ansible-collections/ibm_zos_core/pull/1028/files#r1363311464) and exploring the use of `ikjeft01`. Thank you.

plugins/module_utils/copy.py
rexemin: Could change `tsocmd` for `mvscmd` without too much problem

ddimatos: Looks good to me, pending pipeline testing I approve this and thank you for the change.

ddimatos (approved):
Looks great, as I said before @rexemin , clean and well thought out code.

rexemin:
Screenshot 2023-10-23 at 19 56 52
Screenshot 2023-10-23 at 19 58 41

PR #1036, opened 2023-10-24
[1.9.0] zos_apf remove try expect pass to better exception handling

### LINKED ISSUE #874: [Enabler] Static Scan issue - linked to '#25'

### Ansible module

_No response_

### Enabler description

[Enabler] Static Scan issue - linked to '#25'

fernandofloresg:
##### SUMMARY

Removed Try Except Pass in favor of a better exception handling.

This does not change any behavior but rather removes a bad practice that can be used as an exploitation.

Fixes #874 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_apf


fernandofloresg:
Still to decide if we should change this one to ready for review since APF tests are failing due to an underlying ZOAU known issue not yet addressed, but team is informed of it.

fernandofloresg:
Screenshot 2023-11-06 at 9 16 40 PM


Screenshot 2023-11-06 at 9 16 59 PM

richp405 (approved):
Looks okay from here.  I'm surprised by the nested 'try' statements.

plugins/modules/zos_apf.py
```diff
@@ -520,29 +520,27 @@ def main():
...
-
-    result['stdout'] = operOut
-
+            except json.JSONDecodeError:
+                module.exit_json(**result)
+            for d in data[2:]:
+                ds = d.get('ds')
+                vol = d.get('vol')
+                try:
+                    if (re.match(library, ds)) or (re.match(volume, vol)) or (sms and sms == vol):
```
ketankelkar, line 540: i'm questioning if the effect of the logic is the same as from the other (maybe it's handled elsewhere and i just missed it), 
the old:
```if (dsRx and dsRx.match(ds)) or (volRx and volRx.match(vol)) or (sms and sms == vol):```
vs the new:
```if (re.match(library, ds)) or (re.match(volume, vol)) or (sms and sms == vol):```

just a few snippets of python:
```python
>>> import re
>>> bool(re.match("", "SOME.DATA.SET"))
True
```
```python
>>> bool("")
False
```
```python3
>>> bool("" and re.match("", "SOME.DATA.SET"))
False 
```

My concern is that there may be some cases where a regex match is reported but an empty string was matched rather than a real value. To me, the old ensures that there's a match with _real_ data by having the two part check per conditional clause `dsRX and dsRX.match(ds)`. I think there could be something similar added to the new, maybe something like `(if library and re.match(library, ds)) or ...`.

changelogs/fragments/1036-apf-try-except.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - zos_apf - Modified Try Except Pass block when operation = list so that now the module
```
ketankelkar, line 2: `when operation = list` can become `when operation is a list`

ketankelkar: present tense is better for minor_changes. 

- zos_apf - Modifies the try except pass block when operation is a list so that now the module...

changelogs/fragments/1036-apf-try-except.yml
```diff
@@ -0,0 +1,4 @@
...
+    handles expected exceptions and performs and specific action.
```
ketankelkar, line 3: think there may be a typo here?
>...expected exceptions *and* performs *and* specific action.

ketankelkar (changes requested):
potentially missed a corner case, some nit-picking for the docfrag (extra scrutiny since this becomes our changelog and goes out to customers)

changelogs/fragments/1036-apf-try-except.yml
fernandofloresg: agree, I always post the ansible developtment guide and refer to this but this time I forgot 😅

changelogs/fragments/1036-apf-try-except.yml
fernandofloresg: yup

plugins/modules/zos_apf.py
fernandofloresg: Nice finding, this is an issue bc if `library == ""`, then resulting match will be 
```
>>> bool(re.match("", "SOME.DATA.SET"))
True
```

Will modify this to honor the current behavior.

fernandofloresg:
Screenshot 2023-11-10 at 2 47 15 PM

fernandofloresg:
Screenshot 2023-11-10 at 2 47 33 PM

ketankelkar (approved):
LGTM, i'd suggest switching `...when operation = list` to `when operations is a list..` in the changelog, but i leave that up to you, it's alright the way it is to

fernandofloresg:
> LGTM, i'd suggest switching `...when operation = list` to `when operations is a list..` in the changelog, but i leave that up to you, it's alright the way it is to

Agree as you said this will be in the release notes and will be read by users :)

Issue #837, opened 2023-07-05
[Bug] Encapsulating MissingZOAUImport in modules may be trapping valuable information


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Are the dependencies a supported version?

- [X] The dependencies are supported.

### IBM Z Open Automation Utilities

v1.2.3

### IBM Enterprise Python

v3.11.x

### IBM z/OS Ansible core Version

v1.5.0

### ansible-version

v2.15.x

### z/OS version

v2.5

### Bug description

For reasons unknown at this time, the modules trap the import error and set it to the instance instead of raising it. For example:

```
try:
    from zoautil_py import datasets
except ImportError:
    datasets = MissingZOAUImport()
``` 
instead of:
```
try:
    from zoautil_py import datasets
except ImportError:
   raise MissingZOAUImport() 
````

It was reported that the traceback only showed:
```
TypeError: Value of unknown type: <class \'TypeError\'>, \'MissingZOAUImport\' object is not callable
```
Which is not clear at all. 

When running the raw Ansible ssh -C command you can see information was kept from propagating:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/u/ibmuser/.local/lib/python3.9/site-packages/zoautil_py/jobs.py", line 13, in <module>
    from zoautil_py.core import call_zoau_library
ImportError: CEE3512S An HFS load of module /u/ibmuser/.local/lib/python3.9/site-packages/zoautil_py/core.abi3.so failed. The system return code was 0000000111; the reason code was EF076015.            
Shared connection to hursvl01.hursley.ibm.com closed.
```

The first issue is to evaluate the modules use of MissingZOAUImport () and determine why we can not raise it instead.

The second consideration is perform some added evaluation in the event of this failure and inspect the environment variables, zoau version, python path, pip install (what ever makes sense that can help the user) when this error occurs. 

 Conversation: https://i<>.slack.com/archives/CSLEJ8VGV/p1688498896899089                                              

### Playbook verbosity output.

```
<hursvl01.hursley.ibm.com> ESTABLISH SSH CONNECTION FOR USER: ibmuser
[...]

PR #1042, opened 2023-11-01
[v1.9.0] Bugfix/837/missing ZOAU imports

rexemin:
##### SUMMARY
Proposal for a solution for #837.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
- zos_operator
- zos_job_query
- module_utils/job.py

##### ADDITIONAL INFORMATION
To bubble up import errors from ZOAU, we'll still use the idea behind `MissingZOAUImport` but change around a few things to avoid losing important context when an import fails. For now it's just for zos_operator and zos_job_query/job.py, to showcase the two types of changes we'll need to do:
- Changes in module_utils where we always import at a module level instead of individual functions, and making sure modules use `to_text` to properly render the exception in the return JSON.
- Changes in modules where we do a check early on to catch import failures.

### DIFF
--- a/plugins/module_utils/import_handler.py
+++ b/plugins/module_utils/import_handler.py
@@ -41,7 +41,7 @@ class ZOAUImportError(object):
     and the moment ANY method gets called, we finally raise an exception.
     """
 
-    def __init__(self, exception_traceback: str):
+    def __init__(self, exception_traceback):
         """When creating a new instance of this class, we save the traceback
         from the original exception so that users have more context when their
         task/code fails. The expected traceback is a string representation of
--- a/plugins/module_utils/import_handler.py
+++ b/plugins/module_utils/import_handler.py
@@ -27,6 +27,48 @@ class MissingZOAUImport(object):
         return method
 
 
+class ZOAUImportError(object):
+    """This class serves as a wrapper for any kind of error when importing
+    ZOAU. Since ZOAU is used by both modules and module_utils, we need a way
+    to alert the user when they're trying to use a function that couldn't be
+    imported properly. If we only had to deal with this in modules, we could
+    just validate that imports worked at the start of their main functions,
+    but on utils, we don't have an entry point where we can validate this.
+    Just raising an exception when trying the import would be better, but that
+    introduces a failure on Ansible sanity tests, so we can't do it.
+
+    Instead, we'll replace what would've been a ZOAU library with this class,
+    and the moment ANY method gets called, we finally raise an exception.
+    """
+
+    def __init__(self, exception_traceback: str):
+        """When creating a new instance of this class, we save the traceback
+        from the original exception so that users have more context when their
+        task/code fails. The expected traceback is a string representation of
+        it, not an actual traceback object. By importing `traceback` from the
+        standard library and calling `traceback.format_exc()` we can
+        get this string.
+        """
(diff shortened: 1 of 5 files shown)


richp405 (approved):
Looks straightforward.  Nice start.

plugins/module_utils/import_handler.py
```diff
@@ -27,6 +27,48 @@ def method(*args, **kwargs):
...
+        standard library and calling `traceback.format_exc()` we can
+        get this string.
+        """
+        self.traceback = exception_traceback
+
+    def __getattr__(self, name):
+        """This code is virtually the same from `MissingZOAUImport`. What we
+        do here is hijack all calls to any method from a missing ZOAU library
+        and instead return a method that will alert the user that there was
+        an error while importing ZOAU.
```
fernandofloresg, line 58: I never stopped to read MissingZOAUImport, but reading through the original issue and your comment I assume just using import and then raise exception breaks the ansible sanity tests?

fernandofloresg (approved):
Just had one question at the time, I do agree with this way of doing things bc we will get better error messages but do feel the same as Demetri said, it's a shame that we need so much effort for so little return, but I guess if is a one time cleanup then is ok.

plugins/module_utils/import_handler.py
rexemin: Yes, sadly the most logical approach breaks sanity tests. Running Ansible's sanity tests with just raising an exception the moment we get an error importing ZOAU gives me this:

```
Running sanity test "import" on Python 3.10
ERROR: Found 4 import issue(s) on python 3.10 which need to be resolved:
ERROR: plugins/module_utils/job.py:36:0: traceback: ModuleNotFoundError: No module named 'zoautil_py'
ERROR: plugins/modules/zos_job_output.py:417:0: traceback: ModuleNotFoundError: No module named 'zoautil_py' (at plugins/module_utils/job.py:36:0)
ERROR: plugins/modules/zos_job_query.py:258:0: traceback: ModuleNotFoundError: No module named 'zoautil_py' (at plugins/module_utils/job.py:36:0)
ERROR: plugins/modules/zos_job_submit.py:610:0: traceback: ModuleNotFoundError: No module named 'zoautil_py' (at plugins/module_utils/job.py:36:0)
```

And this (the other 3 errors are about ignore files and one about doing things unidiomatically, so unrelated to your comment right now): 

```
Running sanity test "validate-modules"
ERROR: Found 6 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/zos_job_output.py:0:0: import-error: Exception attempting to import module for argument_spec introspection, 'No module named 'zoautil_py''
ERROR: plugins/modules/zos_job_query.py:0:0: import-error: Exception attempting to import module for argument_spec introspection, 'No module named 'zoautil_py''
ERROR: plugins/modules/zos_job_submit.py:0:0: import-error: Exception attempting to import module for argument_spec introspection, 'No module named 'zoautil_py''
```

ddimatos (approved):
This is creative solution to an unfortunate issue we have to manage in that we are forced into this because of the certification sanity tests not being able to find zoautil_py.

This is a reasonable solution and for now, likely the only one. If zoautil_py ever ends up in PYPI, we might have other options.

PR #1044, opened 2023-11-02
1043 bug title zos operator is passing wrong value to zoauopercmd

### LINKED ISSUE #1043: [Bug] [zos_operator] Module is passing wrong wait time property to zoau opercmd

v1.2.5

### IBM Enterprise Python

v3.9.x

v1.8.0-beta.1

### ansible-version

v2.14.x

### z/OS version

v2.5

### Ansible module

zos_operator, zos_operator_action_query

### Bug description

User reported inconsistency in wait-delay, which was recently addressed.  

When calling opercmd.exec, zos_operator passes the 'wait' boolean as kwarg 'wait_arg'.  This happened from a mis-reading of the opercmd source code.  It should be passing 'wait', not 'wait_arg'.  Originally diagnosed by James McKee.

I've added zos_operator_action_query to double-check, since it was changed at the same time.

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

richp405:
##### SUMMARY
zos_operator functions were passing wait_arg boolean instead of wait.

Fixes #1043 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_operator and zos_operator_action_query

##### ADDITIONAL INFORMATION
We changed to blocking behavior when times are provide, but didn't pass the behavior to zoau.


richp405:
image

ddimatos:
@richp405 or any of the team. I spent a bit of time evaluating how this bug got past our regression, I think I can summarize and hopefully who ever works on an updated test case it will help them.

When we introduced support for ZOAU option `wait` (zoau version 1.2.5) which is really a blocking call, meaning it will block on the time set for `wait_time_s`, we overlooked creating a test case for it specifically. 

While we do have this time boxed test case (below), what this does is just validate that our default 1 second value is not used by setting `wait_time_s=10` and then we ensure that the elapsed time is 2 times less than wait_time_s, its difficult to test the older behavior (zoau < 1.2.5) because ideally we would want a command to run a known amount of time , say 5 seconds which then we would have additional checks like  `assert result.get('elapsed') >= wait_time_s`, thus I don't think we can do much to improve this test unless we create our own operator command. Thus no change for this test but I think we can use this test to model one for zoau 1.2.5 or later. 
```
def test_zos_operator_positive_verbose_with_quick_delay(ansible_zos_module):
    hosts = ansible_zos_module
    wait_time_s=10
    results = hosts.all.zos_operator(
        cmd="d u,all", verbose=True, wait_time_s=wait_time_s
    )

        assert result["rc"] == 0
        assert result.get("content") is not None
        # Account for slower network
        assert result.get('elapsed') <= (2 * wait_time_s)
```

To address zoau 1.2.5 or later, we need focus on the fact it is a blocking call, so if you set `wait_time_s` = `5` seconds, the elapsed time should be minimally 5 seconds, its going to be a bit more because elapsed time takes into account some other costs. Thus copying the above test and asserting `assert result.get('elapsed') >= wait_time_s` and `assert result.get('elapsed') <= (2 * wait_time_s)` should help test the zoau 1.2.5 block call use of `wait`. The only thing is now your test case has to be cognizant of the zoau version, assided modifying our test fixture to return this value, which would be a new work item and out of scope for this issue, you could use the raw/command/shell module or even zos_script to detect the zoauversion and logically act on this test case. I would say if the zoau version is less than 1.2.5 you make the test be a noop and return true, else you run through the test code and validate. 

Hope this helps.

rexemin:
So updating the test is going to be left for another PR in the future?

richp405:
> So updating the test is going to be left for another PR in the future?

No... I put it in draft mode so I could have a chance to make that change.

richp405:
With the new test... just passed pipeline:

image

tests/functional/modules/test_zos_operator_func.py
```diff
@@ -119,13 +120,31 @@ def test_zos_operator_positive_verbose_with_quick_delay(ansible_zos_module):
...
+            assert result.get("content") is not None
+            # Account for slower network
+            assert result.get('elapsed') >= wait_time_s
+
+
+
 def test_response_come_back_complete(ansible_zos_module):
     hosts = ansible_zos_module
-    results = hosts.all.zos_operator(cmd="\$dspl")
+    results = hosts.all.zos_operator(cmd="\\$dspl")
```
fernandofloresg, line 142: Why is this backslash added?

tests/functional/modules/test_zos_operator_func.py
```diff
@@ -119,13 +120,31 @@ def test_zos_operator_positive_verbose_with_quick_delay(ansible_zos_module):
...
+        results = hosts.all.zos_operator(
+            cmd="d u,all", verbose=True, wait_time_s=wait_time_s
+        )
+
+        for result in results.contacted.values():
+            assert result["rc"] == 0
+            assert result.get("changed") is True
+            assert result.get("content") is not None
+            # Account for slower network
+            assert result.get('elapsed') >= wait_time_s
```
fernandofloresg, line 136: Why was this assertion `assert result.get('elapsed') <= (2 * wait_time_s)` proposed by Demetri excluded ?

changelogs/fragments/1043-bug-title-zos_operator-is-passing-wrong-value-to-zoauopercmd.yml
```diff
@@ -0,0 +1,7 @@
+bugfixes:
+  - zos_operator - Corrected variable passed from wait_arg to wait in kwargs to zoau.
```
fernandofloresg, line 2: Will ask to stick to this style when writing bugfixes changelogs "Use past tense to describe the problem and present tense to describe the fix." from (https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment).

We could say in the past tense that the module ignored the wait arg and fix now honors the wait time as implemented by zoau (?).

fernandofloresg (changes requested):
Just a small changelog fragment rewrite request and couple of questions, no need to run the test suite again.

tests/functional/modules/test_zos_operator_func.py
richp405: With 1 backslash, it was being flagged by a linter (don't remember which one).

tests/functional/modules/test_zos_operator_func.py
richp405: Demetri was proposing a new test, to make sure elapsed >< wait_time_s, which you show here.  He was referring to an existing test, with < 2 * wait, which was already there.

changelogs/fragments/1043-bug-title-zos_operator-is-passing-wrong-value-to-zoauopercmd.yml
richp405: I keep forgetting tense from the style guide.  Thanks for seeing that.  I'm pushing a fix shortly.

tests/functional/modules/test_zos_operator_func.py
richp405: Here's the code he was referencing, above that change:

def test_zos_operator_positive_verbose_with_quick_delay(ansible_zos_module):
    hosts = ansible_zos_module
    wait_time_s=10
    #startmod = time.time()
    results = hosts.all.zos_operator(
        cmd="d u,all", verbose=True, wait_time_s=wait_time_s
    )
    # endmod = time.time()
    # timediff = endmod - startmod
    # assert timediff < 15

        assert result["rc"] == 0
@@ -119,13 +120,31 @@ def test_zos_operator_positive_verbose_with_quick_delay(ansible_zos_module):
        assert result.get('elapsed') <= (2 * wait_time_s)

changelogs/fragments/1043-bug-title-zos_operator-is-passing-wrong-value-to-zoauopercmd.yml
fernandofloresg: Hey don't worry I do too, is always easier to remember when doing a code review

tests/functional/modules/test_zos_operator_func.py
fernandofloresg: I see, I ran the same command in my local and works fine, was worried it would change something 👍

fernandofloresg (approved):
This looks good to me now Thanks Rich

changelogs/fragments/1043-bug-title-zos_operator-is-passing-wrong-value-to-zoauopercmd.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - zos_operator - The module was ignoring the wait time argument. The module now passes the wait time argument to zoau.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1044).
+
+  - zos_operator_action_query - The module was ignoring the wait time argument. The module now passes the wait time argument to zoau.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1044).
```
rexemin, line 6: Everything looks good, I would just say that it would look better to write `ZOAU` in all caps.

richp405: I agree.

Issue #990, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 2


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

2)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:8: sanity[cannot-ignore]: Ignore file contains validate-modules:parameter-type-not-in-doc at line 8, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_copy.py validate-modules:parameter-type-not-in-doc # Passing args from action plugin

ddimatos:
After review, this issue was found to be unresolved and now being reopened for a later date. I will remove the reference from the PR 1048 as well.

Issue #785, opened 2023-05-31
[Enabler] Update pipeline to include Ansible 2.15


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

`ansible-core` 2.15 has GA'd, our pipeline was using 2.15 from devel as pipeline option latest, likely now that is pointing to 2.16, we have a gap with no option for 2.15 , the pipeline needs an update to use 2.15 so we can certify and GA on it by June.

richp405:
Have a script including 2.15.0, but it is generating 'module not found' error.  I will track this down on Tuesday.

richp405:
been dealing with an error that ONLY shows up starting in Ansible 2.15... it appears that older behavior was finally locked out in March 2023 (2.15.0 release).

Error is: 
    from ansible_collections.ibm.ibm_zos_core.plugins.module_utils.data_set import (
E   ModuleNotFoundError: No module named 'ansible_collections'

I found adding pythonpath back to the environment has moved the error, so I think I'm close to getting this resolved.

All work so far is on Jenkinsfile5b, so it doesn't interfere with normal builds (which use Jenkinsfile5)

ddimatos:
Will move this to Q3:sprint 6 so we can clean up and prepare for next quarter

richp405:
Created -Experimental link, connected to Jenkinsfile5d in rep.  Showed it to Marcel today

PR #1048, opened 2023-11-03
Update sanity test ignore

### LINKED ISSUE #989: [Enabler] Review and update sanity test ignore entry issue 1

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

1)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:6: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-type-does-not-match-spec at line 6, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_copy.py validate-modules:doc-type-does-not-match-spec # doc type should be str, while spec type is path to allow user path expansion

AndreMarcel99:
##### SUMMARY
Change the logic of receiving data information and eliminate unnecessary inheritance of data that is not in use.

Fixes #989 #990 #991 and validate #785 

##### ISSUE TYPE
- Docs Pull Request

##### ADDITIONAL INFORMATION
To the arguments of the module
Change path to str, our function already works clear with string as path in the src argument.
Add arguments owner and group to args besides delete copy member and set to a validation.
And delete the inheritance arguments of document module.

[Pipeline with only run of zos_copy](https://billies1.fyre.ibm.com/job/zosAnsible-Experimental/218/)
Captura de pantalla 2023-11-02 a la(s) 6 09 42 p m
Captura de pantalla 2023-11-02 a la(s) 6 10 02 p m


richp405 (approved):
Nice.

tests/sanity/ignore-2.15.txt
fernandofloresg, line 1: I see the PR refers to 3 different sanity issues but in this file only two are removed, why is not a third one removed ?

plugins/modules/zos_copy.py
```diff
@@ -3053,7 +3057,7 @@ def run_module(module, arg_def):
 def main():
     module = AnsibleModule(
         argument_spec=dict(
-            src=dict(type='path'),
+            src=dict(type='str'),
```
fernandofloresg, line 3060: At first sight this change might look dangerous, if we change the src type to `str` from `path` then validations performed by BetterArgParser will not work? No, after checking the code `argument_spec` is not validated by BetterArgParser but rather `arg_def` which will keep the `type='data_set_or_path'` definition, so validation on `src` will continue to be performed by BetterArgParser, I think we don't have a test case for it but neither we need it.

From what I see this change looks ok.

plugins/modules/zos_copy.py
```diff
@@ -2576,6 +2578,8 @@ def run_module(module, arg_def):
         if volume:
             dest_data_set["volumes"] = [volume]
 
+    copy_member = is_member(dest)
```
fernandofloresg, line 2581: Adding the member validation into the zos_copy module will not make a big impact since is just a regex validation.

tests/sanity/ignore-2.15.txt
fernandofloresg: Asking just to make sure we don't close issues that are incomplete, even if we repair part of it, which is good, we still want to be aware of which issues were not possible to fix and the reason, that will give us more leverage with Red Hat in case we definitely need to keep some ignore files. Please make sure to document those cases properly

fernandofloresg (changes requested):

tests/sanity/ignore-2.15.txt
rexemin: I went through the 3 issues linked and they are for:
- `validate-modules:undocumented-parameter`
- `validate-modules:parameter-type-not-in-doc`
- `validate-modules:doc-type-does-not-match-spec`

I see `undocumented-parameter` and `parameter-type-not-in-doc` are still there in the ignore file, so I'm wondering if @AndreMarcel99 kept those ones in because there were args that were too costly to compute in a remote host. Besides that, this PR also removes `doc-default-does-not-match-spec`, which is appropiate, I see the changes to the `arg_spec`, but I don't recall right now if there's another issue we could link to this PR to better reflect that.

plugins/modules/zos_copy.py
```diff
@@ -114,7 +114,7 @@
         be deleted and recreated following the process outlined in the C(volume) option.
       - When the C(dest) is an existing VSAM (RRDS), then the source must be an RRDS.
         The VSAM (RRDS) will be deleted and recreated following the process outlined
-        in the C(volume) option.
+       in the C(volume) option.
```
rexemin, line 117: Have you tried generating docs after this change? Dunno if a single space would actually break something, but you never know with YAML

tests/sanity/ignore-2.15.txt
AndreMarcel99: Thanks @rexemin  
What happen is that

- validate-modules:parameter-type-not-in-doc
- validate-modules:doc-type-does-not-match-spec

Are just a little bit clean up by deleting values not in use from our parameters, but can be removal at all because of many of the workflow pass in the action_plugin where all is done in the controller, if we put all on the code is overthink that parameters.
An curious by change the issue 6 the 5 also is resolve.

plugins/modules/zos_copy.py
AndreMarcel99: In new modules are work properly with str as type, because ansible is expecting primitives data types

plugins/modules/zos_copy.py
AndreMarcel99: And is a validation over the dest, that doesn't need to be done on controller.

tests/sanity/ignore-2.15.txt
fernandofloresg: @AndreMarcel99  Can you provide some examples of this ?  And what do you mean by "is overthink that parameters"?

plugins/modules/zos_copy.py
AndreMarcel99: No change but it was part of the issues of sanity test that space

tests/sanity/ignore-2.15.txt
AndreMarcel99: [Here](url) case like src_member, is_mvs_dest and is_uss are use for the logics nexts conditionals and case [like](https://github.com/ansible-collections/ibm_zos_core/blob/b30f892a8b6031369985df5ff348879cf6c2a1ed/plugins/action/zos_copy.py#L134) is_uss [and](https://github.com/ansible-collections/ibm_zos_core/blob/b30f892a8b6031369985df5ff348879cf6c2a1ed/plugins/action/zos_copy.py#L105) src_member are with the information of the controller and would recreate a restructure of the code to the module not action plugin.

rexemin (approved):
Looks good, thanks for having the patience to answer all my questions for this PR!

fernandofloresg (approved):

AndreMarcel99:
I put the label Do not merge because if I do not remember well with one of the change one sanity issue for 2.14 will come up

AndreMarcel99:
Captura de pantalla 2023-11-08 a la(s) 1 07 53 p m
I will update this and also check for previous versions

AndreMarcel99:
Ready to be merged, 2.14 run fine 
Captura de pantalla 2023-11-08 a la(s) 2 28 44 p m

fernandofloresg:
Thanks @AndreMarcel99. I also ran a scan with bandit and Appscan and no new security issues were found. This is ready to to merge.

plugins/modules/zos_copy.py
fernandofloresg: This one is weird, I did run the html generation and looks fine. What was the sanity issue raised due to this space?

ddimatos:
I am removing issues 990 and 991 from this PR because after evaluating I found the issues still existed.

PR #1049, opened 2023-11-03
Clean temporary data sets created during XMIT unarchive operation

### LINKED ISSUE #1047: [Bug] [zos_unarchive] leaves temporary data sets after failure when using xmit format and failing during RECEIVE command

v1.2.5

### IBM Enterprise Python

v3.12.x

v1.8.0-beta.1

### ansible-version

latest

### z/OS version

v2.5

### Ansible module

_No response_

### Bug description

When executing XMIT and if the RECEIVE command step fails environment will not be cleaned of any temporary data sets created when copying to remote or unpacking from a ADRDSSU step.

This [line](https://github.com/ansible-collections/ibm_zos_core/blob/b30f892a8b6031369985df5ff348879cf6c2a1ed/plugins/modules/zos_unarchive.py#L856) present on Terse Unarchive avoids that problem by cleaning the env after a failure.

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

There is a method that cleans the environment when unpack action fails on remote with mvs commands, currently when remote_src is False temporary data set was not removed, changed that behavior to ensure it happens when mvs formats unpack fails.

Since this is a cleanup of temporary data set I thought about creating test cases, but it can fail in multiple steps, I do have a playbook which you can use to test the behavior but let me know If you think is necessary to create a test case for it.

Fixes #1047 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_unarchive

##### ADDITIONAL INFORMATION

```paste below
- hosts: zvm
  collections : 
    - ibm.ibm_zos_core
  vars:
    ZOAU: "/zoau/v1.2.2"
    DATASET_NAME: "OMVSADM.ARCHIVE.TEST"
    PYZ: "/allpython/3.9/usr/lpp/IBM/cyp/v3r9/pyz"
  gather_facts: False
  tasks:
    - name: Remove SEQ dataset
      zos_data_set:
        name: "{{ DATASET_NAME }}"
        state: absent
        type: seq
      register: result

    - name: Remove SEQ dataset
      zos_data_set:
        name: "OMVSADM.ARCHIVE.RESULT.XMIT"
        state: absent
        type: seq
      register: result
    
    - name: Create SEQ dataset
      zos_data_set:
        name: "{{ DATASET_NAME }}"
        state: present
        type: seq
      register: result
    
    - name: Add text into the dataset
      ansible.builtin.shell:
        cmd: decho 'this is a test line' "{{ DATASET_NAME }}"
      register: result
    
    - name: Compress data set into trs
      zos_archive:
        src: "{{ DATASET_NAME }}"
        dest: "OMVSADM.ARCHIVE.RESULT.XMIT"
        format:
          name: xmit
        force: True
      register: result

    - name: fetch the archive into local
      zos_fetch:
        src: "OMVSADM.ARCHIVE.RESULT.XMIT"
        dest: ./test_files/
        # flat: true
        is_binary: true
      register: result

    - name: print fetch
      debug:
        msg: "{{ result }}"

    - name: Remove SEQ dataset
      zos_data_set:
        name: "{{ DATASET_NAME }}"
        state: absent
        type: seq

    - name: Unarchive the data set in second vm
      zos_unarchive:
        src: ./test_files/zvm/OMVSADM.ARCHIVE.RESULT.XMIT
        format:
          name: terse
        force: True
        remote_src: False
        tmp_hlq: FAIL
        dest_data_set:
          type: SEQ
          record_format: FB
          record_length: 80
      register: result

    - name: print unarchive result
      debug:
        msg: "{{ result }}"


```

After running this data set, running a `dls -l FAIL.*` no temporary data set should be left.

### DIFF
--- a/changelogs/fragments/1045-local-uss-unarchive.yml
+++ b/changelogs/fragments/1045-local-uss-unarchive.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - zos_unarchive - Using a local file with a USS format option failed when sending to
+    remote because dest_data_set option had an empty dictionary. Fix now leaves dest_data_set
+    as None when using a USS format option.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1045).
--- a/changelogs/fragments/1047-xmit-temporary-residual.yml
+++ b/changelogs/fragments/1047-xmit-temporary-residual.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - zos_unarchive - When zos_unarchive fails during RECEIVE command it does not clean the
+    temporary data sets created. Fix now removes the temporary data sets.
+    ()
--- a/changelogs/fragments/1047-xmit-temporary-residual.yml
+++ b/changelogs/fragments/1047-xmit-temporary-residual.yml
@@ -1,4 +1,0 @@
-bugfixes:
-  - zos_unarchive - When zos_unarchive fails during RECEIVE command it does not clean the
-    temporary data sets created. Fix now removes the temporary data sets.
-    ()
--- a/changelogs/fragments/1049-xmit-temporary-data-sets.yml
+++ b/changelogs/fragments/1049-xmit-temporary-data-sets.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - zos_unarchive - When zos_unarchive fails during RECEIVE command it does not clean the
+  - zos_unarchive - When zos_unarchive fails during unpack either with xmit or terse it does not clean the
     temporary data sets created. Fix now removes the temporary data sets.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1049).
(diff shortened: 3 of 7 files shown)


rexemin (approved):
Looks good, thanks for working on this so quickly. I tried your playbook and could verify that temp data sets get cleaned up now. I skimmed through our test suite and didn't really see that we test often for clean up, so I'm okay if you leave the PR as is, unless somebody else thinks to the contrary.

fernandofloresg:
Thanks for reviewing @rexemin , I'm not sure about creating tests but I have a thought in favor of creating them, @ddimatos mentioned that ensuring that our collection does not leave temporary files is something asked by users, at least for these I have an idea, one test case that fails in multiple ways, using `tmphlq = FAIL`  and then use `dls FAIL.*` to check the env afterwards.

And do the same for /tmp/ folder, maybe just take an snapshot and then check back.

rexemin:
So there was someone that thought to the contrary 😬
You have a good point, and I think your idea for this test case is pretty good. If you have the chance then it would be cool if you added the test. After that, we can start looking at expanding coverage to the rest of the modules

fernandofloresg:
@rexemin Don't worry I'm also unsure and I think we will follow Demetri's lead. Our test suite is already pretty large and will wait for other people ideas too. Thanks for sharing yours.

richp405 (approved):
Looks correct from here.

PR #1055, opened 2023-11-09
Removed run command  use of subprocess from encode defaults infavor of using AnsibleModule run command

### LINKED ISSUE #873: [Enabler] Static Scan issue - linked to '#24'

### Ansible module

_No response_

### Enabler description

[Enabler] Static Scan issue - linked to '#24'

### LINKED ISSUE #875: [Enabler] Static Scan issue - linked to '#23'

### Ansible module

_No response_

### Enabler description

[Enabler] Static Scan issue - linked to '#23'

fernandofloresg:
##### SUMMARY

Removed use of subprocess from system utils, since the only use of it could be replaced for AnsibleModule run_command method. Removing the use of subprocess from our workflow avoids us having to explain the need to open a subprocess and rather filter the command through ansible own dedicated method (AnsibleModule.run_command). That way we are reverting back from a change done 3 years ago, when it was decided to use our own implementation of subprocess instead of Ansible's, but looking at our modules, we actually use this AnsibleModule wrapper all around and can't think of a reason of why not using it here too.

Also, to avoid future implementations of system.run_command without a solid reason I commented the method's code.

I used a trivial tag as changelog to avoid displaying this to the final user and the change was pretty simple.

Fixes #873 
Fixes #875 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

module_utils/encode,py (This is used by many modules, all of their test suites were tested.)

### DIFF
--- a/changelogs/fragments/1055-remove-subprocess-encode.yml
+++ b/changelogs/fragments/1055-remove-subprocess-encode.yml
@@ -1,0 +1,4 @@
+trivial:
+  - encode_utils - Removed use of subprocess from system utils, since the only
+    use of it could be replaced for AnsibleModule runcommand method.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1055).
--- a/plugins/module_utils/encode.py
+++ b/plugins/module_utils/encode.py
@@ -62,7 +62,8 @@ class Defaults:
         """
         system_charset = locale.getdefaultlocale()[1]
         if system_charset is None:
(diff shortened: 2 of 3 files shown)


fernandofloresg:
I will share the pipeline regression results to the team privately, since those are stored behind firewall text sharing app.

richp405 (approved):
Nice, clean, simple = approved.

PR #1057, opened 2023-11-10
Bugfix/374/module zos mvs raw errors with long multi line quoted string in content field

### LINKED ISSUE #374: [Bug] Module zos_mvs_raw errors with long multi-line quoted string in content field
##### SUMMARY

When using module ```zos_mvs_raw``` with program ```IDCAMS``` in the parameter ```CONTENT``` inserting a string the quote is interpreted as content and escaped, getting the message ```"IDC3211I KEYWORD '\"' IS IMPROPER"```

##### COMPONENT NAME

```zos_mvs_raw```

##### ANSIBLE VERSION

```
Ansible version output: 
ansible [core 2.13.1]
  config file = None
  configured module search path = ['/Users/pedrotrevino/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/6.0.0/libexec/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/pedrotrevino/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.10.5 (main, Jun 23 2022, 17:14:57) [Clang 13.1.6 (clang-1316.0.21.2.5)]
  jinja version = 3.1.2
  libyaml = True
```

##### SPECIFY ANSIBLE COLLECTION VERSION

```version": "1.4.0-beta.1```

##### SPECIFY THE Z OPEN AUTOMATION UTILITIES VERSION

```zoau version 1.1.1```

##### ENVIRONMENT

```z/OS 2.4```
```IBM Enterprise Python 3.8.2```

##### STEPS TO REPRODUCE

In the definition of clusters, on the CONTENT paramter, using large strings
---
      - name: Define cluster 1
        zos_mvs_raw:
          program_name: idcams
          auth: yes
          dds:
            - dd_output:
                dd_name: sysprint
                return_content:
                  type: text
            - dd_input:
                dd_name: sysin
                content:
                  "  DEFINE CLUSTER(                                -
                                    NAME({{ DFS_AUTH_LIB_HLQ1 }}.{{ DFS_AUTH_LIB_HLQ2 }}.RECON1) -
                                    CYL(10 10)  -
                                    FREESPACE(20 20)                -
                                    INDEXED                         -
                                    KEYS(32 0)                      -
                                    NOERASE                         -
                                    NONSPANNED                      -
                                    NOREUSE                         -
                                    SHAREOPTIONS(3 3)               -
                                    SPEED                           -
                                    UNORDERED                       -
                                    RECORDSIZE(4086 32600)          -
                    {% if DFS_SMS_ENVIRONMENT %}
                                    STORAGECLASS({{DFS_SMS_CLASS}})                 -
                    {% endif %}
                    {% if not DFS_DS_VOLUME1 == None  %}
                                    VOLUMES({{ DFS_DS_VOLUME1 }})                 -
                    {% endif %}
                                    UNIQUE                          -
                                    )"

When you try to execute the playbooks, it will fail on the ```Define clusters``` task with the next message ```"IDC3211I KEYWORD '\"' IS IMPROPER"```

##### EXPECTED RESULTS

Is expected that ```CONTENT``` read the entire block of JCL code without any problem, allowing the definition of the cluster that will lead to the creation of 2 files with extension ```.DATA``` and ```.INDEX```

##### ACTUAL RESULTS

The execution of the playbook continues but without define the clusters, it just returns ```"IDC3211I KEYWORD '\"' IS IMPROPER"``` and code 12

```
Playbook output:
changed: [zosendpoint] => {
    "backups": [],
    "changed": true,
    "dd_names": [
        {
            "byte_count": 1089,
            "content": [
                "\fIDCAMS  SYSTEM SERVICES                                           TIME: 14:49:07        08/15/22     PAGE      1",
                "",
                "        ",
                "\"  DEFINE CLUSTER(                                -",
                "                  NAME(IMSTESTL.IMS1.RECON1) -",
                "                  CYL(10 10)  -",
                "                  FREESPACE(20 20)                -",
                "                  INDEXED                         -",
                "                  KEYS(32 0)                      -",
                "                  NOERASE                         -",
                "                  NONSPANNED                      -",
                "                  NOREUSE                         -",
                "                  SHAREOPTIONS(3 3)               -",
                "                  SPEED                           -",
                "                  UNORDERED                       -",
                "                  RECORDSIZE(4086 32600)          -",
                "                      VOLUMES(USRV01)                 -",
                "                    UNIQUE                          -",
                "                  )\"",
                "",
                "IDC3211I KEYWORD '\"' IS IMPROPER",
                "",
                "IDC3202I ABOVE TEXT BYPASSED UNTIL NEXT COMMAND. CONDITION CODE IS 12",
                "",
                "        ",
                "",
                "IDC0002I IDCAMS PROCESSING COMPLETE. MAXIMUM CONDITION CODE WAS 12",
                ""
            ],
            "dd_name": "sysprint",
            "name": "OMVSADM.P6777661.T0727611.C0000000",
            "record_count": 28
        }
    ],
    "invocation": {
        "module_args": {
            "auth": true,
            "dds": [
                {
                    "dd_concat": null,
                    "dd_data_set": null,
                    "dd_dummy": null,
                    "dd_input": null,
                    "dd_output": {
                        "dd_name": "sysprint",
                        "return_content": {
                            "response_encoding": "iso8859-1",
                            "src_encoding": "ibm-1047",
                            "type": "text"
                        }
                    },
                    "dd_unix": null,
                    "dd_vio": null
                },
                {
                    "dd_concat": null,
                    "dd_data_set": null,
                    "dd_dummy": null,
                    "dd_input": {
                        "content": "\"  DEFINE CLUSTER(                                -\n                  NAME(IMSTESTL.IMS1.RECON1) -\n                  CYL(10 10)  -\n                  FREESPACE(20 20)                -\n                  INDEXED                         -\n                  KEYS(32 0)                      -\n                  NOERASE                         -\n                  NONSPANNED                      -\n                  NOREUSE                         -\n                  SHAREOPTIONS(3 3)               -\n                  SPEED                           -\n                  UNORDERED                       -\n                  RECORDSIZE(4086 32600)          -\n                      VOLUMES(USRV01)                 -\n                    UNIQUE                          -\n                  )\"\n",
                        "dd_name": "sysin",
                        "return_content": null
                    },
                    "dd_output": null,
                    "dd_unix": null,
                    "dd_vio": null
                }
            ],
            "parm": null,
            "program_name": "idcams",
            "verbose": false
        }
    },
    "ret_code": {
        "code": 12
    }
}
Read vars_file 'vars_zvm/ims-dbdc.yml'
```

#### Ansible.cfg

```
'ansible-config dump --only-changed' output:
DEFAULT_FORKS(/Users/pedrotrevino/git/github/playbooks/ansible.cfg) = 25
DEFAULT_MODULE_PATH(env: ANSIBLE_LIBRARY) = ['/Users/pedrotrevino/.ansible/collections/ansible_collections/ibm/ibm_zos_core/plugins/module>
```

#### Inventory

```
Inventory content:
zvm:
  hosts:
    zvm:
      ansible_host: ec33023a.vmec.svl.ibm.com
      ansible_user: omvsadm
      ansible_python_interpreter: /python3/usr/lpp/IBM/cyp/v3r8/pyz/bin/python3
```

#### Vars

```
'group_vars' or 'host_vars' content:
        "system_environment": {
            "LANG": "C",
            "LIBPATH": "/zoau/v1.1.1-ptf1/lib:/python3/usr/lpp/IBM/cyp/v3r8/pyz/bin/python3/lib:/lib:/usr/lib:.",
            "PATH": "/zoau/v1.1.1-ptf1/bin:/python3/usr/lpp/IBM/cyp/v3r8/pyz/bin/python3/bin:/bin:/var/bin:/usr/lpp/java/java180/J8.0_64/bin:/usr/sbin:.",
            "PYTHONPATH": "/zoau/v1.1.1-ptf1/lib",
            "STEPLIB": "IMSBLD.I15RTSMM.COMRESL1",
            "ZOAU_HOME": "/zoau/v1.1.1-ptf1",
            "_BPXK_AUTOCVT": "ON",
            "_CEE_RUNOPTS": "FILETAG(AUTOCVT,AUTOTAG) POSIX(ON)",
            "_TAG_REDIR_ERR": "txt",
            "_TAG_REDIR_IN": "txt",
            "_TAG_REDIR_OUT": "txt"
        }
```

### PR DESCRIPTION
**Pull request by AndreMarcel99**
##### SUMMARY
Add validations and fixes for proper ways of written multiline texts in playbooks for a JCL

Fixes #374 and #665 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
modify_contetns
add_spaces

##### ADDITIONAL INFORMATION
**modify_contents**: 

- If content is a list is just required to be validate with add spaces function an then be join with /n 

- If content is a string add spaces at the beginning if necessary then split into a list for every new line  and pass to add_spaces function

**add_spaces**: with  a list check every line the len if exceed the 78 characters would throw error if not add double spaces at the beginning if is not already to validate for a correct JCL 

***note:** All the changes will affect correct ways of writing multi lines text in yaml as:

```
content: |1
                INCLUDE INLIB
                NAME {{ MEM_TEST }}(R)
content: |
                INCLUDE INLIB
                NAME {{ MEM_TEST }}(R)
content:
                - "INCLUDE INLIB"
                - "NAME {{ MEM_TEST }}(R)"
```
Captura de pantalla 2023-11-10 a la(s) 12 12 42 p m

### CODE REVIEW

File: plugins/modules/zos_mvs_raw.py [hunk 1]
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+def add_space(lines):
+    """Return the array with two spaces at the beggining.
+
+    Args:
+        lines (list): The list with a line of a program.
+
+    Returns:
+        lines: The list in a proper two spaces and the code.
+    """
+    module = AnsibleModule
```
fernandofloresg, line 3130: I had suspicions on why using AnsibleModule here instead of AnsibleModuleHelper, went a head and tested and looks like failure message is not working as intended.
I tested using the following task: 
```
- name: List data sets matching pattern in catalog,
            save output to a sequential data set. If the data set exists,
            then reuse it, if it does not exist, create it. Returns output as text.
      zos_mvs_raw:
        program_name: idcams
        auth: true
        dds:
          - dd_output:
              dd_name: sysprint
              return_content:
                type: text
                src_encoding: ibm-1047
                response_encoding: iso8859-1
          - dd_input:
              dd_name: sysin
              content: "  DEFINE CLUSTER -
                            (NAME(IMSTESTL.IMS1.RECON4) -
                            CYL(10 10) -
                            FREESPACE(20 20) -
                            INDEXED -
                            KEYS(32 0) -
                            NOERASE -
                            NONSPANNED -
                            NOREUSE -
                            SHAREOPTIONS(3 3) -
                            SPEED -
                            UNORDERED -
                            RECORDSIZE(4086 32600) -
                            VOLUMES(222222) -
                            UNIQUE)"
      register: result

```
and this was the result : 

```
fatal: [zvm]: FAILED! => {"backups": [], "changed": false, "dd_names": [], "msg": "TypeError(\"AnsibleModule.fail_json() missing 1 required positional argument: 'self'\")", "ret_code": {"code": 8}}
```

fernandofloresg: As you can see is throwing a TypeError that is not related to the root cause and ignores your error message

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+    for line in lines:
+        if len(line) == 0:
+            pass
+        else:
+            if line[0] != " " or line[0] != "-":
+                if len(line) > 78:
+                  module.fail_json(msg="Length of the line {0} over 80, dataset can not be written".format(line))
```
fernandofloresg, line 3137: This is a good start for error messages, I don't know if I mentioned this over the slack thread, but this is an opportunity to tell users why this is failing and how they can fix it. Tell that expected JCL maximum length is 78 for this module and give the advice to use either - or + in addition to splitting the JCL content when using it as an input for the module.

changelogs/fragments/1057-module_zos_mvs_raw_errors_with_long_multi_line_quoted_string_in_content_field.yml
fernandofloresg, line 1: This is not a valid format for a changelog I executed `antsibull-changelog lint` and got:

`changelogs/fragments/1057-module_zos_mvs_raw_errors_with_long_multi_line_quoted_string_in_content_field.yml:0:0: section "bugfixes" must be type list not dict`

changelogs/fragments/1057-module_zos_mvs_raw_errors_with_long_multi_line_quoted_string_in_content_field.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  zos_mvs_raw:
+  - module wasn't able to wrote properly a JCL with multi-line | or |1-9.
+    Change now validate the length of lines and allow multi-line declaration with | or |1-9.
```
fernandofloresg, line 4: I think this description needs to mention that the actual bug truncated the first one or two characters writing an syntactically incorrect JCL and now two empty spaces are added into the content because some programs expect to have the first or first two columns as empty spaces.

plugins/modules/zos_mvs_raw.py
```diff
@@ -2454,7 +2461,7 @@ def __init__(
         disposition_abnormal=None,
         block_size=None,
         directory_blocks=None,
-        record_format=None,
+        record_format='FB',
```
fernandofloresg, line 2464: Why is this change needed? RawDatasetDefinition is used to create a dd_data_set object but the issue is actually dealing with the contents of dd_input. If this change is needed then we also need to change the type in the docstring.

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+
+    Returns:
+        lines: The list in a proper two spaces and the code.
+    """
+    module = AnsibleModule
+    for line in lines:
+        if len(line) == 0:
+            pass
+        else:
+            if line[0] != " " or line[0] != "-":
```
fernandofloresg, line 3135: Question, why char "-"  is also accepted at the start of the line? Reading an exctract from [here](https://www.ibm.com/docs/en/abo/2.1?topic=modules-line-continuation), "which indicates the next line should be read as if it were a part of the previous line. " tells me is used as a way to end the current line and start a new one, not at the start of a line, I could be wrong though

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
         return stdout
 
 
+def modify_contents(contents):
```
fernandofloresg, line 3100: Something in the workflow of this method is weird, so if we have a list, we add the char '\n' in each line and then pass it to `add_space` function but if is not a list, we split by the '\n' and then convert it into a list (this ends up being the same case as above right?) and then pass it to `add_space`, so you can either pass an str or a list into `add_space` but the conversion to the other type is performed anyways? 

I think that can be reduced to something like this?
```
if not isinstance(contents, list):
    if contents[0] != " " or contents[0] != "-":
        contents = "  {0}".format(contents)
    contents = list(contents.split("\n"))
contents = add_space(contents)
contents = "\n".join(contents)
return contents
```

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+    module = AnsibleModule
+    for line in lines:
+        if len(line) == 0:
+            pass
+        else:
+            if line[0] != " " or line[0] != "-":
+                if len(line) > 78:
+                  module.fail_json(msg="Length of the line {0} over 80, dataset can not be written".format(line))
+                else:
+                  line = "  {0}".format(line)
```
fernandofloresg, line 3139: I would prefer this addition to be a "smart" addition by only adding if the first two chars are not empty spaces, but will leave this to your consideration, not a deal breaker.

plugins/modules/zos_mvs_raw.py
```diff
@@ -2166,8 +2166,15 @@ def dd_content(contents, dependencies):
...
     if contents is None:
         return None
+    if contents is not None:
+        # Empty string can be passed for content but not modify to ensure proper entry
+        if len(contents) <= 0:
+            return contents
+        contents = modify_contents(contents)
+        return contents
     if isinstance(contents, list):
-        return "\n".join(contents)
```
fernandofloresg, line 2170: Why was this changed?

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+    else:
+        if contents[0] != " " or contents[0] != "-":
+            contents = "  {0}".format(contents)
+        contents = list(contents.split("\n"))
+        contents = add_space(contents)
+        contents = "\n".join(contents)
+    return contents
+
+
+def add_space(lines):
```
fernandofloresg, line 3121: maybe spaces in plural since we add two? another option is prepend_spaces

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3097,49 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+    Args:
+        lines (list): The list with a line of a program.
+
+    Returns:
+        lines: The list in a proper two spaces and the code.
+    """
+    module = AnsibleModule
+    for line in lines:
+        if len(line) == 0:
+            pass
```
fernandofloresg, line 3133: is ok having a pass if len(line) == 0, but why don't we just do something if len(line) > 0? instead of having an if else block ?

fernandofloresg (changes requested):
I will paste my comments on the slack thread since I see some were ignored: 

Due to JCL constraints with input data set
Issue 1: JCL content currently cannot be longer than 80 chars
**Document this in our module**, it is stated in the MVS JCL User guide and we can not modify that behavior, indicate how the user can input multiline yaml texts to not remove newline chars.

Issue 2: Us prepending spaces in each newline
I like your proposal and would like to see the code for it, you got my green light for it, what I think we could do is do it the smart way, **just add the needed empty spaces if those are not already in place.**

This is missing some needed documentation on how we will handle user JCL input from now on, that is very important so users just don't see random record length errors with the new release.

On another front, I will prefer that we only prepend empty spaces when is absolutely needed.

I also added a strange error that I come up while testing a new added code, be sure to test new functionality that you add even if regression tests run fine because we most likely don't have a testing case for that new functionality.

And since we are adding a new functionality, a new test case would be welcomed.

plugins/modules/zos_mvs_raw.py
AndreMarcel99: My bad.

plugins/modules/zos_mvs_raw.py
AndreMarcel99: My bad.

plugins/modules/zos_mvs_raw.py
AndreMarcel99: I now should work on that way but is required that logic, not sure it helps now.

fernandofloresg: I didn't get it. could you explain?

AndreMarcel99: Never mind thanks for the idea

changelogs/fragments/1057-module_zos_mvs_raw_errors_with_long_multi_line_quoted_string_in_content_field.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - zos_mvs_raw - module wasn't able to wrote properly a JCL with multi-line because truncated
+    the first one or two characters writing an syntactically incorrect JCL.
+    Change now validate the length of lines and allow multi-line declaration with | or |1-9 by added
+    two spaces to be syntactically corrects.
```
richp405, line 5: suggestion:
  - zos_mvs_raw - module wasn't able to wrIte a multi-line JCL because it truncated
    the first one or two characters, resulting in syntactically incorrect JCL.
    Change now validates the length of lines and allow multi-line declaration with | or |1-9 by adding
    two spaces to be syntactically corrects.

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3095,43 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+
+    Args:
+        contents (str or list): The string or list with the program.
+
+    Returns:
+        contents: The content in a proper multi line str.
+    """
+    if not isinstance(contents, list):
+        if len(contents) > 1 and contents[0] != " " and contents[1] != " ":
+            contents = "  {0}".format(contents)
```
richp405, line 3109: Would there ever be a case where only 1 space is there?  This caught my eye. but I don't think it needs changed.

richp405 (changes requested):
This looks decent.  I left a question and a suggestion for the changelog fragment.

AndreMarcel99:
Captura de pantalla 2023-11-13 a la(s) 10 43 28 p m
Captura de pantalla 2023-11-13 a la(s) 10 43 43 p m

plugins/modules/zos_mvs_raw.py
AndreMarcel99: Could happen, but by the recommendations is to put 2 spaces

plugins/modules/zos_mvs_raw.py
fernandofloresg: done

plugins/modules/zos_mvs_raw.py
fernandofloresg: done

plugins/modules/zos_mvs_raw.py
```diff
@@ -560,6 +560,9 @@
                 proper literal block style indicator "|".
               - If a list of strings is provided, newlines will be
                 added to each of the lines when used as input.
+              - The maximum length of the line is 80 columns including 2 spaces at the beginning.
+                If the two spaces aren't introduces the module will added to be syntactically correct
```
rexemin, line 564: I think you could rework this a little bit into something like:

> If the two spaces aren't present, the module will add them to form syntactically correct JCL (more information in the Notes section).

plugins/modules/zos_mvs_raw.py
```diff
@@ -1208,6 +1211,10 @@
     - 2. L(zos_mvs_raw,./zos_mvs_raw.html) module execution fails when invoking DFSRRC00 with parm
       "UPB,PRECOMP", "UPB, POSTCOMP" or "UPB,PRECOMP,POSTCOMP". This issue is
       addressed by APAR PH28089.
+    - 3. For syntactically correct JCL programs see the
+      L(JCL reference,https://www.ibm.com/docs/en/zos-basic-skills?topic=collection-coding-your-own-jcl).
+    - 4. For continue a line of dd_input see reference of symbols (+/-) for extended lines
```
rexemin, line 1216: And for this part something like this:

> For continuing a line of C(dd_input) see the reference for symbols (+/-) for extended lines...

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3103,42 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+        lines (list): The list with a line of a program.
+
+    Returns:
+        new_lines: The list in a proper two spaces and the code.
+    """
+    module = AnsibleModuleHelper(argument_spec={})
+    for index, line in enumerate(lines):
+        if len(line) > 0:
+            if len(line) > 78:
+                module.fail_json(msg="""Length of the line {0} over 80. The maximum length of the line is 80 columns including 2 spaces at the beginning.
```
rexemin, line 3135: And lastly, for this error message (italics to highlight changes, the bulk of the message is really good already):

> Length of line {0} _is_ over 80. The maximum length _allowed_ is 80 columns, including 2 spaces at the beginning. If the two spaces are not _present_, the module will _add them to form syntactically correct JCL_.

rexemin (changes requested):

AndreMarcel99:
Captura de pantalla 2023-11-16 a la(s) 3 30 46 p m
Captura de pantalla 2023-11-16 a la(s) 3 31 10 p m

rexemin (approved):

fernandofloresg (approved):

plugins/modules/zos_mvs_raw.py
```diff
@@ -1208,6 +1211,10 @@
     - 2. L(zos_mvs_raw,./zos_mvs_raw.html) module execution fails when invoking DFSRRC00 with parm
       "UPB,PRECOMP", "UPB, POSTCOMP" or "UPB,PRECOMP,POSTCOMP". This issue is
       addressed by APAR PH28089.
+    - 3. For syntactically correct JCL programs see the
+      L(JCL reference,https://www.ibm.com/docs/en/zos-basic-skills?topic=collection-coding-your-own-jcl).
+    - 4. For continuing a line of C(dd_input) see the reference for symbols (+/-) for extended lines
```
ddimatos, line 1216: This should also be removed for a couple of reasons, the doc is linked to program [IBM Automatic Binary Optimizer for z/OS](https://www.ibm.com/docs/en/abo), this doc is talking about its own features not standard JCL which can be found in the JCL reference https://www.ibm.com/docs/en/zos/2.2.0?topic=mvs-zos-jcl-reference.

More importantly, as I explained above , zos_mvs_raw is about invoking programs, just like this module we (the authors) allow certain inputs, if someone is invoking programs IDCAMS, ICKDSF , etc, the user should go to that programs doc and see what if any continuation charector is supported, yes "+" and "-" are commonly supported but not always, sometimes z/OS program authors don't support any continuation symbols.

plugins/modules/zos_mvs_raw.py
```diff
@@ -3090,6 +3102,42 @@ def get_content(formatted_name, binary=False, from_encoding=None, to_encoding=No
...
+
+    Args:
+        lines (list): The list with a line of a program.
+
+    Returns:
+        new_lines: The list in a proper two spaces and the code.
+    """
+    module = AnsibleModuleHelper(argument_spec={})
+    for index, line in enumerate(lines):
+        if len(line) > 0:
```
ddimatos, line 3132: I think we have a few gaps; the logic here checks if the line is greater than 0 which is good, next it checks if the line is greater than 78 which is going to lead to problems if a user gives us content that is 80 chars long and includes the 2 blanks, eg 2 blank chars + 78 non-blanks = 80. 

In the case above, I believe what we should be doing is checking the line is >0, then if the line is <=80 continue, if so, check that positions 0 and 1 are blank , this allows a user to provide us lines that are 80 in length that includes 2 blanks. 

I believe this logic I am proposing will also work for a YAML block indicator e.g. `content: | 4` as this will produce a line with 4 blanks + some non-blank chars up to 80. We would only want to check that the line length is <= 80 and postions 0 and 1 are blank.

If we find a line that is <=80 and has no blanks, we can prepend the blanks in columns 1 and 2 but we also need to check that we don't create a line linger than 80 chars after we add the spaces, if we do we need an appropriate message to the user.

ddimatos (changes requested):
I think we still need a bit of work @AndreMarcel99 , I would also suggest that we maybe add an example using a block sytle indicator with 2 , eg `content: | 2" , pretty sure I had some examples in the original git issues I created you can pull from. They don't have to be lengthy examples but its good if we show one. 

I don't think we can automate with the block style indicator with the pytest-ansible , maybe share a couple of examples in the pull request from the ones I created or ones of your own for historical purposes?

plugins/modules/zos_mvs_raw.py
```diff
@@ -560,6 +560,9 @@
                 proper literal block style indicator "|".
               - If a list of strings is provided, newlines will be
                 added to each of the lines when used as input.
+              - The maximum length of the line is 80 columns including 2 spaces at the beginning.
```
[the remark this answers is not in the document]
ddimatos: One more thing to note, the dd_input: `content` block of text (doc) is in two places in this module, so this has to be updated there as well (copy and paste over it).

AndreMarcel99:
Captura de pantalla 2023-11-28 a la(s) 9 44 15 a m
Captura de pantalla 2023-11-28 a la(s) 9 44 28 a m

ddimatos (approved):
LGTM, thank you.

PR #1058, opened 2023-11-10
Bug 1041 zos submit job honor return output literally

### LINKED ISSUE #1041: [Bug] [zos_submit_job] Option `return_output` when set to false continues to access job output

v1.2.5

### IBM Enterprise Python

v3.9.x

v1.6.0

### ansible-version

v2.14.x

### z/OS version

v2.5

### Ansible module

zos_job_submit

### Bug description

It was reported that when using `zos_job_submit` with option `return_output=false` that the module continues to invoke code that performs job output reads (jobs.read_output). I have not reviewed the code to understand why we access `jobs.read_output`, the fist thing would be to see if there is anything specific the code needs to operate minimally or is this an unnecessary invocation. 

I can see the value in avoiding this call if its not necessary because it is added overhead but more so,

Note: This could also serve as a work around to the known issue of reading chars that don't correspond to UTF-8 (which will be fixed in Q1 2024) by not reading non-UTF8 content. Although it could help, its only note and not the actual issue being highlighted here. 

This was reported by @jbyibm 
Internal thread: https://<sanitized>.slack.com/archives/CSLEJ8VGV/p1697623287742339?thread_ts=1697595176.094609&cid=CSLEJ8VGV

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

richp405:
##### SUMMARY
bugfix: zos_job_submit was not returning dd values, as instructed, but was still requesting them.
This fix should implement 'return output=false" by not asking for dd values in the first place.

fixes #1041 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_job_submit
also added optional input parameter to job_output, but other routines are not calling this function.


richp405:
image

plugins/module_utils/job.py
```diff
@@ -36,7 +36,7 @@
 )
 
 
-def job_output(job_id=None, owner=None, job_name=None, dd_name=None, duration=0, timeout=0, start_time=timer()):
+def job_output(job_id=None, owner=None, job_name=None, dd_name=None, dd_scan=True, duration=0, timeout=0, start_time=timer()):
```
fernandofloresg, line 39: Since dd_scan is added as a parameter in job_output function please also update the docstring of the function

fernandofloresg (changes requested):
Overall looks good, just requested one change in a docstring of a function. I thought about asking for test case but this is my rationale for not asking for one, currently, if we set `return_output` as false, we won't get the result, but internally the code was fetching it, so a test case would have passed even without this change. Hence I think a test case is not required at the moment, would be nice to have one for this feature to ensure we don't break it in the future, will leave that decision to the team, if @richp405 doesn't think we need to include it in this PR is ok, and honestly I think is the best option, but otherwise please create a GH issue to create one test case 😄 .

changelogs/fragments/1041-bug-zos-submit-job-honor-return-output-literally.yml
```diff
@@ -0,0 +1,4 @@
+trivial:
```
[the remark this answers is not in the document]
richp405: I didn't realize trivial were not included.  I've made this change.

plugins/module_utils/job.py
richp405: Thank you for seeing this.  I've made this change.

richp405: Overall, I'm not sure how we can test that data is NOT being pulled, which is why I didn't add a test.

fernandofloresg: agree

richp405:
Added issue #1061 to cover building a test to validate this during unit testing.

changelogs/fragments/1041-bug-zos-submit-job-honor-return-output-literally.yml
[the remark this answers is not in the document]
fernandofloresg: done

fernandofloresg:
Thanks Rich, this looks good.

fernandofloresg (approved):

rexemin (approved):
Looks good, thanks for working on this

PR #1060, opened 2023-11-14
Enabler/1024/remote_tmp for zos_script

### LINKED ISSUE #1024: [Enabler] [zos_script] Deprecate the use of tmp_path in lieu of the engines/ssh.py remote_tmp var.

### Ansible module

_No response_

### Enabler description

This issue is in place so that we remove the the **tmp_path** option currently in the **zos_script** module that allows users to specify where the scripts will be written to on the remote node for execution and instead leverage the one that ansible users which is [remote_tmp](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/sh_shell.html#parameter-remote_tmp) that is part of the ssh connection. 

There are some draw backs to this approach which is we have to extract this value from the engine, the good news is we have already taken control of the ssh connection in our copy module, so we have examples, it should not be too difficult to figure out how to obtain access to that value if we look at the zos_copy plugin or zos_fetch plugin source.

The other draw back is this adds more overhead and tighter coupling with the communities code and means we have to be more diligent in testing ansible-core releases before they GA as we have seen in the past playcontext warnings and variables deprecated. 

See conversation [here](https://github.com/ansible-collections/ibm_zos_core/pull/961/files#r1346820457) for more context.

rexemin:
Fixes #1024. Instead of implementing its own temp path, the module now takes advantage of Ansible's config.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_script

##### ADDITIONAL INFORMATION
Removed the test case for `tmp_path` since now it belongs to the Ansible engine.

### DIFF
--- a/plugins/action/zos_script.py
+++ b/plugins/action/zos_script.py
@@ -56,7 +56,9 @@ class ActionModule(ActionBase):
         if not remote_src:
             script_path = path.abspath(path.normpath(script_path))
             script_name = path.basename(script_path)
-            tmp_path = module_args.get('tmp_path')
+            # Accessing the globally-defined temporary directory
+            # that Ansible expects to be used.
+            tmp_path = self._connection._shell._options.get("remote_tmp")
 
             # Getting a temporary path for the script.
             tempfile_args = dict(
--- a/plugins/modules/zos_script.py
+++ b/plugins/modules/zos_script.py
@@ -92,18 +92,6 @@ options:
         script will not be executed.
     type: str
     required: false
-  tmp_path:
-    description:
-      - Directory path in the remote machine where local scripts will be
-        temporarily copied to.
-      - When not specified, the module will copy local scripts to
-        the default temporary path for the user.
-      - If C(tmp_path) does not exist in the remote machine, the
-        module will not create it.
-      - All scripts copied to C(tmp_path) will be removed from the managed
-        node before the module finishes executing.
-    type: str
-    required: false
 
 extends_documentation_fragment:
   - ibm.ibm_zos_core.template
@@ -153,12 +141,6 @@ EXAMPLES = r"""
     cmd: /u/user/scripts/ARGS "1,2"
     remote_src: true
     chdir: /u/user/output_dir
-
-- name: Run a local Python script that uses a custom tmp_path.
-  zos_script:
-    cmd: ./scripts/program.py
-    executable: /usr/bin/python3
-    tmp_path: /usr/tmp/ibm_zos_core
 
 - name: Run a local script made from a template.
   zos_script:
@@ -251,7 +233,6 @@ def run_module():
             executable=dict(type='str', required=False),
             remote_src=dict(type='bool', required=False),
             removes=dict(type='str', required=False),
-            tmp_path=dict(type='str', required=False),
             use_template=dict(type='bool', default=False),
             template_parameters=dict(
                 type='dict',
@@ -287,7 +268,6 @@ def run_module():
         executable=dict(arg_type='path', required=False),
         remote_src=dict(arg_type='bool', required=False),
         removes=dict(arg_type='path', required=False),
-        tmp_path=dict(arg_type='path', required=False),
         use_template=dict(arg_type='bool', required=False),
         template_parameters=dict(
             arg_type='dict',
--- a/plugins/modules/zos_script.py
+++ b/plugins/modules/zos_script.py
@@ -100,6 +100,12 @@ notes:
   - When executing local scripts, temporary storage will be used
     on the remote z/OS system. The size of the temporary storage will
     correspond to the size of the file being copied.
+  - The location in the z/OS system where local scripts will be copied to can be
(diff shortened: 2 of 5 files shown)


rexemin:
Pipeline results with Ansible 2.14, Python 3.11 and ZOAU 1.2.5.

Screenshot 2023-11-14 at 14 57 15
Screenshot 2023-11-14 at 14 57 27

richp405 (approved):
Nice little surgery.

fernandofloresg (approved):
Nice and clean, looks good to me.

plugins/modules/zos_script.py
```diff
@@ -112,6 +100,12 @@
   - When executing local scripts, temporary storage will be used
     on the remote z/OS system. The size of the temporary storage will
     correspond to the size of the file being copied.
+  - The location in the z/OS system where local scripts will be copied to can be
+    configured through Ansible's C(remote_tmp) option. Refer to
+    L(Ansible's documentation,https://docs.ansible.com/ansible/latest/collections/ansible/builtin/sh_shell.html#parameter-remote_tmp)
```
ketankelkar, line 105: curious why the link here is to the shell module (which seems to be an example use case of the remote_tmp environment var)


Maybe the link here should be one of the following:
- https://docs.ansible.com/ansible/latest/collections/environment_variables.html#envvar-ANSIBLE_REMOTE_TEMP
- https://docs.ansible.com/ansible/latest/collections/environment_variables.html#envvar-ANSIBLE_REMOTE_TMP

Not sure what the difference between tmp and temp is...

BUT the link you provided may also be the correct one. If the unix shell plugin is the only one in the list that's applicable, it may be nicer to point users to a more detailed description of the remote_tmp var (current link) over the links i've provided above which point to the generic env vars in the list of all ansible env vars
image

plugins/modules/zos_script.py
```diff
@@ -154,12 +148,6 @@
     remote_src: true
     chdir: /u/user/output_dir
 
-- name: Run a local Python script that uses a custom tmp_path.
```
ketankelkar, line 157: i think it's worth leaving a version of this example in the docs. i see you have a statement about how to set temp path in the `notes` section, but i think more users will jump straight to the `examples` section and possibly miss learning about the available config. I'm thinking the example could be something like:
```yml
- name: Run a local Python script in the temporary directory specified in the Ansible environment variable 'remote_tmp'.
   cmd: ./scripts/program.py
   executable: /usr/bin/python3
```
where just the playbook task name briefly highlights where the config is located/managed from.

plugins/action/zos_script.py
```diff
@@ -56,7 +56,9 @@ def run(self, tmp=None, task_vars=None):
         if not remote_src:
             script_path = path.abspath(path.normpath(script_path))
             script_name = path.basename(script_path)
-            tmp_path = module_args.get('tmp_path')
+            # Accessing the globally-defined temporary directory
+            # that Ansible expects to be used.
+            tmp_path = self._connection._shell._options.get("remote_tmp")
```
ketankelkar, line 61: i think we briefly talked about this over slack, but just to re-reconfirm, is there a base/default value here? is there any chance that the call will return `None` or something similar which could set that `tmp_path` to something invalid and cause issues later on in the module execution?

ketankelkar (approved):
LGTM, some minor (and optional) suggestions, nothing major, hence the approval

plugins/modules/zos_script.py
ddimatos: If I recall `remote_tmp` was at one time the replacement for `remote_temp` which never seemed to have happened. Back in ansible 2.5 they had this in the release notes:
```
The ANSIBLE_REMOTE_TMP environment variable has been added to supplement (and override) ANSIBLE_REMOTE_TEMP. This matches with the spelling of the config value. ANSIBLE_REMOTE_TEMP will be deprecated in the future
```

I believe most the plugins eventually discuss the env var `remote_tmp` or `remote_temp` and point to the link you suggested above (https://docs.ansible.com/ansible/latest/collections/environment_variables.html#envvar-ANSIBLE_REMOTE_TMP) and (https://docs.ansible.com/ansible/latest/collections/environment_variables.html#envvar-ANSIBLE_REMOTE_TEMP).

ddimatos (approved):
@rexemin nice work, I did not see anything more than what @ketankelkar has already identified so I will approve pending the discussions be resolved with @ketankelkar .

plugins/action/zos_script.py
rexemin: There's a default value that Ansible sets when the user doesn't specify anything in config files or environment variables: `~/.ansible/tmp`.

plugins/modules/zos_script.py
rexemin: I agree that linking to the environment variable directly is probably more correct, but the shell module has a more in-depth explanation of what it does, while also linking to the environment vars docs.

ketankelkar:
I think all my comments have been addressed, I'm good with this being merged

PR #1068, opened 2023-11-23
[v1.8.0] [Backport] [zos_script] remote_tmp for zos_script

### LINKED ISSUE #1024: [Enabler] [zos_script] Deprecate the use of tmp_path in lieu of the engines/ssh.py remote_tmp var.

rexemin:
##### SUMMARY

Backport of #1060 for v1.8.0. Fixes #1024.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_script

### DIFF
--- a/changelogs/fragments/1060-remote_tmp_zos_script.yml
+++ b/changelogs/fragments/1060-remote_tmp_zos_script.yml
@@ -1,0 +1,5 @@
+minor_changes:
+  - zos_script - add support for remote_tmp from the Ansible
+    configuration to setup where temporary files will be created,
+    replacing the module option tmp_path.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1060).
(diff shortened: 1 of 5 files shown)


AndreMarcel99 (approved):

fernandofloresg (approved):
lgtm

PR #1079, opened 2023-12-04
[zos_copy] fix for executables copied from local fail with iconv error

### LINKED ISSUE #1080: [Bug] [zos_copy] zos_copy fails when copying a local executable

v1.2.5

### IBM Enterprise Python

v3.12.x

v1.8.0-beta.1

### ansible-version

latest

### z/OS version

v2.5

### Ansible module

zos_copy

### Bug description

Forward port fix for #1079 

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

This fixes #1080 for branch v1.8.0, essentially this removes any encoding for when a local file is sent to remote, also, when dest_type is not available i.e. dest doesn't not already exists now we set the dest_ds_type = 'PDSE' when executable is selected and we know dest is not USS (validation performed on src checking for '/'), which then gets changed to "LIBRARY", note that I also added a check when copying to PDS/PDSE, that is because when the a local file is copied to remote dest_ds_type gets changed to "LIBRARY" however this doesn't occur when copying an executable from the remote (I believe the data set type we get is PDSE) and that needs to be further investigated in #1080 for v1.9.0 fix.

This current version does solve the issue and does not break anything that we are testing for, so should be a fix for v1.8.0.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_copy

### DIFF
--- a/changelogs/fragments/1079-zos-copy-local-executable.yml
+++ b/changelogs/fragments/1079-zos-copy-local-executable.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - zos_copy - When copying an executable data set from controller to managed node,
+    copy operation failed with an encoding error. Fix now avoids encoding when executable
+    option is selected.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1079).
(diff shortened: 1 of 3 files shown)


richp405 (approved):
Looks good.

ddimatos (approved):
@fernandofloresg - this looks great, what a fast turn around I am sure everyone appreciates that , I know I do :).  Great test case design as well.

Issue #999, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 11


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

11)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:24: sanity[cannot-ignore]: Ignore file contains pylint:catching-non-exception at line 24, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_job_submit.py pylint:catching-non-exception # False positive, Exception is inherited

Issue #998, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 10


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

10)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:23: sanity[cannot-ignore]: Ignore file contains validate-modules:undocumented-parameter at line 23, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_job_submit.py validate-modules:undocumented-parameter # Passing args from action plugin

Issue #997, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 9


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

9)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:22: sanity[cannot-ignore]: Ignore file contains validate-modules:parameter-type-not-in-doc at line 22, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_job_submit.py validate-modules:parameter-type-not-in-doc # Passing args from action plugin

PR #1091, opened 2023-12-12
[Enabler][Zos Job Submit]Update undocumented argument and import exception

AndreMarcel99:
##### SUMMARY
Update temp_path and how to import exceptions.

Fiexes #997 #998 #999 

##### ISSUE TYPE
- Sanity Issue Pull Request

##### ADDITIONAL INFORMATION
Remove tmp path from options and only work with source as name of tmp file, because it was undocumented argument for ansible and could be any name, also only in case is local make the tmp file, so didn't expect a dataset or file with the same name on the system. Also make a implementation of imports of exceptions by using new implementation to avoid losing context when an import fails.

Pipeline Without Zos_copy
Captura de pantalla 2023-12-12 a la(s) 4 08 41 p m
Captura de pantalla 2023-12-12 a la(s) 4 09 21 p m

Pipeline with zos_copy

### DIFF
--- a/plugins/modules/zos_job_submit.py
+++ b/plugins/modules/zos_job_submit.py
@@ -620,7 +620,6 @@ from ansible_collections.ibm.ibm_zos_core.plugins.module_utils.data_set import (
     DataSet,
 )
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.six import PY3
 from timeit import default_timer as timer
 from tempfile import NamedTemporaryFile
 from os import remove
@@ -638,10 +637,6 @@ try:
 except Exception:
     jobs = ZOAUImportError(traceback.format_exc())
 
-if PY3:
-    from shlex import quote
-else:
-    from pipes import quote
 
 JOB_COMPLETION_MESSAGES = frozenset(["CC", "ABEND", "SEC ERROR", "JCL ERROR", "JCLERR"])
 JOB_ERROR_MESSAGES = frozenset(["ABEND", "SEC ERROR", "SEC", "JCL ERROR", "JCLERR"])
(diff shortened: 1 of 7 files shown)


plugins/modules/zos_job_submit.py
```diff
@@ -902,7 +899,7 @@ def run_module():
     from_encoding = parsed_args.get("from_encoding")
     to_encoding = parsed_args.get("to_encoding")
     # temporary file names for copied files when user sets location to LOCAL
-    temp_file = parsed_args.get("temp_file")
+    temp_file = parsed_args.get("src")
```
fernandofloresg, line 902: I would suggest doing this change instead : 

```suggestion
    if location == "LOCAL":
        temp_file = parsed_args.get("src")
```

That way `temp_file` is indeed a temporary file on the remote and we don't need to worry on the `location` conditions in the future, if `temp_file` is None we know we don't have to remove anything.

plugins/modules/zos_job_submit.py
```diff
@@ -1039,7 +1036,7 @@ def run_module():
         module.exit_json(**result)
 
     finally:
-        if temp_file:
+        if location != "DATA_SET" and location != "USS":
```
fernandofloresg, line 1039:
```suggestion
        if temp_file:
```

I suggest keeping it this way bc that will help us avoid having to check for state and just remove it if it exists.

fernandofloresg (changes requested):
Requested changes to avoid tech debt in the future and also add a changelog please.

plugins/modules/zos_job_submit.py
AndreMarcel99: It generates a error if the local file is named as a dataset (USER.LOCAL) then the file /user.local will fail that verification

fernandofloresg: I'm sorry I don't get how that relates to this code suggestion, could you explain?

AndreMarcel99:
Pipeline for all except zos_copy
Captura de pantalla 2023-12-14 a la(s) 10 22 01 a m

Captura de pantalla 2023-12-14 a la(s) 10 21 47 a m

AndreMarcel99:
Pipeline zos_copy
Captura de pantalla 2023-12-14 a la(s) 2 34 44 p m
Captura de pantalla 2023-12-14 a la(s) 2 35 19 p m

fernandofloresg:
I made some extra changes and I'm waiting for the pipeline to finish.

fernandofloresg:
Screenshot 2023-12-15 at 12 15 28 PM

fernandofloresg:
To add in case anyone wonders why `iconv` section is because when `zos_copy` module was integrated into `zos_job_submit` encoding option was not used, this way we can leverage `zos_copy` encoding capabilities and standardize copying files into remote.

fernandofloresg (approved):
Looks good to me now

richp405 (approved):
Looks good!

rexemin (approved):

Issue #1121, opened 2023-12-20
[Enabler] [Dev] [module_utils/job.py] Migrate module, module utilities and test cases to ZOAU 1.3 


ddimatos:
## module_utils
`job.py`

### See Epic
[[Enabler] [Epic] Adopt new ZOAU 1.3 Python changes](https://github.com/ansible-collections/ibm_zos_core/issues/931)

### Enabler description

With the adoption of ZOAU 1.3, modules, module utilities and test cases may need to be updated according to the [migration guide](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier). 

APIs changes affecting module:
- Impacted by ZOAU Dataset API: No
- Impacted by ZOAU jobs API: Yes
- Impacted by the rename of types.py to ztypes.py: No
- Impacted by the contents of module volumes.py having been moved to datasets.py: No
- Impacted by opercmd timeout to milli seconds: No
- Impacted by JSON schema change: No
- Impacted by Exceptions schema: Yes

Review the [zoautil_py datasets module migration doc](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier#changes-to-the-datasets-module).

Upon review, this module is affected by:
- the jobs.listing() function is renamed to jobs.fetch_multiple()
   - see `entries = jobs.listing(**kwargs)` - 2 lines like this
- the `id` attribute is renamed to `job_id`
   - see `if not fnmatch.fnmatch(entry.id, job_id):`
   - see `job["job_id"] = entry.id`
   - see `list_of_dds = jobs.list_dds(entry.id)`
   - see `list_of_dds = jobs.list_dds(entry.id)`
   - see `entry.id, single_dd["stepname"], single_dd["dataset"])`
- the `rc` attribute is renamed to `return_code`
   - see `job["ret_code"]["msg"] = entry.status + " " + entry.rc`
   - see `job["ret_code"]["msg_code"] = entry.rc`
   - see
      ```
            if len(entry.rc) > 0:
                if entry.rc.isdigit():
                    job["ret_code"]["code"] = int(entry.rc)
     ```
[...]

PR #1169, opened 2024-01-16
[v1.10.0] [Enabler] [job] ZOAU v1.3.0 migration for job.py

rexemin:
##### SUMMARY
Addresses #1121. It only changes the module util and no more.

##### ISSUE TYPE
- Enabler Pull Request

##### COMPONENT NAME
module_util/job.py

##### ADDITIONAL INFORMATION
Since the module util relies on the object that ZOAU returns, I don't see much we can gain from switching to the JSON output that ZOAU now offers, so the changes are only for calls to the new interface.


richp405 (approved):

ddimatos (approved):
@rexemin looks great, nice catch on the svc_class that was not documented.

fernandofloresg (approved):

PR #1176, opened 2024-01-17
[v1.9.0] [zos_copy] Enhancement/764/copy members

fernandofloresg:
##### SUMMARY

Improved the performance of copying a PDS/PDSE with multiple members into another PDS/PDSE by providing the complete list of members to copy to datasets._copy zoau call.

This is some performance numbers comparing current dev copy as "old copy" and proposed solution as "new copy"

Testing copy with 30 members

old copy 474.17s, 339.17s, 310.62s, 607.13s
new copy 35.75s, 55.13s, 45.11, 51.61s

Testing copy with 100 members

old copy 1275.75s (21 mins), 1400.44s(23 mins)

new copy 47.15s, 115.71s

##### ISSUE TYPE

- Feature Pull Request

Also corrected a couple of false positives in the test cases section, one was trying to copy an empty PDS into another and was a false positive in the module code. Another, was always creating a PDSE regardless of the type described in the test.

##### COMPONENT NAME

zos_copy

### DIFF
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# Copyright (c) IBM Corporation 2019 - 2023
+# Copyright (c) IBM Corporation 2019 - 2024
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


fernandofloresg:
Screenshot 2024-01-22 at 1 24 50 PM

Screenshot 2024-01-22 at 1 37 12 PM

richp405 (approved):

rexemin (approved):

plugins/modules/zos_copy.py
```diff
@@ -1705,33 +1705,56 @@ def copy_to_pdse(
...
                 overwritten_members.append(destination_member)
             else:
                 new_members.append(destination_member)
-
+            bulk_src_members += "{0} ".format(src_member)
+
+        # Copy section
+        if src_ds_type == "USS" or self.asa_text or len(src_members) == 1:
+            """
+            USS -> MVS : Was kept on member by member basis bc file names longer that 8
```
ddimatos, line 1721: Your comment made me think of something, this is a good point **"file names longer that 8 characters will throw an error when copying to a PDS, because of the member name character limit."** , do we need an issue to address this, we could document that when we encounter names longer than 8 going to a PDS/E we will truncate them to 8, this is probably more common in a `dir` to PDSE copy where they did not carefully inspect the file names in a directory or just let it error. 

There is a `is_compatible` code section we could fit this in the future.

ddimatos (approved):
Really nice work @fernandofloresg - the detailed comments are awesome, made it so easy to follow. I had one question which you can comment and resolve, the other comment was more general, is there a need to let users know that MVS --> MVS is faster, I am thinking we don't need to document this because it will be confusing, maybe just point it out in the release blog we made improvements in one type of copy.

plugins/modules/zos_copy.py
fernandofloresg: Current behavior when copying a dir to a PDS/E is to truncate to the first 8 characters and remove any dots found(maybe we can explore what happens with files including special char, it may fail). We could add into the `src` section :
```
If src is USS, file names will be truncated and/or modified to ensure a valid name for a data set or member.
```
But that is kind of obvious?

Also, about documenting copy between MVS -> PDS/E being faster, we could add into `asa_text` section:
```
Using asa_text will result in a slower copy performance when copying multiple members.
```

What do you think?

PR #1179, opened 2024-01-18
[v1.10.0][zos_lineinfile] Remove zos encode from zos lineinfile

### LINKED ISSUE #1134: Enabler] Remove zos_encode dependency from zos_lineinfile test cases to aid in ZOAU 1.3 migration
### Ansible module functional test case

`test_zos_lineinfile_func.py`

### Enabler description

Module `zos_encode` is used in the identified test case `test_zos_lineinfile_func.py` that results in several behaviors that this issue will change:
- It will remove the need to run this test suite for every `zos_encode` change, which is detected by our dependency finder.
- It will reduce the test case module dependency to ease the current migration and future ones. 

See the  following is a sample code:
```
hosts.all.zos_encode(src=full_path, dest=full_path, from_encoding="IBM-1047", to_encoding=params["encoding"])
```

fernandofloresg:
##### SUMMARY

Remove zos_encode from zos_lineinfile.

Fixes #1134

##### ISSUE TYPE

- Enhancement Pull Request

##### COMPONENT NAME

zos_lineinfile

### DIFF
--- a/changelogs/fragments/1179-remove-zos_encode-from_zos_lineinfile-tests.yml
+++ b/changelogs/fragments/1179-remove-zos_encode-from_zos_lineinfile-tests.yml
@@ -1,0 +1,3 @@
+trivial:
+  - zos_lineinfile - Remove zos_encode dependency from zos_lineinfile test cases.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1179).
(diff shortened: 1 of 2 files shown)


fernandofloresg:
Screenshot 2024-01-18 at 12 21 10 PM

Screenshot 2024-01-18 at 12 21 33 PM

rexemin (approved):
Looks good. As we discussed yesterday, I think it's fine to leave the use of f-strings since the tests are not packaged with the rest of the release and thus not subject to problems with Python 2. And they're faster and prettier IMO.

PR #1184, opened 2024-01-25
[Enabler] [zos_copy] Remove zos_fetch call in loadlib test

### LINKED ISSUE #1147: [Enabler] Remove zos_fetch dependency from zos_copy test cases to aid in ZOAU 1.3 migration
### Ansible module functional test case

`test_zos_copy_func.py`

### Enabler description

Module `zos_fetch` is used in the identified test case `test_zos_copy_func.py` that results in several behaviors that this issue will change:
- It will remove the need to run this test suite for every `zos_fetch` change, which is detected by our dependency finder.
- It will reduce the test case module dependency to ease the current migration and future ones. 

See the  following is a sample code:
```
fetch_result = hosts.all.zos_fetch(src=src_lib, dest=tmp_folder.name, is_binary=True)
```

rexemin:
##### SUMMARY
Fixes #1147.

##### ISSUE TYPE
- Enabler Pull Request

##### COMPONENT NAME
- zos_copy (tests)

##### ADDITIONAL INFORMATION
One of the loadlib tests had a call to zos_fetch, so our dependency finder dictated that every time zos_fetch got updated, we had to run all tests for zos_copy, which takes forever.
This change replaces that call to instead use `scp` to transfer loadlibs to the local filesystem.

### DIFF
--- a/changelogs/fragments/1184-remove-zos-fetch-dep-from-zos-copy.yml
+++ b/changelogs/fragments/1184-remove-zos-fetch-dep-from-zos-copy.yml
@@ -1,0 +1,3 @@
+trivial:
+  - zos_copy - Remove zos_fetch dependency from zos_copy test cases.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1184).
--- a/tests/functional/modules/test_zos_copy_func.py
+++ b/tests/functional/modules/test_zos_copy_func.py
@@ -3502,18 +3502,19 @@ def test_copy_local_pds_loadlib_to_pds_loadlib(ansible_zos_module, is_created):
         # string instead of a list.
         remote_host = hosts["options"]["inventory"].replace(",", "")
 
-        tmp_folder =  tempfile.TemporaryDirectory(prefix="tmpfetch")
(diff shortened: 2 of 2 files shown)


ddimatos:
I took a look while in draft, interesting problem and solution :). I suppose you considered the community fetch module, the change looks good to me, thank you.

rexemin:
Testing pipeline results.
Screenshot 2024-01-30 at 12 39 28
Screenshot 2024-01-30 at 12 39 48

fernandofloresg (approved):
Change looks good, thanks.

rexemin:
Do note that in the process of removing the call to `zos_fetch` I added a call to a subprocess. Security scans won't like that but @fernandofloresg told me that tests are out of those scans and so it should be okay. Another path would have been to install the `scp` library both locally and in the Jenkins VM, but since it depends on `paramiko`, I'm not sure if we would have licensing problems. In the end, I decided it was easier to deal with the subprocess instead.

ddimatos (approved):
Looks good to me, nice solution, thank you.

richp405 (approved):

Issue #994, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 6


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

6)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:13: sanity[cannot-ignore]: Ignore file contains validate-modules:undocumented-parameter at line 13, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_data_set.py validate-modules:undocumented-parameter # Keep aliases to match behavior of old module spec, but some aliases were functionally inaccurate, and detailing in docs would only confuse user.

Issue #993, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 5


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

5)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:11: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-type-does-not-match-spec at line 11, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_data_set.py validate-modules:doc-type-does-not-match-spec # Have to use raw here for backwards compatibility with old module args, but would confuse current users if exposed.

Issue #992, opened 2023-09-28
[Enabler] Review and update sanity test ignore entry issue 4


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

This issue is part of epic #811 that will remove the need for the sanity test ignore entry.

4)**ansible-lint error:**
tests/sanity/ignore-2.15.txt:10: sanity[cannot-ignore]: Ignore file contains validate-modules:doc-choices-do-not-match-spec at line 10, which is not a permitted ignore.

**ignore-txt entry:**
plugins/modules/zos_data_set.py validate-modules:doc-choices-do-not-match-spec # We use our own argument parser for advanced conditional and dependent arguments.

PR #1191, opened 2024-01-26
Correct sanity alerts in zos_data_set module.

richp405:
##### SUMMARY
This ticket is for corrections to the zos_data_set module in the current 1.9 staging branch.

Changes will resolving sanity test linting issues #992 #993 #994

Also corrected 'truthy' findings in the zos_data_set in-code documentation
('true' replacing 'Yes' in examples)

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
zos_data_set

##### ADDITIONAL INFORMATION
There were a few features that had a list of choices in the documentation, but loaded the value as a 'string' in code.  This caused sanity check errors, so we put these issues into 'sanity ignore'.  The ability to ignore warnings is being removed, so we need to address the issues and remove the ignore statements.


richp405:
closing in favor of 1285

PR #1203, opened 2024-02-02
[Github] Issues template update proposal 

fernandofloresg:
##### SUMMARY

This proposal updates the ansible core collection version up to v1.10 and adds latest as an option to select.
Also, for ZOAU it adds v1.3.0, which even if is not supported since is GA we can have it there, also, added None as a value for zoau version selector.
Updated the list of selectable modules since some like zos_archive, zos_unarchive and zos_volume_init were missing.

##### ISSUE TYPE

- Docs Pull Request


richp405 (approved):
If I remember correctly, we don't usually make changelog fragments for the issue form.

ketankelkar (approved):
LGTM - Thanks for being on top of this!!

Although if we don't update the form for the newer version of our collection, that would means users couldn't open up bugs against them right? 🤔 
Pretty sure that's how it works

fernandofloresg:
#1226 better addresses this

PR #1219, opened 2024-02-12
[Sanity] Changelog has wrong value "minor_change" instead of "minor_changes"

fernandofloresg:
##### SUMMARY

Updated a changelog that was breaking the `antsibull-changelog lint` command to the correct key value.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
changelog


richp405 (approved):
At least you don't have to make a changelog entry for a change in the changelog!

fernandofloresg:
Hey will take this back bc even if it breaks current dev branch this should be removed shortly by other PR.

Issue #1223, opened 2024-02-13
[Enabler] [ac] The development tool lacked support for ansible-core 2.16, a typo causing copy errors and did not support conditional operators for Python version.


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enabler description

1. The `ac` tool had a typo that was causing copy and paste errors when using the help function, `passsword`.
2. The `ac` tooldid not support in the requirements.env the idea of a conditional operator for python, eg, "==" so "==:python:3.9" or ">:python:3.9" which is helpful when you want to use the latest version of python on a host that is permitted for the version of ansible. Before this, with the the operator, the latest version of python is selected, eg python 13 and it might not be supported by the venv being created for an ansible version. 
3. The `ac` tool also when no password is supplied would echo out the env vars to use, but to aid users now a dummy stub configuration is included with the venv creation that can be later edited easing a user from having to figure out the format if `info.env`.
4. Add a requirements file for ansible-core 2.16
5. Also the search order is changed so that the latest version on a host is evaluated first.

PR #1224, opened 2024-02-13
Enabler updates the AC tool with operator support for python and changes python search order.

ddimatos:
##### SUMMARY
Addresses #1223 

- The ac tool had a typo that was causing copy and paste errors when using the help function, passsword.
- The ac tool did not support in the requirements.env the idea of a conditional operator for python, eg, "==" so "==:python:3.9" or ">:python:3.9" which is helpful when you want to use the latest version of python on a host that is permitted for the version of ansible. Before this, with the the operator, the latest version of python is selected, eg python 13 and it might not be supported by the venv being created for an ansible version.
- The ac tool also when no password is supplied would echo out the env vars to use, but to aid users now a dummy stub configuration is included with the venv creation that can be later edited easing a user from having to figure out the format if info.env.
- Add a requirements file for ansible-core 2.16
- Also the search order is changed so that the latest version on a host is evaluated first.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/scripts/requirements-2.16.env
+++ b/scripts/requirements-2.16.env
@@ -1,6 +1,6 @@
 #!/bin/sh
 # ==============================================================================
-# Copyright (c) IBM Corporation 2022, 2023
+# Copyright (c) IBM Corporation 2024
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/scripts/requirements-2.16.env
+++ b/scripts/requirements-2.16.env
@@ -22,14 +22,11 @@
 # ==============================================================================
 
 requirements=(
-"ansible-core:2.11.12"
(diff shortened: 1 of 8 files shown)


scripts/requirements-2.16.env
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+# ==============================================================================
+# Copyright (c) IBM Corporation 2022, 2023
```
richp405, line 3: 2024, since this is a new script?

richp405 (approved):
noting that all the .env files are copyright -2023, but I think at least the newer one (2.16) should be 2024.  Not a show stopper.

scripts/requirements-2.16.env
ddimatos: Done, thank you.

fernandofloresg (approved):
Looking good from here

Issue #1244, opened 2024-02-21
[Bug] zos_lineinfile_did_not_work_properly_with_datasets_using_regex_option_and_insert_before_or_after


AndreMarcel99:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Are the dependencies a supported version?

- [X] The dependencies are supported.

### IBM Z Open Automation Utilities

v1.2.5

### IBM Enterprise Python

v3.11.x

### IBM z/OS Ansible core Version

v1.8.0-beta.1

### ansible-version

v2.15.x

### z/OS version

None

### Ansible module

zos_lineinfile

### Bug description

Output of a uss: `'dsedhelper -d -c IBM-1047 -s -e /abcxyz/c\\ZOAU_ROOT=/mvsutil-develop_dsed/$ -e /ZOAU_ROOT=/a\\ZOAU_ROOT=/mvsutil-develop_dsed/$ -e $ a\\ZOAU_ROOT=/mvsutil-develop_dsed /tmp/ansible-core-tests/zos_lineinfile/test_uss_line_replace_nomatch_insertafter_match '`
Output of a ds: `'BGYSC4909E Unable to write into dataset ANSIBLE.PCV3BTKF.T4233329.CX3V6Z02\n'` with the call `'dsedhelper -d -c IBM-1047 -s -e /abcxyz/c\\ZOAU_ROOT=/mvsutil-develop_dsed/$ -e /ZOAU_ROOT=/a\\ZOAU_ROOT=/mvsutil-develop_dsed/$ -e $ a\\ZOAU_ROOT=/mvsutil-develop_dsed ANSIBLE.PCV3BTKF.T4233329.CX3V6Z02`

### Ansible configuration.

```YAML
With ZOAU 1.3
```

Issue #1111, opened 2023-12-20
[Enabler] [Dev] [ zos_lineinfile] Migrate module, module utilities and test cases to ZOAU 1.3


ddimatos:
## Ansible module
`zos_lineinfile`

### See Epic
[[Enabler] [Epic] Adopt new ZOAU 1.3 Python changes](https://github.com/ansible-collections/ibm_zos_core/issues/931)

### Enabler description

With the adoption of ZOAU 1.3, modules, module utilities and test cases may need to be updated according to the [migration guide](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier). 

APIs changes affecting module:
- Impacted by ZOAU Dataset API: Yes
- Impacted by ZOAU jobs API: No
- Impacted by the rename of types.py to ztypes.py: No
- Impacted by the contents of module volumes.py having been moved to datasets.py: No
- Impacted by opercmd timeout to milli seconds: No
- Impacted by JSON schema change in dsed: Yes 
- Impacted by Exceptions schema: No

Review the [zoautil_py datasets module migration doc](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier#changes-to-the-datasets-module).

Upon review, this module is affected by:
- the `dataset` class being relocated to zoautil_py.datasets.Dataset.
   - See `from zoautil_py import datasets`
- the function datasets.lineinfile(), the parameter `ins_bef` is renamed to `insert_before`, and the parameter `ins_aft` is renamed to `insert_after`.
  - see: 
    ```
        return datasets.lineinfile(
        src,
        line,
        regex=regexp,
        ins_aft=ins_aft,
        ins_bef=ins_bef,
        encoding=encoding,
        first_match=first_match,
        backref=backrefs,
        state=True,
        debug=True,
        force=force,
    )
    ```
   - see: 
     ```
      return_content = present(src, quotedString(line), quotedString(regexp), quotedString(ins_aft), quotedString(ins_bef), encoding, firstmatch, backrefs, force)
    ```
   - see:
     ```
        if ins_bef:
            stdout = stdout.replace(ins_bef, quotedString(ins_bef))
     ```
   - see: `def present(src, line, regexp, ins_aft, ins_bef, encoding, first_match, backrefs, force):`
   - see:
      ```
        if ins_aft:
[...]

PR #1256, opened 2024-02-26
[Enabler][Migration]Migrate_zos_blockinfile_and_lineinfile

### LINKED ISSUE #1105: [Enabler] [Dev] [zos_blockinfile] Migrate module, module utilities and test cases to ZOAU 1.3
## Ansible module
`zos_blockinfile`

### See Epic
[[Enabler] [Epic] Adopt new ZOAU 1.3 Python changes](https://github.com/ansible-collections/ibm_zos_core/issues/931)

### Enabler description

With the adoption of ZOAU 1.3, modules, module utilities and test cases may need to be updated according to the [migration guide](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier). 

APIs changes affecting module:
- Impacted by ZOAU Dataset API: Yes
- Impacted by ZOAU jobs API: No
- Impacted by the rename of types.py to ztypes.py: No
- Impacted by the contents of module volumes.py having been moved to datasets.py: No
- Impacted by opercmd timeout to milli seconds: No
- Impacted by JSON schema change in dmod: Yes , but this appears to be isolated to the zoautil python APIs and no impact to this module
- Impacted by Exceptions schema: No

Review the [zoautil_py datasets module migration doc](https://www.ibm.com/docs/en/zoau/1.3.x?topic=planning-migrating-zoau-v13-from-v12x-earlier#changes-to-the-datasets-module).

Upon review, this module is affected by:
- the `dataset` class being relocated to zoautil_py.datasets.Dataset.
   - See `from zoautil_py import datasets`
- the functionsdatasets.blockinfile(), the parameter `ins_bef` is renamed to `insert_before`, and the parameter `ins_aft` is renamed to `insert_after`.
   - see: 
        ```
        return datasets.blockinfile(src, block=block, marker=marker, ins_aft=ins_aft, ins_bef=ins_bef, encoding=encoding, state=True, options=force, as_json=True)
        ```
        ```
        return datasets.blockinfile(src, marker=marker, encoding=encoding, state=False, options=force, as_json=True)
        ```

Note, that the content for the `zoauresponse` has been encapsulated with a `data` object (#954), while I don't believe we return any content, validation should be covered through regression testing.

For reference the JSON Schema changes with examples should also be reviewed:
https://www.ibm.com/docs/en/zoau/1.3.x?topic=examples-json-schema

#954 - we may cancel this issue if the work is handled on each module

This issue depends on shared import identified by git issues:
- [module_utils/backup.py](https://github.com/ansible-collections/ibm_zos_core/issues/1115)
- [module_utils/better_arg_parser.py](https://github.com/ansible-collections/ibm_zos_core/issues/1116)
- [module_utils/copy.py](https://github.com/ansible-collections/ibm_zos_core/issues/1117)
- [module_utils/data_set.py](https://github.com/ansible-collections/ibm_zos_core/issues/1118)
- [module_utils/dd_statement.py](https://github.com/ansible-collections/ibm_zos_core/issues/1119)
- [module_utils/encode.py](https://github.com/ansible-collections/ibm_zos_core/issues/1120)

AndreMarcel99:
##### SUMMARY
Migrate the block and lineinfile to zoau 1.3

Fixes #1105 #1111 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Migrate the modules zos lineinfile and zos blocinfile to zoau 1.3

##### ADDITIONAL INFORMATION
During the test cases review it came a issue #1244 needs follow up with NAZARE-10439
Captura de pantalla 2024-02-26 a la(s) 9 26 05 a m

Captura de pantalla 2024-02-26 a la(s) 9 25 46 a m


plugins/modules/zos_blockinfile.py
```diff
@@ -325,17 +325,18 @@
...
 from ansible_collections.ibm.ibm_zos_core.plugins.module_utils.import_handler import (
-    MissingZOAUImport,
+    ZOAUImportError,
 )
 
 try:
     from zoautil_py import datasets
 except Exception:
-    Datasets = MissingZOAUImport()
+    Datasets = ZOAUImportError(traceback.format_exc())
```
rexemin, line 339: This one should be `datasets`, all in lower case, so the exception gets handled properly when calling the ZOAU API later in the code.

plugins/modules/zos_blockinfile.py
```diff
@@ -540,7 +536,7 @@ def main():
         marker_begin = 'BEGIN'
     if not marker_end:
         marker_end = 'END'
-    force = '-f' if force else ''
+    force = True if force else False
```
rexemin, line 539: Can `force` be `None`? If it can be, I get this line, and it can stay. If not, could we remove it?

rexemin (changes requested):
I'm requesting really minor changes, the change looks good. Just as Rich said, if you could label the tests that depend on ZOAU fixing a bug so we can remember to enable them again later, that would be cool. Specially if we have the JIRA ticket in the comment.

plugins/modules/zos_blockinfile.py
AndreMarcel99: By the way we use, is required to get a value

AndreMarcel99:
Captura de pantalla 2024-02-26 a la(s) 3 54 50 p m
Captura de pantalla 2024-02-26 a la(s) 3 55 09 p m

plugins/modules/zos_blockinfile.py
fernandofloresg: @AndreMarcel99 didn't understand. Could you be more descriptive? I think you are referring that force does not have a default value here, hence you force assign one. But based on line 494 ` force=dict(arg_type='bool', default=False, required=False),` it gets assigned the value of False then is captured in the force vars by `force = parsed_args.get('force')`. 

If that is correct I think is removable right?

plugins/modules/zos_blockinfile.py
```diff
@@ -574,42 +570,30 @@ def main():
...
-
-        result['cmd'] = ret['cmd']
-        result['changed'] = ret['changed']
-        result['found'] = ret['found']
-        # Only return 'rc' if stderr is not empty to not fail the playbook run in a nomatch case
-        # That information will be given with 'changed' and 'found'
-        if len(stderr):
-            result['stderr'] = str(stderr)
-            result['rc'] = rc
+        ret = json.loads("""{0}""".format(stdout))
```
fernandofloresg, line 582: I don't think there is a need to use a multiline string and format here. We could just use `json.loads(stdout)`?

tests/functional/modules/test_zos_blockinfile_func.py
```diff
@@ -862,24 +863,24 @@ def test_uss_block_insert_with_indentation_level_specified(ansible_zos_module):
...
-        params["path"] = full_path
-        results = hosts.all.zos_blockinfile(**params)
-        for result in results.contacted.values():
-            assert result.get("changed") == 1
-        results = hosts.all.shell(cmd="cat {0}".format(params["path"]))
-        for result in results.contacted.values():
-            assert result.get("stdout") == EXPECTED_DOUBLE_QUOTES
-    finally:
-        remove_uss_environment(ansible_zos_module)
+# Test case base on bug of dataset.blockifile
```
fernandofloresg, line 866: Do you have the validation task number (validation as the GH issue we will use to retest in the future) ? That is bc even if this states we need to fix it is not easy to identify which issue are we talking about.

fernandofloresg: You can add the ZOAU issue number or GH validation task number

tests/functional/modules/test_zos_lineinfile_func.py
```diff
@@ -675,92 +675,96 @@ def test_ds_line_replace_match_insertbefore_ignore(ansible_zos_module, dstype):
...
-        params["path"] = ds_full_name
-        results = hosts.all.zos_lineinfile(**params)
-        for result in results.contacted.values():
-            assert result.get("changed") == 1
-        results = hosts.all.shell(cmd="cat \"//'{0}'\" ".format(params["path"]))
-        for result in results.contacted.values():
-            assert result.get("stdout") == EXPECTED_REPLACE_NOMATCH_INSERTBEFORE_NOMATCH
-    finally:
-        remove_ds_environment(ansible_zos_module, ds_name)
+#@pytest.mark.ds
```
fernandofloresg, line 678: Is just the same comment as the blockinfile test cases, better to add the validation number in this same PR to not forget about coming back and knowing exactly which test cases to validate.

fernandofloresg (changes requested):
Changes are really minor, no need to run regression again just validate that the change for blockinfile does not break usability.

plugins/modules/zos_blockinfile.py
AndreMarcel99: In this case as we aren't escaping any special character the """ """ allow us to pass values as _ / ' to the python API so is required

AndreMarcel99:
Captura de pantalla 2024-02-26 a la(s) 5 22 09 p m
Captura de pantalla 2024-02-26 a la(s) 5 22 21 p m

plugins/modules/zos_blockinfile.py
fernandofloresg: Got it! would be nice to have it as a comment

fernandofloresg (approved):
This looks good for me. Tests removed have the GH issue added to them so we can later enable them. One thing we are missing is to add these two issues into the known issues entry.

plugins/modules/zos_blockinfile.py
rexemin: I'm actually not sure what happens when `default` and `required` are both defined, Ansible wants us to only use one or the other, so I don't know the precedence they have. If `default` is honored, then yes, we could remove it. If not, it's just like Marcel said, we need this other line so `force` actually has a value.

AndreMarcel99: I remove because the default value take care about it, Fernando is right

rexemin (approved):

Issue #1258, opened 2024-02-26
[Bug]When zos_blockinfile tries to introduce a block with double quotes generates a false negative


AndreMarcel99:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Bug description

Using double quotes in text block to insert in ZOAU datasets.blockinfile resulted in an error. 

```python
>>> a = 'some " quotes " '
>>> rc = datasets.blockinfile('omvsadm.test.seq', state=True, block=a, ins
ert_after='eof', as_json=True)
>>> print(rc)
[ZOAUResponse]
	rc: 8
	response_format: UTF-8
	stdout_response:
	stderr_response: BGYSC5309E Failed to insert new entry.
BGYSC5306E Failed to parse JSON text into array.
BGYSC4820E Failed to process JSON output.

	command: dmod -b -j  -s -e "/eof/a\some \" quotes \" /$" -e "$ a\some \" quotes \" " "omvsadm.test.seq"
```

This was fixed in ZOAU 1.3.1 and now that result when `as_json` is true is no longer failing.

```python
>>> print(rc)
[ZOAUResponse]
	rc: 0
	response_format: UTF-8
	stdout_response:
	stderr_response:
	command: dmod -b  -s -e "/eof/a\some \" quotes \" /$" -e "$ a\some \" quotes \" " "omvsadm.test.seq"
```

Some workaround code was added to not fail these, but that involves calling the shell utilities, now we want to remove this from the code in an effort to reduce technical debt.

When any option on the module such as block or marker that contain double quotes `" ` on the te content of those options. The module will be correctly execute, but when returning the response on json format on [line](https://github.com/ansible-collections/ibm_zos_core/blob/6e84b32c01b8eed7817743679959aa867bfff3eb/plugins/modules/zos_blockinfile.py#L834) will fail because is not properly scape some characters.

The functions were created as work around, with the solution by ZOAU we won't need these functions.
```python
 def execute_dmod(src, block, marker, force, encoding, state, module, ins_bef=None, ins_aft=None):
    """Execute in terminal dmod command directly.

    Parameters
    ----------
    src : str
        The z/OS USS file or data set to modify.
    block : str
        The block to insert/replace into the src.
    marker : str
        The block will be inserted/updated with the markers.
    force : bool
        If not empty passes True option to dmod cmd.
    encoding : str
        Encoding of the src.
    state : bool
        Determine if will add or delete the block.
    module : obj
        Object to execute the command.
[...]

PR #1284, opened 2024-03-06
Add_at_the_documentation_about_known_issues

AndreMarcel99:
##### SUMMARY
Add in notes about documentation about known issues

##### ISSUE TYPE
- Docs Pull Request

Captura de pantalla 2024-03-06 a la(s) 12 02 50 p m


richp405 (approved):

plugins/modules/zos_blockinfile.py
```diff
@@ -185,6 +185,9 @@
     the block will be overwritten on each iteration.
   - When more then one block should be handled in a file you must change
     the I(marker) per task.
+  - If using ZOAU version 1.3.0 and double quotes (") in the block argument, the module
+    will throw a false negative response. Follow up on the
+    L(issue in the collection's repository,https://github.com/ansible-collections/ibm_zos_core/issues/1258).
```
ketankelkar, line 190: Perhaps the note here can be expanded upon to include some additional details. Maybe something like:
> With ZOAU version 1.3.0,  if double quotes (") are used in the I(block) parameter, the module
    fails with an error message containing: 'return content is NOT in json format'. This is currently a false negative response, where the I(src) is changed, but an error is reported anyhow. Follow up on the issue tracked L(here,https://github.com/ansible-collections/ibm_zos_core/issues/1258).

Feel free to re-reword what I've got, my point is just to add more info about the behavior for the user.

plugins/modules/zos_lineinfile.py
```diff
@@ -195,6 +195,9 @@
     needs to be encoded, it should be cataloged first.
   - For supported character sets used to encode data, refer to the
     L(documentation,https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/resources/character_set.html).
+  - When working with an MVS file and ZOAU version 1.3.0, combining C(regexp)
```
ketankelkar, line 198: I'm not sure if `MVS file` is an official z/OS term, if it's data sets, maybe use `data sets`? Valid data sets are already defined above, so we probably don't need to re-specify them

>When working with data sets and ZOAU version 1.3.0, combining ...

Also, "throw an error" is more of a developer-y word to me, maybe we can say the "module fails or something like that?

Maybe:
> When working with ZOAU version 1.3.0, the module fails when C(src) is a data set and either C(insertafter) or C(insertbefore) is set in addition to C(regexp).

^ I like having the ZOAU version right at the beginning so if anyone is on a different version, they dont have to bother reading the rest of the sentence

PR #1285, opened 2024-03-06
[v1.10.0] [Enabler] [zos_data_set] Enabler/992/fixsanityfourin110

richp405:
SUMMARY
This ticket is for the forward-part of the zos_data_set module for the new 1.3.0 development work.

Changes will resolving sanity test linting issues #992, #993, #994

ISSUE TYPE
Bugfix Pull Request
COMPONENT NAME
zos_data_set

### DIFF
--- a/changelogs/fragments/992-fix-sanity4to6.yml
+++ b/changelogs/fragments/992-fix-sanity4to6.yml
@@ -1,3 +1,7 @@
 trivial:
   - zos_data_set.py - Corrected references to input variable definitions
-    (https://github.com/ansible-collections/ibm_zos_core/pull/xxxx).
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1285).
+  - data_set.py - Updated exception handler to match what was returned.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1285).
+  - test_zos_data_set_func.py - Removed test of discontinued function.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1285).
--- a/changelogs/fragments/992-fix-sanity4to6.yml
+++ b/changelogs/fragments/992-fix-sanity4to6.yml
@@ -1,0 +1,3 @@
+trivial:
+  - zos_data_set.py - Corrected references to input variable definitions
+    (https://github.com/ansible-collections/ibm_zos_core/pull/xxxx).
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1025,8 +1025,6 @@ class DataSet(object):
         except exceptions.ZOAUException as create_exception:
             raise DatasetCreateError(
                 name,
-                # create_exception.response.rc,
-                # create_exception.response.stdout_response + create_exception.response.stderr_response
                 create_exception.rc,
                 create_exception.stdout_response + create_exception.stderr_response
             )
@@ -1037,8 +1035,7 @@ class DataSet(object):
                     return 0
             raise DatasetCreateError(
                 name,
-                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU." +
-                str(formatted_args) + "--rf originally=" + record_format + ".",
+                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU.",
             )
         # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
         # response.rc now we just return 0 if nothing failed
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1039,7 +1039,7 @@ class DataSet(object):
             raise DatasetCreateError(
                 name,
                 msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU." +
-                str(formatted_args) + e.stdout_response + e.stderr_response,
+                str(formatted_args),
             )
         # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
         # response.rc now we just return 0 if nothing failed
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1032,9 +1032,10 @@ class DataSet(object):
             )
         except exceptions.DatasetVerificationError as e:
             # verification of a data set spanning multiple volumes is currently broken in ZOAU v.1.3.0
-            if len(volumes) > 1:
-                if DataSet.data_set_cataloged(name, volumes):
-                    return 0
+            # - note that you can't check length on a potential NoneType
+            # if len(volumes) > 1:
+            if DataSet.data_set_cataloged(name, volumes):
+                return 0
             raise DatasetCreateError(
                 name,
                 msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU.",
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1038,7 +1038,8 @@ class DataSet(object):
                 return 0
             raise DatasetCreateError(
                 name,
-                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU.",
+                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU." +
+                str(formatted_args) + e.stdout_response + e.stderr_response,
             )
         # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
         # response.rc now we just return 0 if nothing failed
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1018,8 +1018,10 @@ class DataSet(object):
         except (exceptions.ZOAUException, exceptions.DatasetVerificationError) as create_exception:
             raise DatasetCreateError(
                 name,
-                create_exception.response.rc,
-                create_exception.response.stdout_response + create_exception.response.stderr_response
+                # create_exception.response.rc,
+                # create_exception.response.stdout_response + create_exception.response.stderr_response
+                create_exception.rc,
+                create_exception.stdout_response + create_exception.stderr_response
             )
         # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
         # response.rc now we just return 0 if nothing failed
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1039,7 +1039,7 @@ class DataSet(object):
             raise DatasetCreateError(
                 name,
                 msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU." +
-                str(formatted_args),
+                str(formatted_args) + "--rf originally=" + record_format + ".",
             )
         # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
         # response.rc now we just return 0 if nothing failed
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -1022,11 +1022,11 @@ class DataSet(object):
         formatted_args = DataSet._build_zoau_args(**original_args)
         try:
             datasets.create(**formatted_args)
-        except exceptions.ZOAUException as create_exception:
+        except exceptions._ZOAUExtendableException as create_exception:
             raise DatasetCreateError(
                 name,
-                create_exception.rc,
-                create_exception.stdout_response + create_exception.stderr_response
+                create_exception.response.rc,
+                create_exception.response.stdout_response + "\n" + create_exception.response.stderr_response
             )
         except exceptions.DatasetVerificationError as e:
             # verification of a data set spanning multiple volumes is currently broken in ZOAU v.1.3.0
--- a/plugins/modules/zos_data_set.py
+++ b/plugins/modules/zos_data_set.py
@@ -134,8 +134,12 @@ options:
     type: int
     required: false
     default: 5
-    aliases:
-      - size
+  size:
+    description:
+      - A compound value with size and units in the same string.
+      - This will split into space_primary and space_type.
+    type: str
+    required: false
   space_secondary:
     description:
       - The amount of secondary space to allocate for the dataset.
@@ -401,8 +405,12 @@ options:
         type: int
         required: false
         default: 5
-        aliases:
-          - size
+      size:
+        description:
+          - A compound value with size and units in the same string.
+          - This will split into space_primary and space_type.
+        type: str
+        required: false
       space_secondary:
         description:
           - The amount of secondary space to allocate for the dataset.
@@ -1011,15 +1019,9 @@ def fix_old_size_arg(params):
                     str(params.get("size"))
                 )
             )
-    if params.get("space_primary"):
-        match = re.fullmatch(
-            r"([1-9][0-9]*)(M|G|K|TRK|CYL)",
-            str(params.get("space_primary")),
-            re.IGNORECASE,
-        )
-    if match:
-        params["space_primary"] = int(match.group(1))
-        params["space_type"] = match.group(2)
+        else:
+            params["space_primary"] = int(match.group(1))
+            params["space_type"] = match.group(2).upper()
     return params
 
 
@@ -1056,6 +1058,7 @@ def parse_and_validate_args(params):
                     default="M",
                 ),
                 space_primary=dict(type="int", required=False, dependencies=["state"]),
+                size=dict(type="str", required=False)
                 space_secondary=dict(
                     type="int", required=False, dependencies=["state"]
                 ),
@@ -1145,6 +1148,7 @@ def parse_and_validate_args(params):
             default="M",
         ),
         space_primary=dict(type="int", required=False, dependencies=["state"]),
+        size=dict(type="str", required=False)
         space_secondary=dict(type="int", required=False, dependencies=["state"]),
         record_format=dict(
             type=record_format,
@@ -1261,7 +1265,8 @@ def run_module():
                     default="M",
                     choices=["K", "M", "G", "CYL", "TRK"],
                 ),
-                space_primary=dict(type="int", required=False, aliases=["size"], default=5),
+                space_primary=dict(type="int", required=False, default=5),
+                size=dict(type="str", required=False)
                 space_secondary=dict(type="int", required=False, default=3),
                 record_format=dict(
                     type="str",
@@ -1326,7 +1331,8 @@ def run_module():
             default="M",
             choices=["K", "M", "G", "CYL", "TRK"],
         ),
-        space_primary=dict(type="int", required=False, aliases=["size"], default=5),
+        space_primary=dict(type="int", required=False, default=5),
+        size=dict(type="str", required=False)
         space_secondary=dict(type="int", required=False, default=3),
         record_format=dict(
             type="str",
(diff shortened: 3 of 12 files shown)


richp405:
image

plugins/module_utils/data_set.py
```diff
@@ -1025,8 +1025,10 @@ def create(
         except exceptions.ZOAUException as create_exception:
             raise DatasetCreateError(
                 name,
-                create_exception.response.rc,
-                create_exception.response.stdout_response + create_exception.response.stderr_response
+                # create_exception.response.rc,
```
fernandofloresg, line 1028: Hey Rich, we can remove this instead of commenting them, those have been recently added so no problem, but before doing that. I'm curious about removing the response access, Did you find out that the structure at ZOAU docs https://www.ibm.com/docs/en/zoau/1.3.x?topic=apis-exceptions, does not align with the python ZOAUException structure ? Meaning, ZOAUException is not having a response attribute?

changelogs/fragments/992-fix-sanity4to6.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
+  - zos_data_set.py - Corrected references to input variable definitions
+    (https://github.com/ansible-collections/ibm_zos_core/pull/xxxx).
```
fernandofloresg, line 3: Just need to add the PR number here.

plugins/module_utils/data_set.py
```diff
@@ -1035,7 +1037,8 @@ def create(
                     return 0
             raise DatasetCreateError(
                 name,
-                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU.",
+                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU." +
+                str(formatted_args) + "--rf originally=" + record_format + ".",
```
fernandofloresg, line 1041: Is this a new error message or just part of the debugging print process ? If this is going to be customer-facing I think we could rephrase it.

plugins/modules/zos_data_set.py
```diff
@@ -24,7 +24,9 @@
 description:
   - Create, delete and set attributes of data sets.
   - When forcing data set replacement, contents will not be preserved.
-author: "Blake Becker (@blakeinate)"
+author:
+  - "Blake Becker (@blakeinate)"
+  - "Rich Parker (@richp405)"
```
fernandofloresg, line 29: Yeah you deserve to be up there after dealing with so many showstoppers.

plugins/modules/zos_data_set.py
```diff
@@ -109,7 +111,7 @@
     description:
       - The data set type to be used when creating a data set. (e.g C(pdse))
       - C(MEMBER) expects to be used with an existing partitioned data set.
-      - Choices are case-insensitive.
+      - Choices are case-sensitive.
```
fernandofloresg, line 114: I think we should create a review item for playbook repository, just to correct any misleading examples.

fernandofloresg: I see this is the case for many different options, so definitely needed.

fernandofloresg (changes requested):

plugins/module_utils/data_set.py
richp405: I'll remove them... yes, the zoau exception returned does not contain 'response' as an element, which crashes things.

changelogs/fragments/992-fix-sanity4to6.yml
richp405: argh!!! Knew I forgot something

plugins/module_utils/data_set.py
richp405: I added the originally section for debugging.  I have removed it.  Good catch!

plugins/modules/zos_data_set.py
richp405: Thank you!

plugins/modules/zos_data_set.py
richp405: We got stuck, because we have to use choices, which are case-sensitive.  Let's mention it to Ketan during stand up, since he's going to check other things in the playbook repository.

Really, I'd like to 'invent' a way to make choices case-insensitive, but didn't see a safe way to do it.

plugins/module_utils/data_set.py
```diff
@@ -1025,8 +1025,8 @@ def create(
         except exceptions.ZOAUException as create_exception:
             raise DatasetCreateError(
                 name,
-                create_exception.response.rc,
-                create_exception.response.stdout_response + create_exception.response.stderr_response
+                create_exception.rc,
+                create_exception.stdout_response + create_exception.stderr_response
```
ketankelkar, line 1029: Curious what triggered this change, from the datasets.py code (internal ZOAU Python repo):
```

# Exceptions from ZOAU Core Calls
class _ZOAUExtendableException(Exception):
    """Custom exception supporting a string message and ZOAUResponse object
    Attributes
    ----------
    message: str
        ZOAUResponse as string (Printed as output by `raise`)
    response : ZOAUResponse
        Addressable ZOAUResponse object
    Notes
    -----
    `ZOAUResponse` attributes
        - rc: str
            Return code
        - stdout_response: str
            Stdout
        - stderr_response: str
            Stderr
        - command: str
            Command performed during a ZOAU core call
        - response_format: str
            Response encoding
    """

    message: str
    response: ZOAUResponse

    def __init__(self, response_obj: ZOAUResponse):
        super().__init__(str(response_obj))
        self.response = response_obj


class ZOAUException(_ZOAUExtendableException):
    """Errors happening during ZOAU core calls, see message for details"""

```
It appears that all the response data (`rc`, `response_stdout`, `response_stderr`) are inside of a `ZOAUResponse` object, which is folded into the `self.response`. So I would think that getting at those would require the `create_exception.response.<item>` where the `self` in this block would be the Exception variable `create_exception`.

In a (very) rough re-create, I also found that we should be `except`-ing a `_ZOAUExtendableException` rather than a `ZOAUException` as ZOAU's internal exception inheritance seems to have undergone a re-structuring where many of the Exceptions now seem to extend the `_ZOAUExtendableException`. 

(This should have ideally been a part of the module migration work, but it appears it was missed, oops) -- I can open up a separate work item to encapsulate/track all this if the changes and validation seem large.

The recreate induces a `DatasetWriteException` which is not triggered from the `datasets.create()` function, but I found it simpler to trigger this error to play around with rather than an error from the create function. I figure they're similar enough since they both have similar definitions and both extend the same `_ZOAUExtendableException` class.

Rough re-create:
```python3
>>> from zoautil_py import datasets, exceptions
>>> 
>>> datasets.write("some.data.set", "helllothere")                                        
0
```
The above should create and write a line to the data set. From here, I opened up the data set in ISPF (`=3.4`) and browsed (`b`) it. Without forcing the write, I could trigger a `DatasetWriteException`:
```python3
>>> try:                                                           
...     datasets.write("some.data.set", "helllothere")             
... except exceptions.ZOAUException as create_exception: 
...     print('hello from ZOAUException')
... except:
...     print('something else')
... 
something else
``` 

This suggested to me that we're capturing the wrong exception. Trying with 
```python3
>>> try:                                                           
...     datasets.write("some.data.set", "helllothere")             
... except exceptions._ZOAUExtendableException as create_exception: 
...     print('hello from ZOAUException')
... except:
...     print('something else')
... 
hello from ZOAUException
``` 

And finally, the contents of the exception:
```python3
>>> try:                                                           
...     datasets.write("some.data.set", "helllothere")             
... except exceptions._ZOAUExtendableException as create_exception:
...     st = str(create_exception.rc) + create_exception.stderr_response
...     print('hi', st)                                            
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/zoau/v1.3.0/lib/3.11/zoautil_py/datasets.py", line 554, in write
    raise DatasetWriteException(response)
zoautil_py.exceptions.DatasetWriteException: [ZOAUResponse]
	rc: 8
	response_format: UTF-8
	stdout_response: 
	stderr_response: BGYSC0705E Failed to open dataset SOME.DATA.SET for write.

	command: decho -- "helllothere" "some.data.set"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
AttributeError: 'DatasetWriteException' object has no attribute 'rc'
>>> 
```
The final line above suggested to me that the `rc`, `stdout` and `stderr` are nested in the `response` attribute:
```python3
>>> try:                                                                                                                              
...     datasets.write("some.data.set", "helllothere")                                                                                
... except exceptions._ZOAUExtendableException as create_exception:                                                                   
...     st = str(create_exception.response.rc) + create_exception.response.stderr_response
...     print('hi', st)                                                                                                               
... 
hi 8BGYSC0705E Failed to open dataset SOME.DATA.SET for write.
```

plugins/modules/zos_data_set.py
ketankelkar: I've expanded the scope of this PR - https://github.com/IBM/z_ansible_collections_samples/pull/246

plugins/modules/zos_data_set.py
```diff
@@ -1179,7 +1182,7 @@ def parse_and_validate_args(params):
             # ["batch", "space_type"],
             # ["batch", "space_primary"],
             # ["batch", "space_secondary"],
-            ["batch", "record_format"],
+            # ["batch", "record_format"],
```
ketankelkar, line 1185: Does this open up a path for ambiguous behavior where `record_format` is defined both inside of a task in `batch` as well as outside?

```yml
- name: Create multiple data sets.
    zos_data_set:
---> record_format: FBA     <------ambiguous??
        batch:
            - name: SOME.DATA.SET
                state: present
                type: PDS
                record_format: FB
                record_length: 80
            - name: ANOTHER.DATA.SET
                state: present
                type: PDS
                record_format: VB
                record_length: 80
```

What is record format of `some.data.set` and `another.data.set`? Are they both `FBA`? Are they `FB` and `VB` respectively? 

With `["batch", "record_format"]` being mutually exclusive, presumably the above task would have failed. But with the new change I'm seeing where the line is commented out (implying those two items are no longer mutually exclusive), my concern is that with such a playbook task as given above, there will be ambiguous behavior. 

From a convenience perspective, maybe not having to define `record_format` for each task in batch and defining it only at the higher level is advantageous. 
For example, say I want to create `SOME.DATA.SET1`, `SOME.DATA.SET2` and `SOME.DATA.SET3` as `FBA` and `ANOTHER.DATA.SET` as `VB`, perhaps I could define a sort of 'default' record format at the higher level and then if record format is differently defined at the nested level, that would override the higher level default:

```yml
- name: Create multiple data sets.
    zos_data_set:
      record_format: FBA
        batch:
            - name: SOME.DATA.SET1
                state: present
                type: PDS
                record_length: 80
            - name: SOME.DATA.SET2
                state: present
                type: PDS
                record_length: 80
            - name: SOME.DATA.SET3
                state: present
                type: PDS
                record_length: 80
            - name: ANOTHER.DATA.SET
                state: present
                type: PDS
----->         record_format: VB   <------ override higher level 'default'
                record_length: 80
```

If so, this would need to be documented and further validated.

plugins/module_utils/data_set.py
richp405: There are 2 exceptions in that segment of code: exceptions.ZOAUException, and exception.DatasetCreationError.  Where I'm a little confused is how to make this better: ZOAUException does not pass .response... I can switch the first on to extendable... where I'm uncertain is should this also apply to DatasetCreationError: should it just go away, stay as is, or what?  I could use some ideas on this one.

plugins/modules/zos_data_set.py
richp405: Actually, it's backwards: 
If a batch is used, ONLY the record_format for the batch entries are used.
If a batch is NOT used, then only the higher (outside of batch) is used, and is a single-shot.

This is because of 2 contradictory requirements: to add choices means NULL/None is not a valid choice, and to use Batch means the outer record_format isn't used.  So, I have it clear the record_format AFTER the validation, based on whether batch is populated or not.

Also, I had to removed the 'exclusive' setting to handle this paradox (record format cannot be blank but does not exist if batch is filled it.)

richp405: I never realized this wasn't described well in the documentation.  I'll think on it tonight, but there are 2 sets of essentially duplicate settings:

In non-batch/1-shot mode, only outer settings are used (name, type, space type, etc)
In batch mode, settings are completely contained within on batch suboption. (so one can be fb sized in megs and another can be vb sized in tracks and another can be pdse without initial size).


It made sense to me after mucking in the code in this quarter, but it probably needs to be clarified in the actual doc.

plugins/module_utils/data_set.py
ketankelkar: I think the code should become:
```python3
  <...>
        try:
            datasets.create(**formatted_args)
        except exceptions._ZOAUExtendableException as create_exception:
            raise DatasetCreateError(
                name,
                create_exception.response.rc,
                create_exception.response.stdout_response + create_exception.response.stderr_response
            )
        except exceptions.DatasetVerificationError as e:
            # verification of a data set spanning multiple volumes is currently broken in ZOAU v.1.3.0
            if volumes and len(volumes) > 1:
                if DataSet.data_set_cataloged(name, volumes):
                    return 0
            raise DatasetCreateError(
                name,
                msg="Unable to verify the data set was created. Received DatasetVerificationError from ZOAU.",
            )
        # With ZOAU 1.3 we switched from getting a ZOAUResponse obj to a Dataset obj, previously we returned
        # response.rc now we just return 0 if nothing failed
        return 0
```

The way I see it, `datasets.create()` can do 1 of 3 things:
1. succeed (yay!)
2. throw an exception which extends `_ZOAUExtendableException`
3. throw an exception which does not extend `_ZOAUExtendableException` merely extends the built-in base `Exception` class

scenario 1 is straightforward.

Scenarios 2 and 3 are non-obvious because in ZOAU python API documentation, it is not mentioned that these *two* distinct categories of exceptions exist. They even state in their docs that all their exceptions extend from `ZOAUException`, which isn't exactly right as of right now.
image


My understanding of the reasoning behind their decision is that they're still determining what information to use to construct those exceptions which do not `_ZOAUExtendableException`. 

I've been referencing the code in their IBM internal repo for their python API -- specifically the `exceptions.py` file, where the exceptions are defined:

```python3
class DatasetCreateException(_ZOAUExtendableException):
    """Errors while trying to create the dataset, see message for details"""
```
```python3
class DatasetVerificationError(Exception):
    pass
```
The `datasets.create()` function we (Ansible) call may raise either of these two exceptions and each gets handled differently.

Also, mainly because I just got slightly dizzy thinking about it, `DatasetCreateException` is defined by ZOAU and `DatasetCreateError` is defined by us (Ansible).

We `raise` a `DatasetCreateError` which is a formatted string `msg`, which we populate based on what we receive from the call to ZOAU via `datasets.create()`. What we receive could be `rc`/`stdout`/`stderr` packaged inside a ZOAUResponse object if the exception extends `_ZOAUExtendableException` or we could get an error message string if we catch something like a `DatasetVerificationError` which just sends out the data set name.

I somewhat recently modified the definition of `DatasetCreateError` so that the string we're constructing may or may not include an RC. This way, for a `DatasetVerificationError`, we do not have to "make up" our own RC and still get the message out to the user.

richp405: I think I understand where the disconnect was.  I'll re-run the new code through the pipeline.

richp405:
Rebuilt with Ketan's discovered (better) exception handler.  Pipeline passed: 
image

plugins/module_utils/data_set.py
fernandofloresg: Great analysis Ketan, it was unclear for me too and now I get it

plugins/modules/zos_data_set.py
ketankelkar: You were right, looking at the code plus your explanation helped clarify this. The initial delta is why I raised a flag here, but it looks like this won't introduce a completely _new_ behavior. The same way a `record_format` param is now "valid" syntax-wise outside of a batch operation, we already had some others - `space_primary`, `space_type`). 

So the params outside the `batch` call are accepted by the Ansible parser but ignored by our logic. I see this is the way to get those "choices" defined to the ansible parser. 

```yml
- name: create data set
  ibm.ibm_zos_core.zos_data_set:
    space_primary: 100
    space_type: TRK
    record_format: FBA
    batch:
      - name: "some.data.set.one"
        type: SEQ
        record_format: VB
        state: present
        space_primary: 50
        space_type: TRK
      - name: "some.data.set.two"
        type: SEQ
        state: present
        record_format: VB
        space_primary: 50
        space_type: TRK
```

The above is valid syntax, but `space_primary: 100`, `space_type: TRK`, and `record_format: FBA` are ignored. Rather the values within the individual batch items are honored. 

image


Since this is not completely new behavior, I think it's this change is fine.
Ideally, those options outside of the batch would raise an error, but that's an exploration for another day.

fernandofloresg (approved):

ketankelkar (approved):
Nice work here Rich!! It's awesome to see those ignore files reducing in size. 😍 
Also thanks for helping make our error handling tighter 💪 

I pulled this branch and ran the below playbook task to confirm the changes in the code path for error handling in data set creation. 

The task requests a data set to be created with a name that's too long:
```yml
- name: create data set
  ibm.ibm_zos_core.zos_data_set:
    name: "somesome.datadata.set1set1.somesome.datadata.set2set2"
    type: SEQ
    space_primary: 5
    space_type: M
    replace: yes
```

Getting latest from branch on git:
```bash
(ansible-2.16) ketan@Ketans-MBP ibm_zos_core % git branch --show-current
Enabler/992/fixsanityfourin110

(ansible-2.16) ketan@Ketans-MBP ibm_zos_core % git pull
From github.com:ansible-collections/ibm_zos_core
 * [new branch]        Enabler/992/fixsanityfourin110 -> origin/Enabler/992/fixsanityfourin110
Already up to date.
```
build and install this version:
```bash
(ansible-2.16) ketan@Ketans-MBP ibm_zos_core % ansible-galaxy collection build -f .
Created collection for ibm.ibm_zos_core at /Users/ketan/dev/ibm_zos_core/ibm-ibm_zos_core-1.9.0-beta.1.tar.gz

(ansible-2.16) ketan@Ketans-MBP ibm_zos_core % ansible-galaxy collection install -f /Users/ketan/dev/ibm_zos_core/ibm-ibm_zos_core-1.9.0-beta.1.tar.gz
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Installing 'ibm.ibm_zos_core:1.9.0-beta.1' to '/Users/ketan/.ansible/collections/ansible_collections/ibm/ibm_zos_core'
ibm.ibm_zos_core:1.9.0-beta.1 was installed successfully
```
Run playbook:
```bash
(ansible-2.16) ketan@Ketans-MBP ibm_zos_core % ansible-playbook -i ~/inventory ~/dev/playbooks/recreate.yml

PLAY [ZOAU130] ***************************************************************************************************************************************************************************************************************************

TASK [create data set] *******************************************************************************************************************************************************************************************************************
fatal: [zvm]: FAILED! => {"changed": false, "message": "", "msg": "DatasetCreateError('An error occurred during creation of data set \"SOMESOME.DATADATA.SET1SET1.SOMESOME.DATADATA.SET2SET2\". RC=8, \\nBGYSC1401E Dataset name too long SOMESOME.DATADATA.SET1SET1.SOMESOME.DATADATA.SET2SET2.\\n')", "names": ["SOMESOME.DATADATA.SET1SET1.SOMESOME.DATADATA.SET2SET2"]}

PLAY RECAP *******************************************************************************************************************************************************************************************************************************
zvm                        : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

And indeed a DatasetCreateError is raised and the expected RC and system messages are cleanly propagated up to the user.

PR #1340, opened 2024-03-25
[Bugfix][1301]Work_around_fix_false_positive

### LINKED ISSUE #1301: [Bug] [1.10.0] [zos_blockinfile] Work-Around to fix false negative when 'block' parameter contains a double quote (")

### Bug description

Reminder to review/revise the doc changes made in #1284 so they are up to date with whatever changes this bugfix implements. 

Related issues:
- #1258 

The above is a tracker for the fix going into ZOAU. This will be the issue that provides the work-around until the dependency in the above issue is completed.

This behavior is a known issue stemming from ZOAU v1.3.0 `dmod`/`dsed`, specifically the JSON parser. 
This is worth changing because currently, the module *changes* the `src` and then reports a `failed` module. 
This is an issue of data integrity. 

Futhermore, we do not know if only the double quote (") character is affected or there are others too. It could potentially be all the characters which are treated as special in the JSON RFC.

playbook task:
```yml
- name: blockinfile
  ibm.ibm_zos_core.zos_blockinfile:
    src: "some.data.set"
    block: JK"L
    insertafter: EOF 
```

result(s):

The playbook task is reported as `failed=True` and `changed=False`.
```sh
TASK [blockinfile] *****************************************************************************************************************************************************************************************************************************************************************
fatal: [zvm]: FAILED! => {"changed": false, "msg": "ZOAU dmod return content is NOT in json format", "rc": 8, "stderr": "BGYSC5309E Failed to insert new entry.\nBGYSC5306E Failed to parse JSON text into array.\nBGYSC4820E Failed to process JSON output.\n", "stderr_lines": ["BGYSC5309E Failed to insert new entry.", "BGYSC5306E Failed to parse JSON text into array.", "BGYSC4820E Failed to process JSON output."], "stdout": "", "stdout_lines": []}
```
But, ssh-ing in and dumping the data set shows the changes were made.
before:
```sh
bash-4.3$ dcat 'some.data.set'
"ABC"
"DEF"
"GHI"
```
after:
```sh
bash-4.3$ dcat 'some.data.set'
"ABC"                                                                           
"DEF"                                                                           
"GHI"                                                                           
# BEGIN ANSIBLE MANAGED BLOCK                                                   
JK"L                                                                            
# END ANSIBLE MANAGED BLOCK 
```
I see several options worth exploring as a temporary work-around until the underlying issue is fixed.

The preferred:
Disable the JSON flag during the call to the ZOAU functionalities (`datasets.blockinfile`)and then capture/parse the stdout/stderr. This should be possible via the ZOAU python interfaces and is preferable, but if not, `run_command` command can be used to call `dmod`/`dsed` directly. 

Since we do not know which characters work and which don't, it may be advisable to revert to the non-JSON for all calls to `blockinfile` rather than limiting the work-around code to only those where the `block` param contains double quotes.

Less preferable:
Detect the presence of double quote (and potentially other 'forbidden' chars) and fail the module with an informative message. 
This option limits usability and is therefore less preferable. Further, we do not know which characters work and which ones do not.

Least preferable:
Capture the exception, inspect the file to verify the changes ...
This would be a ton of work, expensive both for development as well as per task. 

v1.10.0-beta.1

v1.3.0

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

zos_blockinfile

### Playbook:
```yml
    - name: create data set
      ibm.ibm_zos_core.zos_data_set:
        name: "some.data.set"
        type: SEQ
        space_primary: 10
        space_type: M
        replace: yes

    - name: write to some data set
      ibm.ibm_zos_core.zos_copy:
        content: |
          ABC
          DEF
          GHI
        dest: "some.data.set"
        force: yes

    - name: add string with double quote to end of file
      ibm.ibm_zos_core.zos_blockinfile:
        src: "some.data.set"
        block: 'JK"L'
        insertafter: EOF

```

AndreMarcel99:
##### SUMMARY
Make a work around about the know issue using (") by calling the system

Fixes #1301 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
A call to shell by dmod creating the command and then clean the response.

##### ADDITIONAL INFORMATION
Captura de pantalla 2024-03-25 a la(s) 2 58 02 p m

Captura de pantalla 2024-03-25 a la(s) 2 58 15 p m


richp405 (approved):
looks okay from here!

changelogs/fragments/1340-Work_around_fix_false_positive.yml
fernandofloresg, line 1: Would change the`"` symbol to `double quotes` or `double quotation marks`. Now, this is a bug fix but bug is not yet in a published released, even though I would still use `bugfixes` tag for this changelog, I would rephrase this changelog to :

```
Using double quotation marks inside a block resulted in a false positive result with ZOAU 1.3. Fix now handles this special case to avoid false negatives.
```

plugins/modules/zos_blockinfile.py
```diff
@@ -569,26 +610,35 @@ def main():
     # state=present, insert/replace a block with matching regex pattern
     # state=absent, delete blocks with matching regex pattern
     if parsed_args.get('state') == 'present':
-        return_content = present(src, block, marker, ins_aft, ins_bef, encoding, force)
+        if '"' in block:
+            rc, cmd = execute_dmod(src, block, quotedString(marker), force, encoding, module=module, ins_bef=quotedString(ins_bef),
```
fernandofloresg, line 614: I think this would yield an inconsistent result when having these two inputs : 
```
block: 'This is a double quote " '
ins_bef: ' Insert before this " double quote " ' 
```
Than this where I removed the double quotation mark from the block option:

```
block: 'This is a double quote'
ins_bef: ' Insert before this " double quote " ' 
```
Assuming the text in the data set is: 

```
 Insert before this " double quote " 
```
That is because when block contains a `"` symbol it will remove any `"` from `ins_bef`, `ins_aft` and `marker`, but it won't do the same if block does not contain `"`. Could you test this scenario?

plugins/modules/zos_blockinfile.py
```diff
@@ -569,26 +610,35 @@ def main():
...
-    except Exception:
-        messageDict = dict(msg="ZOAU dmod return content is NOT in json format", stdout=str(stdout), stderr=str(stderr), rc=rc)
-        if result.get('backup_name'):
-            messageDict['backup_name'] = result['backup_name']
-        module.fail_json(**messageDict)
-
-    result['cmd'] = ret['data']['commands']
-    result['changed'] = ret['data']['changed']
-    result['found'] = ret['data']['found']
+    if '"' not in block:
```
fernandofloresg, line 624: Could you add a comment around why this if is in place, I know is because of this issue but will be better having it identified

fernandofloresg (changes requested):

plugins/modules/zos_blockinfile.py
fernandofloresg: done

plugins/modules/zos_blockinfile.py
fernandofloresg: resolved

fernandofloresg (approved):
I see my comments were addressed and this looks ok now

richp405:
@AndreMarcel99 I don't know if you saw, but this tripped pylint:
ERROR: Found 1 pylint issue(s) which need to be resolved:
ERROR: plugins/modules/zos_blockinfile.py:660:17: used-before-assignment: Using variable 'return_content' before assignment

ketankelkar:
Added a `Do to Merge` label because this item is not currently passing pipelines, yet because it has approvals, it is technically merge-able. Adding the label to prevent an accidental merge

fernandofloresg:
Screenshot 2024-04-22 at 11 43 36 AM

This passed the tests without modification. Issue were just latency problems with testing machines.

PR #1345, opened 2024-03-26
[Bugfix][1239][zos job submit]max_rc_more_than_0_doesn_not_put_change_as_true

### LINKED ISSUE #1239: [Enhancement] [zos_job_submit] Set "changed=true" when "max_rc" is informed and job's return code is less than or equal "max_rc" value

### Ansible module

zos_job_submit

Hello all,
I was running three different jobs using `zos_job_submit`, informing `max_rc: 4` on all executions.
It was noticed that:

- when the job ended with _return code 0_, the task ended with attribute `changed: true`
- when the job ended with _return code 1_ (which is less than the `max_rc: 4`), the task ended with attribute `changed: false`

Since we are not able to discover if the job being submitted really changed something,
- **Could we have the `changed` attribute set to `true` if the `max_rc` parameter is informed on `zos_job_submit` execution and the job's return code is less than or equal to the informed `max_rc`?**

This way we assume that changes might have occurred if something ended accordingly to the informed max return code.
If the user knows that his job is not changing anything, he could use `changed_when: false` on the task's call.

Thank you!

AndreMarcel99:
##### SUMMARY

Fixes #1239 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Validate argument or max_rc in case exist any value equal or under will be accepted and the module will return a changed True in the json response.

##### ADDITIONAL INFORMATION
Change test case `test_job_submit_max_rc` because the job will return a rc of 8 but the case of  max_rc 12 will be changed true because 8 is under the max_rc

Captura de pantalla 2024-04-02 a la(s) 10 47 56 a m
Captura de pantalla 2024-04-02 a la(s) 10 48 11 a m
Captura de pantalla 2024-04-03 a la(s) 12 43 49 p m
Captura de pantalla 2024-04-03 a la(s) 12 44 00 p m
Captura de pantalla 2024-04-03 a la(s) 4 18 49 p m
Captura de pantalla 2024-04-03 a la(s) 4 19 08 p m


changelogs/fragments/1345-max_rc_more_than_0_doesn_not_put_change_as_true.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - zos_job_submit - job submit when mx rc was given by the user the change value return as false
```
fernandofloresg, line 2: I don't understand this sentence "job submit when mx rc was given by the user the change value return as false
    when the max rc is not 0." could you rephrase it?  Also, this other part "Fix now return a change value of true when the rc is not 0 and
    max rc is above or equal the value the user insert." I understand that "change" value really is "changed" from the response dict right?

plugins/modules/zos_job_submit.py
```diff
@@ -1108,7 +1108,9 @@ def assert_valid_return_code(max_rc, job_rc, ret_code, result):
     # should NOT be 'changed=true' even though the user did override the return code,
     # a non-zero return code means the job did not change anything, so set it as
     # result["chagned"]=False,
-    if job_rc != 0:
+    if max_rc and job_rc > max_rc:
```
fernandofloresg, line 1111: Could you explain this if statement? By reading it I understand that if both `max_rc` and `job_rc` are greater that `max_rc` we return False, but by testing I see this behavior : 

```python
Python 3.10.9 (main, Mar  1 2023, 12:20:14) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> bool(0 and 12 > 8)
False
>>> bool(0 and 8 > 12)
False
>>> bool(4 and 8 > 12)
False
>>> bool(0 and 0 > 12)
False
>>> if 4 and 8 > 12:
...     print("hi")
... else:
...     print("nope")
... 
nope
>>> bool(0 and 0 > 0)
False
>>> bool(0 and 0 > -1)
False
```

I ran multiple tests but I don't see a case where it is True, because if whatever is evaluated at the `and`  section is True (or 1) we need a negative or 0 value in max_rc to be True right? that assumption might be wrong on my end so please explain

fernandofloresg (approved):

plugins/modules/zos_job_submit.py
AndreMarcel99: There are some job that a return code not 0 can be success depends on the user and job, that's why the option of `max_rc`  but in these case if the value is gave (2) and the job returns a value over 2 (>2) is that not work for the user so it fails other case as the rc come 1-2 is a success your examples should be:
```
>>> bool(0 and 12 > 0)
True
>>> bool(0 and 8 > 0)
True
>>> bool(4 and 8 > 4)
True
>>> if 4 and 8 > 4:
...     print("hi")
... else:
...     print("nope")
... 
hi
```
The test `test_job_submit_max_rc` comes with a example, the job return a code of 8 if the `max_rc` is 0 or 4, the rc will be over and set to false, but when `max_rc` is 12  is fine for the user
```
bool(12 and 8 > 12):
False
```

changelogs/fragments/1345-max_rc_more_than_0_doesn_not_put_change_as_true.yml
AndreMarcel99: Yes thanks for the changed value

richp405 (approved):

PR #1388, opened 2024-04-03
[v1.10.0] [Enabler] Standardization of choices in modules

### LINKED ISSUE #1360: [Enabler][all modules] Standardize the contents of the "choices" list for module options in all ibm_zos_core modules.
### Enabler description

Historically, the BetterArgParser was used to accept a pre-defined list of "choices" for a given module option. These choices were not case sensitive. The choices were only defined in the module docs section but not in the module source code because the Ansible parser requires case-sensitivity. The playbook input was pre-processed before passed into the BetterArgParser, which enabled case-insensitivity, but that approach is not available to the built in Ansible options parsing. The mis-match in code and docs resulted in sanity issues, which were added to ignore lists). To remain certified, the initiative was taken to reduce the entries in the ignore lists, which meant defining case *sensitive* choices in the Ansible parser. 

This issue aims to standardize the "choices" defined across the `ibm_zos_core` collection. Since choices will be case-sensitive going forward, the choices should be defined as _lower cased_ in all instances for consistency across the collection.

These will be breaking changes per module, so it is important to profusely document these changes. For now, ensure they are added to the changelog fragments as `major_changes` so that they will make at least make it into the release notes.

The following modules have been identified. These will all need to be changed and regression must be run. The code change will be handled by one person, the team will all contribute to passing regression for each item. 

- [x] zos_apf (doesn't require changes and has no dependencies)
  - [x] successful regression pipeline run
- [x] zos_archive
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_backup_restore
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_blockinfile (doesn't require changes but tests depend on zos_data_set)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_copy
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_data_set
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_find (doesn't require changes but tests depend on zos_data_set)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_job_submit **
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_lineinfile (doesn't require changes but tests depend on zos_data_set)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_mount
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_mvs_raw **
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment
- [x] zos_script (doesn't require changes)
  - [x] successful regression pipeline run
- [x] zos_unarchive
  - [x] src code change
  - [x] test case change
  - [x] successful regression pipeline run
  - [x] include under `major_changes` in changelog fragment

There are "choices" defined in `template.py` too, but those don't look like they require any changes.

Other modules without choices defined, but which tests may be impacted:
- [x] zos_encode (depends on zos_data_set, zos_job_submit)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_fetch (depends on zos_data_set, zos_job_submit)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_job_output (depends on zos_job_submit)
  - [x] test case change
  - [x] successful regression pipeline run
- [x] zos_job_query (depends on zos_data_set, zos_job_submit)
  - [x] test case change
  - [x] successful regression pipeline run

rexemin:
##### SUMMARY
Fixes #1360

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
ibm_zos_core

##### ADDITIONAL INFORMATION

```paste below

```


plugins/module_utils/data_set.py
```diff
@@ -919,7 +919,7 @@ def _build_zoau_args(**kwargs):
                 secondary += space_type
 
         type = kwargs.get("type")
-        if type and type == "ZFS":
+        if type and type.upper() == "ZFS":
             type = "LDS"
```
richp405, line 923: shouldn't this be 'type="lds""?

plugins/modules/zos_copy.py
```diff
@@ -2802,24 +2802,15 @@ def run_module(module, arg_def):
 
             # dest_data_set.type overrides `dest_ds_type` given precedence rules
             if dest_data_set and dest_data_set.get("type"):
-                dest_ds_type = dest_data_set.get("type")
+                dest_ds_type = dest_data_set.get("type").upper()
```
richp405, line 2805: This is explicitly calling upper, when the point of this ticket is everything should be lower.  Am I missing something here?

richp405 (changes requested):
This is an amazing piece of work!  Very detailed.  I have found 2 issues which may need changing, or I may misunderstand what I saw.  Let me know either way.

plugins/modules/zos_copy.py
fernandofloresg: That is because data_set.DataSet.MVS_PARTITIONED values are kept in uppercase, as I said in the other comment not every mention to options changed to lowercase. May I ask @rexemin Why were data_set.DataSet frozen sets not updated to lowercase ?

fernandofloresg (approved):
I don't have any question that is not currently asked by Rich.

plugins/module_utils/data_set.py
rexemin: This particular call to ZOAU requires args to be uppercase, hence the call to .upper(). I originally changed it to lowercase and only found out in testing that it broke this code path.

plugins/modules/zos_copy.py
rexemin: I tried to modify as much as possible that didn't require changing every call inside the collection. There are certain utils and system calls that need args to be in uppercase, and to try and hunt them down would block the rest of the PRs for a lot longer, hence why I focused mostly on changing interfaces and just making sure internals still worked correctly.

ketankelkar (approved):
LGTM- amazing work here Ivan!! Thanks for delivering this mighty effort (which was merely an idea just last week)!! 

I've also validated the functionality delivered here while updating (and running) the playbooks in the samples repo. 

I think there are some places where stylistic changes could be made, eg:
-  instead of removing a call to `upper()`, replace it with a call to `lower()`
-  leaving params in lower-case until upper is required (eg for certain zoau calls)

But, we should make those as part of other development efforts rather than here. The interface has been updated and I saw no gaps in functionality, so I'd consider this issue completed.

PR #1452, opened 2024-04-16
Enhancement/1170/make pipeline 217 compatible

richp405:
##### SUMMARY
Changes to 1.10 ignore and sanity test include for Ansible 2.17 compatibility

Fixes 1170

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
tests/sanity needed ignore-2.17 file
Also tested on test_zos_data_set_func.py, which exemplified the need for a change to import.


richp405:
image

tests/functional/modules/test_zos_data_set_func.py
```diff
@@ -18,7 +18,8 @@
 import pytest
 import time
 import subprocess
-from pipes import quote
+# from pipes import quote
```
fernandofloresg, line 21: We can just remove this

fernandofloresg:
Got a question, why is this tested only in one test case? I searched for quote() function and is used in multiple tests. I think we should at least test those right?

fernandofloresg (approved):
Looks good to me

PR #1467, opened 2024-04-22
[Enabler][module_utils/data_set] Add GDG and GDS name resolve functions.

### LINKED ISSUE #1264: [Enabler] [module_utils/data_set.py] Revert extra conditional in DataSet.ensure_present()

### Enabler description

ZOAU Tracker - `JIRA-10451`

Expected outcome:
Error messages surfaced from the underlying system should be properly propagated through ZOAU python Exceptions and the conditional to determine when to attempt a re-catalog, which was widened can be narrowed back to its original state to save on unnecessary function calls.

revert from
```python3
if "DatasetVerificationError" in e.msg or "Error Code: 0x4704" in e.msg:
```
to:
```python3
 if "Error Code: 0x4704" in e.msg:
```
[link to code](https://github.com/ansible-collections/ibm_zos_core/blob/73ba2c1174f4e22a4796f9c23072f62ea2ace497/plugins/module_utils/data_set.py#L187)

Context:
As of the creation of this issue (Feb 28, 2024), the `DatasetVerificationError` thrown by the ZOAU python API does not surface any error information from any underlying tools/calls/systems. This behavior primarily stemmed from the migration from the `datasets._create()` function call to the `datasets.create()` function call, where the former throws the `DatasetVerification` Exception while the latter returns a `ZOAUResponse` object, which contained stdout/stderr info. 

The above scenario (and code path) is triggered when the `ensure_present()` function calls `DataSet.create()` for an _existing_ and _un-cataloged_ data set on a particular volume. In this case, the "create" operation is _expected- to fail (and it does) and the appropriate action would be to re-catalog the data set in question. 

`datasets._create()` returns a `ZOAUResponse` Object whose contents are:
```4  BGYSC1402E Dynalloc Failure. Error Code: 0x4704, Information Code: 0x5c```

`datasets.create()` raises a `DatasetVerificationError` whose contents are:
```KETAN.PRIVATE.TESTDS```

The `zos_data_set` module code was recently refactored to call the `create()` function rather than the “hidden” `_create()`, which is how this discrepancy was exposed and why the conditional highlighted needed to be widened.

Recreate (see attached file for `python` script):
- create a data set
- un-catalog data set
- ** attempt to re-create same data set
- with _create()
- with create()
- cleanup
- re-catalog data set
- remove data set

[recreate-no-dynalloc-error-code.py.txt](https://github.com/ansible-collections/ibm_zos_core/files/14441103/recreate-no-dynalloc-error-code.py.txt)

### LINKED ISSUE #1382: [Enabler] [module_utils/data_set] Investigate zoau special char support and lmplement escape name function

- \#
- @
- $
- \-

Goal is to create a function or group of functions to given a data set name, will return the escaped name.

Before investing into creating this function, test some of the most used zoau python APIs to validate that we need to use escaped characters.

Then, need to review this pending PR #1020 , the way it handles escaping special characters and see if we can leverage the whole function or a modified version of it.

Analyze if it makes sense to add this functions into BetterArgParser or module_utils/data_set.py

----- This is context and not part of the task
My idea is to use a class data set so in modules we can take a more object-oriented way to handle data sets, so instead of having a var src, ds_type and so, we could just instantiate a new data set object and just call dataset.ds_type or dataset.name, that way we can pass it into functions and use the escaped name like `dataset.escaped_name`, and if something fails we just use `dataset.name` to return the input data set name.
Anyways, we will need to change the signature of some functions to deal with two data set names, so I will rather use a dataset object instead of adding escaped_name as a new parameter.

Right now, already exists a `DataSet` module in module_utils/data_set.py, which by the looks of it, looks more of a DataSetHandler bc it only performs operations on data set, I don't think is a good idea to rename it since it will involve changes in 16 modules, rather we can create a new data set class named MVSDataSet or zOSDataSet.
----- This is context and not part of the task

Analyze if we can add unit tests for this.

### Ansible module

BetterArgParser or module_utils/data_set.py

### LINKED ISSUE #1386: [Enabler] [module_utils/data_set.py] GDG/GDS support

### Enabler description

GDG Support in data set module utils to identify a gdg data set and resolve the relative name into an absolute name.

```python
DataSet.is_gdg() # resolve if a name is gdg
```
We can know if a provided data set name is a GDS relative name just by looking into the parenthesis content, given `ZOAUSER.GDG(0)` or `ZOAUSER.GDG(-2)` just review `(0)` or `(-2)`.

When GDS relative name is positive like `(+1)` we don't need to resolve the name bc when a new data set is created we can use the relative name to pass either to `datasets.copy` or `datasets.create` but that is reserved for `zos_data_set` or `zos_copy` modules. 

GDG relative name -> absolute name should be done in each module via module utils like
```python
DataSet.resolve_gds_abs_name() # get absolute name from relative gds name
```

### Ansible module

_No response_

fernandofloresg:
##### SUMMARY

Adds two new functions to DataSet class, `is_gds_relative_name` and `resolve_gds_absolute_name`, the first, given a data set name identifies if it is a GDS relative name. The second one, takes a GDS relative name and resolves its absolute GDG name.

Fixes #1386 
Fixes #1264 
Fixes #1382 
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

module_utils/data_set
##### ADDITIONAL INFORMATION

```paste below

```


richp405 (approved):

plugins/module_utils/data_set.py
```diff
@@ -1717,6 +1797,69 @@ def _process_listcat_output(self, output):
         return result
 
 
+class MVSDataSet():
```
ketankelkar, line 1800: i'm curious as to the purpose of a new `MVSDataSet` class. Primarily, I'm unclear on why the existing DataSet class cant be enhanced. 

Also, thinking into future maintainability, will it create confusion if there's a DataSet object going around as well as an MVSDataSet object? Will it be clear which class/type is best suited for any given scenario?

tests/unit/test_data_set_utils.py
```diff
@@ -0,0 +1,53 @@
...
+special_chars_test_data = [
+    {"name": "USER.SPECIAL.@TEST", "escaped_name" : r"USER.SPECIAL.\@TEST"},
+    {"name": "USER.SPECIAL.A$A", "escaped_name" : r"USER.SPECIAL.A\$A"},
+    {"name": "USER.SPECIAL.$TEST#@", "escaped_name" : r"USER.SPECIAL.\$TEST\#\@"},
+    {"name": "USER.SPECIAL.TEST", "escaped_name" : r"USER.SPECIAL.TEST"},
+    ]
+
+@pytest.mark.parametrize("data_set", special_chars_test_data)
+def test_data_set_name_escaping(data_set):
+    print(data_set)
```
AndreMarcel99, line 51: Remove the print

plugins/module_utils/data_set.py
fernandofloresg: Sorry I left that part out, I was in a rush and didn't explain a lot of things in this PR.

So, DataSet class description is `Perform various data set operations such as creation, deletion and cataloging.`, in a sense is more like a DatasetUtils module, we do have plenty of static methods effectively using this class as a library.

New proposed MVSDataSet class is meant to be a representation of a single data set during the execution of a module, took this one from my current work in `zos_data_set` module, by having a single dataset attributes encapsulated in a class we can pass around the object instead of having functions with many arguments in their signature.

Effectively changing it from this :

Screenshot 2024-05-02 at 9 33 57 AM


To this (still in progress) :

Screenshot 2024-05-02 at 9 34 50 AM

Apart from reducing the number of arguments encapsulation benefits us when having to do data set operations like creating a new dataset or copying it, by having its own attributes in the object and not in some other variables scattered around we can simplify complex flows like in `zos_copy`.

Another benefit is that this object at the moment of initialization has attributes `raw_name` and `name`, `name` is escaped in case of needing it and if the name is a gds relative name (non positive one) is resolved into an absolute name, meaning, that `name` contains the name we use internally for the data set operations, whilst `raw_name` is the original name, so that when we need to show any error message to our users we display `data_set.raw_name`.

MVSDataSet was used because DataSet was already taken and updating it would have to be done in 120+ places. I'm open for suggestions just added this class into this PR because Ivan, Rich and Marcel have work items that depend on this, another option name was zDataSet...

ketankelkar: that makes sense to me! thanks for the explanation

ketankelkar (approved):
LGTM!!

Reminder to update the copyright years for `better_arg_parser.py` and `test_arg_parser`

PR #1485, opened 2024-04-26
[Documentation] Inform users the need to escape commands with special characters

### LINKED ISSUE #1471: [Documentation] [zos_operator] Include instructions to the user to escape # and $ characters for JES2 commands.

Related to issue #1444 

Currently users that need to execute a JES2 command using `zos_operator` have to escape the `$` symbol, we do provide an example in the docs but need to be more specific by instructing users when to escape those symbols.

Documentation in the short term, where the docs might instruct users to escape command prefixes, eg $, # and the doc include some updated samples. One might be to show and example of how to figure out the command prefix the system is using and possibly mention this in documentation as well.
For example, this command D OPDATA,PREFIX (short version I believe is D O) yields:

Current example in the docs is :

```yaml
- name: Execute an operator command to purge all job logs (requires escaping)
  zos_operator:
    cmd: "\\$PJ(*)"

```

v1.9.0

### Ansible module

zos_operator

fernandofloresg:
##### SUMMARY

Updated doc entry with escaping instructions when using special characters.

Fixes #1471

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

zos_operator

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/source/modules/zos_operator.rst
+++ b/docs/source/modules/zos_operator.rst
@@ -33,6 +33,10 @@ cmd
 
   For example, change the command "...,P='DSN3EPX,-DBC1,S'" to "...,P=''DSN3EPX,-DBC1,S'' ".
 
+  If the command contains any special characters ($, \\#, etc), they must be escaped using double backslashes like \\\\$.
+
+  For example, to display info on a volume \`\`$abcd\`\` , the command would have \`\`cmd="d u,,,\\\\$abcd"\`\`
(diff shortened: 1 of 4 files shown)


ketankelkar (approved):
Remember to update copyright years, otherwise LGTM

I like that you provided an example of special characters and an explicit example in the docs too

PR #1494, opened 2024-05-02
[Documentation] [zos_operator] Add notes to zos_operator

### LINKED ISSUE #1471: [Documentation] [zos_operator] Include instructions to the user to escape # and $ characters for JES2 commands.

fernandofloresg:
##### SUMMARY

Add a notes section in zos_operator that instructs users how to find the command prefixes needed for certain commands.

Fixes #1471 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

zos_operator

This is how the docs look like rendered:

Screenshot 2024-05-02 at 11 32 49 AM


ketankelkar (approved):
LGTM - 

I think it's fine the way you have it, here's just another phrasing to consider:
> Certain commands may require specific prefixes (e.g. '$'). The prefix can be discovered by issuing the command C(D OPDATA,PREFIX)

PR #1511, opened 2024-05-17
[Enabler][zos_archive][zos_unarchive] Add support for GDGs and validate symbols support

### LINKED ISSUE #1367: [Enabler] [zos_archive] GDG/GDS and special character support
# Enabler description

```python
MVSDataSet(
            name = params.get("name"),
        )
```

GDG/GDS Support
-------------------

`src` option in `zos_archive` accepts a list of string either as a pattern or a list of data set names, so user can provide a list of data set names that contain some GDS relative data set names, in that case, the module needs to resolve the relative name into absolute name before packing to correctly archive the given data set name.

```yaml
- name: Compress data set into a terse, specify pack algorithm and use adrdssu
  zos_archive:
    src: 
      -  "USER.GDG(0)"
      -  "USER.NEWGDG(-2)"
    dest: "USER.ARCHIVE.RESULT.TRS"
    format:
      name: terse
      format_options:
        terse_pack: "SPACK"
        use_adrdssu: True
```

```python
```

Symbols support
-----------------
As part of the special character support we need to leverage the escape function from the betterArgParser created on #1382 to escape each data set name. This module uses different system utilities like TERSE, TSOCMD, ADRDSSU, **as part of this task we still need to validate if escaping is needed or not in this module**.

```python
```

- [ ] Do the development work

### Ansible module

zos_archive

fernandofloresg:
##### SUMMARY

Add support for GDGs and add tests for special character support.

Fixes #1367 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_archive

##### ADDITIONAL INFORMATION

```paste below

```


fernandofloresg:
zos_archive 
Screenshot 2024-06-11 at 11 22 07 AM

zos_unarchive
Screenshot 2024-06-11 at 11 22 26 AM

changelogs/fragments/1367-zos_archive-gdg-support.yml
richp405, line 1: The comment for the PR says this is #1511, for zos_archive, but the fragment doesn't say that.

plugins/modules/zos_unarchive.py
```diff
@@ -35,7 +35,7 @@
       - The remote absolute path or data set of the archive to be uncompressed.
       - I(src) can be a USS file or MVS data set name.
       - USS file paths should be absolute paths.
-      - MVS data sets supported types are C(SEQ), C(PDS), C(PDSE).
```
richp405, line 38: Should line 38 be removed?  I thought you added gdg, but didn't remove SEQ/PDS/et cetera.

richp405 (changes requested):
my concern is the changelog fragment: 1 - the link should point to the pull request, 2 - the file needs to mention archive and unarchive.  I see the data set change was probably needed for this to work, so it should be mentioned in the fragment, but after the main changes.

changelogs/fragments/1367-zos_archive-gdg-support.yml
fernandofloresg: sorry forgot to update this.

AndreMarcel99 (approved):

plugins/modules/zos_unarchive.py
rexemin: I'm wondering the same

PR #1516, opened 2024-05-21
[Enabler][1376]lineinfile_blockinfile_gdg/gds_and_special_character_support

### LINKED ISSUE #1376: [Enabler] [zos_lineinfile] GDG/GDS and special character support
# Enabler description

New MVSDataset class
-----------------------

```python
MVSDataSet(
            name = params.get("name"),
        )
```

GDG/GDS Support
-------------------

Option `src` and `backup_name` accept a data set name or path, enabler is to support GDS relative names.

```yaml
- name: Ensure value of a variable in the sequential data set
  zos_lineinfile:
    src: ZOAUSER.GDG(0)
    regexp: '^VAR='
    line: VAR="some value"

- name: Ensure value of a variable in the sequential data set
  zos_lineinfile:
    src: USER.PRIVATE.TEST
    regexp: '^VAR='
    line: VAR="some value"
    backup: True
    backup_name: ZOAUSER.GDG(+1)

```

Internally, in the case of a data set name being a GDS relative name we should resolve those names into absolute GDS names.

```python
```

Symbols support
-----------------
 
data set name with symbol

`ZOAUSER.TE$T`

```python
```

- \#
- @
- $
- \-

- [ ] Do the development work
### Ansible module

zos_lineinfile

AndreMarcel99:
##### SUMMARY
Add gdgs and special special character support for blockinfile and lineinfile and backup utils.
Fixes #1376 and #1369

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
Add execution of deed

##### ADDITIONAL INFORMATION
Add execution of special cases in block and lineinfile modules.

Captura de pantalla 2024-05-22 a la(s) 3 40 33 p m


richp405 (approved):

plugins/modules/zos_blockinfile.py
```diff
@@ -666,19 +666,27 @@ def main():
...
+    gdg = False
     # analysis the file type
+    if "/" not in src:
+        dataset = data_set.MVSDataSet(
+            name=src
+        )
+        src = dataset.name
+        gdg = dataset.name
+
+    if ("(" in src and ")" in src) and ("+" in src or "-" in src) and gdg is False:
```
rexemin, line 678: Couldn't you use `is_gds_positive_relative_name` or the other `is_gds_positive_relative_name` here to see if the user supplied a data set that doesn't exist yet?

fernandofloresg (changes requested):

plugins/module_utils/backup.py
```diff
@@ -349,3 +356,22 @@ def __init__(self, message, rc=None, stdout=None, stderr=None):
         self.stdout = stdout
         self.stderr = stderr
         super(BackupError, self).__init__(self.msg)
+
+
+def is_gds_relative_name(name):
```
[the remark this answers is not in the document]
AndreMarcel99: Already move

plugins/modules/zos_lineinfile.py
```diff
@@ -377,6 +377,80 @@ def absent(src, line, regexp, encoding, force):
...
+    else:
+        if regex:
+            if line:
+                options += f'-s -e "/{regex}/d" -e "/{line}/d" "{src}" '
+            else:
+                options += f'"/{line}/d" "{src}" '
+        else:
+            options += f'"/{line}/d" "{src}" '
+
+    cmd = "dsedhelper {0}{1}{2}{3}".format(force, backrefs, encoding, options)
```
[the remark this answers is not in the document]
AndreMarcel99: In the ZOAU repository they still use `cccccbiegnfudcrnrhkcebkdrtfjbdfijvedrrilgngl
` as it option for `dataset.lineinfile()`

AndreMarcel99:
Captura de pantalla 2024-05-27 a la(s) 10 53 48 a m
Captura de pantalla 2024-05-27 a la(s) 10 54 01 a m

plugins/modules/zos_blockinfile.py
```diff
@@ -666,19 +666,28 @@ def main():
...
+    gdg = False
     # analysis the file type
+    if "/" not in src:
+        dataset = data_set.MVSDataSet(
+            name=src
+        )
+        src = dataset.name
+        gdg = dataset.name
+
+    if data_set.DataSet.is_gds_relative_name(src) and gdg is False:
```
fernandofloresg, line 678: I think I mentioned this in the previous review, but `gdg` variable name is used to store two different data types, and gdg variable is used for anything else than the negative case. What if we change this line to 

```suggestion
    elif data_set.DataSet.is_gds_relative_name(src):
```
and we just get rid of `gdg` variable here

plugins/modules/zos_blockinfile.py
```diff
@@ -666,19 +666,28 @@ def main():
...
+        module.fail_json(msg="{0} does not exist".format(src))
```
fernandofloresg, line 679: Is this error message correct? Seems to me that if src contains "/" and is a GDS relative name then would fail. Let's say we have for some reason something like "OMVSADM.A/A.(+1)" which seems to be the case here. Isn't this catch by the betterargparser? Also, if this is failing when a GDS relative name is found wouldn't that get resolved by line 672? Could you share a failing example for this case I just can't wrap my head around a failing scenario.

plugins/modules/zos_lineinfile.py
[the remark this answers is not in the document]
fernandofloresg: I repeat, those are for ZOAU internal use and will be removed in future, they can use it as they please but we should not be dependent on those.

Unless, there is a case we cannot cover with just using `dsed`, in such scenario we should open a ticket with them to request compatibility.

plugins/modules/zos_lineinfile.py
```diff
@@ -500,32 +573,44 @@ def main():
...
     # analysis the file type
+    if "/" not in src:
+        dataset = data_set.MVSDataSet(
+            name=src
+        )
+        src = dataset.name
+        gds = dataset.is_gds_active
+
+    if data_set.DataSet.is_gds_relative_name(src) and gds is False:
+        module.fail_json(msg="{0} does not exist (1)".format(src))
```
fernandofloresg, line 590: Here I just have the same question about the error scenario, `gds` variable does look like is used in other places.

fernandofloresg (changes requested):

plugins/modules/zos_blockinfile.py
fernandofloresg, line 1: We need to add into the docs the fact that we support GDG and GDS now and provide an example for it.

plugins/modules/zos_lineinfile.py
fernandofloresg, line 1: We need to add into the docs the fact that we support GDG and GDS now and provide an example for it.

plugins/modules/zos_blockinfile.py
AndreMarcel99: This case with the catch of relative name suggest that is a (+1) GDG and blockinfile and lineinfile required datasets already created, so if still as relative name throw the error

plugins/modules/zos_lineinfile.py
AndreMarcel99: Is a validation variable to call or no to call deed execution not lineinfile python API

AndreMarcel99:
Captura de pantalla 2024-05-28 a la(s) 3 04 08 p m
Captura de pantalla 2024-05-28 a la(s) 3 04 18 p m
Captura de pantalla 2024-05-28 a la(s) 3 04 34 p m

AndreMarcel99:
Captura de pantalla 2024-05-28 a la(s) 5 12 50 p m
Captura de pantalla 2024-05-28 a la(s) 5 12 59 p m
Captura de pantalla 2024-05-29 a la(s) 10 08 07 a m
Captura de pantalla 2024-05-29 a la(s) 10 08 17 a m

plugins/modules/zos_blockinfile.py
```diff
@@ -279,6 +279,20 @@
...
+
+- name: Add a block to a gds
+  zos_blockinfile:
+    src: SYS1.SOME.PARTITIONED(0)
+    insertafter: EOF
+    block: "{{ CONTENT }}"
+
+- name: Add a block to dataset and backup in a new generation of gds
+  zos_blockinfile:
+    src: SOME.PARTITIONED.TEST
```
fernandofloresg, line 291: small comment, looks like this is called partitioned but is used as a sequential, better to chance the name to not confuse users.

plugins/modules/zos_blockinfile.py
fernandofloresg: cool

fernandofloresg (changes requested):
Nice examples, just need to add the GDS support in options documentation like `backup_name` and `src`

AndreMarcel99:
Captura de pantalla 2024-05-30 a la(s) 4 15 31 p m
Captura de pantalla 2024-05-30 a la(s) 4 15 41 p m

plugins/modules/zos_blockinfile.py
```diff
@@ -38,6 +38,8 @@
       - The location can be a UNIX System Services (USS) file,
         PS (sequential data set), member of a PDS or PDSE, PDS, PDSE.
       - The USS file must be an absolute pathname.
+      - Generation data set (GDS) relative name of generation already
+        created SOME.CREATION(-1).
```
fernandofloresg, line 42: When adding the example is better to change the style of it and indicate is an example.
```suggestion
        created.  C(e.g. SOME.CREATION(-1).)
```

fernandofloresg (changes requested):
Just one last change I swear :)

fernandofloresg (approved):

PR #1518, opened 2024-05-23
[zos_find] Support GDG and special characters in data set names for zos_find

### LINKED ISSUE #1374: [Enabler] [zos_find] GDG/GDS and special character support
### Enabler description

```python
MVSDataSet(
            name = params.get("name"),
        )
```

GDG/GDS Support
-------------------

Option `patterns` and `pds_patterns` accept a list of string either as a pattern or a list of data set names, so user can provide a list of data set names that contain some GDS relative data set names, in that case, the module needs to resolve the relative name into absolute name before packing to correctly archive the given data set name.

```yaml
- name: Exclude data sets that have a low level qualifier 'TEST'
  zos_find:
    patterns: ZOAUSER.GDG(-1)
    contains: "hello"

```

```python
```

Symbols support
-----------------
 
Add symbol support to the dest_data_set name option.

data set name with symbol

`ZOAUSER.TE$T`

```python
```

- \#
- @
- $
- \-

- [ ] Do the development work

### Ansible module

zos_find

### LINKED ISSUE #1524: [Enhancement] allow zos_find to search and filter GDGs by attributes

We want to expand zos_find to allow GDG-related attributes to be used as filter of a find.

Initial discussion with Fernando during #1374 review (#1518)

[fernandofloresg] /1518#discussion_r1612307235
Hey Rich, I don't know if this was added into the development task but I think we also should be able to filter the GDGs by its attributes, I mean, return GDGs with purge = True or return where extended = True. Is that something that can be added in this same task ?

### Ansible module

zos_find

richp405:
##### SUMMARY
Add support and test for GDG data set convention, and validate special-character use in ds names.

Fixes #1374
Fixes #1524 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_find


richp405:
image

richp405:
I had to re-push, because dev changed while I was testing.

fernandofloresg (changes requested):

tests/functional/modules/test_zos_find_func.py
fernandofloresg, line 1: Hey Rich, I don't know if this was added into the development task but I think we also should be able to filter the GDGs by its attributes, I mean, return GDGs with purge = True or return where extended = True. Is that something that can be added in this same task ?

richp405: I'll check on Friday for this... not sure, offhand, but we can touch up the interface.

tests/functional/modules/test_zos_find_func.py
```diff
@@ -33,6 +33,18 @@
     "TEST.FIND.VSAM.FUNCTEST.FIRST"
 ]
 
+GDG_NAMES = [
```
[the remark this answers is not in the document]
richp405: I'll read through that... the reasons it was hardcoded before is that the testing was checking for the files with asterisks (like test.find.pds.*).. I think I can work up a compromise (like test.find.pds.randomname?)

tests/functional/modules/test_zos_find_func.py
richp405: It's a good thought, but I think we'll do it in a separate ticket, because changing the interface needs to be separate from adding options.  We can still fit this into 1.11-b1

tests/functional/modules/test_zos_find_func.py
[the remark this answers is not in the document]
richp405: I'm working through the change to run the tests against variable names.  good idea.  It will be ready Tuesday.

tests/functional/modules/test_zos_find_func.py
fernandofloresg: Sure, I noticed is not mentioned in the base issue, my bad.

fernandofloresg (changes requested):
I code wise test is good. The only thing I would ask to add that is in the issue description is to modify the zos_apf docs to include the fact that GDGs are now supported.

tests/functional/modules/test_zos_find_func.py
richp405: Added #1524 as new enhancement to split up the work.

tests/functional/modules/test_zos_find_func.py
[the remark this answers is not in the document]
richp405: Make that Friday of this week.  I'll re-issue invites then.

tests/functional/modules/test_zos_find_func.py
fernandofloresg: Great thanks Rich!

tests/functional/modules/test_zos_find_func.py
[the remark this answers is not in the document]
fernandofloresg: Sure, I don't expect you to change all tests just this one to not introduce new hardcoded content.

fernandofloresg (approved):
I did some changes myself, hence moving from changes requested to approved. Will merge once we have 3 approvals

fernandofloresg:
Screenshot 2024-07-14 at 10 32 15 PM

rexemin (approved):

PR #1519, opened 2024-05-23
[Enabler] [zos_fetch] Add gdg support to zos_fetch

### LINKED ISSUE #1373: [Enabler] [zos_fetch] GDG/GDS and special character support

### Enabler description

New MVSDataset class
-----------------------

```python
MVSDataSet(
            name = params.get("name"),
        )
```

GDG/GDS Support
-------------------

Option `src` accepts a data set name or path, enabler is to support GDS relative names.

```yaml
- name: Fetch a latest GDS in a GDG
  zos_fetch:
    src: ZOAUSER.GDG(0)
    dest: /tmp/
    flat: True

- name: Fetch a previous generation GDS in a GDG
  zos_fetch:
    src: ZOAUSER.GDG(-1)
    dest: /tmp/
    flat: True

```

```python
```

Symbols support
-----------------
 
data set name with symbol

`ZOAUSER.TE$T`

```python
```

- \#
- @
- $
- \-

- [x] Do the development work
- [x] Update the docs to talk about GDG and GDS handling in the module.
- [x] Update the docs with at least 1 example on how to use with GDSs.
- [x] Create a new test case for GDSs.
- [x] Create a new test case for symbols.

### Ansible module

zos_fetch

rexemin:
##### SUMMARY
Fixes #1373.

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
- zos_fetch
- module_util/data_set.py


rexemin:
Screenshot 2024-05-27 at 11 51 59
Screenshot 2024-05-27 at 11 52 14

tests/functional/modules/test_zos_fetch_func.py
```diff
@@ -664,3 +666,133 @@ def test_fetch_flat_create_dirs(ansible_zos_module, z_python_interpreter):
     finally:
         if os.path.exists(dest_path):
             shutil.rmtree("/tmp/" + remote_host)
+
+
+def test_fetch_sequential_data_set_with_special_chars(ansible_zos_module):
+    hosts = ansible_zos_module
+    TEST_PS = "USER.TEST.@N$IBLE#"
```
AndreMarcel99, line 673: I recommend for the name with symbols we have the option of get_tmp_ds_name(symbols=True)

AndreMarcel99 (changes requested):

tests/functional/modules/test_zos_fetch_func.py
rexemin: Done

fernandofloresg (approved):
Great work Ivan

PR #1527, opened 2024-05-30
[Enabler] [zos_backup_restore] Add support for GDG and special characters

### LINKED ISSUE #1368: [Enabler] [zos_backup_restore] GDG/GDS and special character support

### Enabler description

New MVSDataset class
-----------------------

```python
MVSDataSet(
            name = params.get("name"),
        )
```

GDG/GDS Support
-------------------

Options `include` and `exclude` under `data_sets` accept a list of data set names or data set patterns, in the case of a data set name being a GDS relative name we should resolve those names into absolute GDS names.

Note that maybe GDS support for `exclude` option when `operation=restore` might not make sense, this needs to be investigated.

Explore if we can just past a GDG base name like `USER.GDG` and backup the whole GDG and then restore back their data sets.

```python
```

Symbols support
-----------------
 
```python
```
- \#
- @
- $
- \-

- [ ] Do the development work

### Ansible module

_No response_

fernandofloresg:
##### SUMMARY

Added support to use GDS relative names in `include` and `exclude` options when operation = `backup`.

Fixes #1368 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_backup_restore

##### ADDITIONAL INFORMATION

```paste below

```


fernandofloresg:
Screenshot 2024-05-31 at 2 42 40 PM
Screenshot 2024-05-31 at 2 42 40 PM

AndreMarcel99 (approved):
I like the solution, all clean

fernandofloresg:
> Only change I would request is to put in a changelog fragment. Code and comments look fine.

Very valid request. Is a must in my opinion I overlooked it...

PR #1549, opened 2024-06-24
[v1.9.2][Bugfix] [zos_copy] Only issue opercmd to check for a locked data set if dest exists

### LINKED ISSUE #1544: [Bug] [zos_copy] Copy uses opercmd to check if the data set is in use even if the data set is created by the module

### Bug description

zos_copy is using opercmd and needs higher privileges, but when creating the destination data set itself is not needed.

Check is in the following line:

https://github.com/ansible-collections/ibm_zos_core/blob/9c09b5899bfcb06055d7c186f95f7c5e4ca0f625/plugins/modules/zos_copy.py#L2860

```python
    if dest_ds_type != "USS":
        if not force_lock:
            is_dest_lock = data_set_locked(dest_name)
            if is_dest_lock:
                module.fail_json(
                    msg="Unable to write to dest '{0}' because a task is accessing the data set.".format(dest_name))

```

We should only check if the destination data set is in use if not created by the module itself.

v1.9.0 (default)

v1.2.5 (default)

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

zos_copy

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

 `zos_copy` was issuing an opercmd to see if the destination data set is locked even if it does not exist. Fix only issues the command if destination exists.

This is the fix meant for 1.9.2 release, did not add a test case for this scenario since automating a test to validate the command was not issued would take me some time, basically that would need to check the syslog to make sure the command was not issued for that specific data set name, I can add it if requested.

Fixes #1544 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_copy

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1549-zos_copy-opercmd.yml
+++ b/changelogs/fragments/1549-zos_copy-opercmd.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - zos_copy - module would use opercmd to check if a non existent
+    destination data set is locked. Fix now only checks if the destination
+    is already present.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1549).
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -2856,7 +2856,7 @@ def run_module(module, arg_def):
     # the machine and not generate a false positive check the disposition
     # for try to write in dest and if both src and dest are in lock.
     # ********************************************************************
-    if dest_ds_type != "USS":
+    if dest_exists and dest_ds_type != "USS":
         if not force_lock:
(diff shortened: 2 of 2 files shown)


fernandofloresg:
Screenshot 2024-06-28 at 12 46 12 PM

richp405 (approved):

rexemin (approved):

ddimatos (commented):
Looks good, nice and easy :)

PR #1632, opened 2024-08-01
[Enhancement][117]Validate_to_restore_keep_orginial_hlq

### LINKED ISSUE #1177: [Enhancement] [zos_backup_restore] Allow for the `restore` to not force a HLQ or default

### Ansible module

zos_backup_restore

The module `zos_backup_restore` when backing up data sets which have a HLQ do not allow to transfer when restoring. Today if you backup and then proceed to restore, the restore operation requires that you provide a HLQ or it will use the TSO ID as the default. The issue is that a back up can consist of different HLQ's such as `SYS1.PROCLIB.PDS` , `SYS2.PROCLIB.PDS` and so on, thus during a restore all those data sets will be given the same HLQ which has 2 issues, it could result in error as s result of duplicate qualifiers and second, it forces the data sets to all have the same qualifier. 

The code should be fairly straight forward and a fix in this block:
```  hlq = None
    if dependencies.get("operation") == "restore":
        hlq = datasets.hlq()
    return hlq
```

AndreMarcel99:
##### SUMMARY
Add test case and validate the new behavior of keep_original_hlq.

Fixes #1177 

##### ISSUE TYPE
- Enhancement Pull Request

##### COMPONENT NAME
Delete hlq_default variable that check for the hlq and add test case to validate it.

Captura de pantalla 2024-08-01 a la(s) 11 42 30 a m
Captura de pantalla 2024-08-01 a la(s) 11 42 18 a m


richp405 (approved):

richp405 (approved):

changelogs/fragments/1632-Validate_to_restore_keep_orginial_hlq.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
```
rexemin, line 1: Since you're changing a little bit what one of the options can do, I would use `minor_changes` instead of `trivial`

plugins/modules/zos_backup_restore.py
```diff
@@ -187,6 +187,7 @@
     description:
       - Specifies the new HLQ to use for the data sets being restored.
       - Defaults to running user's username.
+      - Use keyword N to restores datasets to their original HLQ from when they were archived.
```
rexemin, line 190: Discussed this change with Marcel and we agreed that using the same option already available to users but reserving a special keyword (or key letter) would be more straightforward for everybody instead of adding a new option to the module's interface. Probably better to use 'K' (as in KEEP) or 'R' (as in RESTORE) though. Should not conflict with valid HLQs since it is just one character.

rexemin (changes requested):

AndreMarcel99:
Captura de pantalla 2024-08-02 a la(s) 12 31 17 p m
Captura de pantalla 2024-08-02 a la(s) 12 31 28 p m

richp405 (approved):

plugins/modules/zos_backup_restore.py
```diff
@@ -187,6 +187,7 @@
     description:
       - Specifies the new HLQ to use for the data sets being restored.
       - Defaults to running user's username.
+      - Use keywords R or K to restores datasets to their original HLQ from when they were archived.
```
ketankelkar, line 190: `restores` should be `restore`

- Use keywords R or K to restore datasets to their original...

plugins/modules/zos_backup_restore.py
```diff
@@ -322,6 +323,12 @@
     backup_name: /tmp/temp_backup.dzp
     sms_storage_class: DB2SMS10
     sms_management_class: DB2SMS10
+
+- name: Restore data sets from backup stored with their original HLQ from when they were archived .
```
ketankelkar, line 327: `stored` seems unnecessary, also add an `a` in reference to the `backup`, switch `their` to `the`, make `HLQ` plural since datasets is plural, and remove the extra space before the period
- name: Restore data sets from a backup with the original HLQs from when they were archived.

plugins/modules/zos_backup_restore.py
```diff
@@ -1005,12 +991,16 @@ def to_dunzip_args(**kwargs):
...
 
-    if kwargs.get("hlq"):
-        zoau_args["high_level_qualifier"] = kwargs.get("hlq")
-
     if kwargs.get("tmp_hlq"):
         zoau_args["tmphlq"] = str(kwargs.get("tmp_hlq"))
 
+    if kwargs.get("hlq") is None:
+        zoau_args["high_level_qualifier"] = datasets.get_hlq()
+    elif kwargs.get("hlq") == "K" or kwargs.get("hlq") == "R":
```
ketankelkar, line 999: im a bit unclear here, what is the point of introducing two new keywords if they both do the same thing?

plugins/modules/zos_backup_restore.py
```diff
@@ -1005,12 +991,16 @@ def to_dunzip_args(**kwargs):
...
+        zoau_args["keep_original_hlq"] = True
+    elif kwargs.get("tmp_hlq"):
```
ketankelkar, line 1001: i think this `elif` might be a nop (or worse, it might create unexpected hlqs). From the way im understanding this if-elif-elif:
- (1) if "hlq" is not defined by the user, 
  - default behavior - which is running user's name
- (2) if "hlq" is keyword R or K, 
  - set the flag
- (3) if "tmphlq" is defined, 
  -  default behavior, same as from the original `if`.

So what if someone wants their restored data sets to have hlq `ABCD`, and thus defines `hlq: ABCD`, but say the temp data sets should have hlq `XYZ`, so `tmp_hlq: XYZ`. Running through the conditional above, `hlq` is not none, so (1) gets skipped, `hlq` is not 'R' or 'K' so (2) gets skipped. `tmp_hlq` is defined, so now `zoau_args["high_level_qualifier"]` gets set to the user name?

ketankelkar (changes requested):
Introducing key words into the `hlq` module option was a clever idea, but it won't work. Data sets _can_ have single character HLQs:
```bash
>dtouch -tseq R.KETAN.TEST
>echo $?
0
>dls -l r.ketan.test
R.KETAN.TEST                                 PS  FB      80 339002

>dtouch -tseq K.KETAN.TEST
>echo $?
0
>dls -l k.ketan.test
K.KETAN.TEST                                 PS  FB      80 9SX901
```

Which means that I would not be able to take a backup and restore it with all the the HLQs being `K`

plugins/modules/zos_backup_restore.py
AndreMarcel99: Good catch, I already make some changes for those cases return the default_hlq to default value for the hlq and also the if check first exist and then see if is a r. And now I'm using and r because or validation required the hlq should be in upper case, by using only one case of lower case could work as keyword.

AndreMarcel99:
Captura de pantalla 2024-08-05 a la(s) 12 13 53 p m
Captura de pantalla 2024-08-05 a la(s) 12 14 02 p m

richp405 (approved):

AndreMarcel99:
Captura de pantalla 2024-08-07 a la(s) 10 26 30 a m
Captura de pantalla 2024-08-07 a la(s) 10 26 40 a m

richp405 (approved):

plugins/modules/zos_backup_restore.py
```diff
@@ -732,28 +732,7 @@ def hlq_type(contents, dependencies):
...
-    Returns
-    -------
-    str
-        The HLQ to use.
-    """
-    hlq = None
-    if dependencies.get("operation") == "restore":
-        hlq = datasets.get_hlq()
-    return hlq
+    return contents
```
ketankelkar, line 735: i'm curious why the HLQ no longer needs to be upper-cased

plugins/modules/zos_backup_restore.py
```diff
@@ -186,7 +186,7 @@
   hlq:
     description:
       - Specifies the new HLQ to use for the data sets being restored.
-      - Defaults to running user's username.
+      - Defaults to running original HLQ.
     type: str
     required: false
   tmp_hlq:
```
ketankelkar, line 192: edit the description of the default hlq behavior here too, it still says the default is Ansible user

>      - The default HLQ is the Ansible user that executes the module and if
>        that is not available, then the value of C(TMPHLQ) is used.

changelogs/fragments/1632-Validate_to_restore_keep_orginial_hlq.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - zos_backup_restore - Validate the datasets keep original hlq on restore operation.
```
ketankelkar, line 2: "Validate the datasets" is an extremely vague phrase. The contents of the fragment will go into the changelog and release notes and what changed will be unclear to anyone unfamiliar with this PR/code.

Maybe try something like:
> Redefines the default behavior of module option hlq. The original high level qualifiers in a backup will be used for a restore operation if no value is provided in the module option.

plugins/modules/zos_backup_restore.py
ketankelkar, line 1: The playbook task description in the example on line 273 needs to updated because it still references the old default behavior for when a value is not provided for the module option hlq.

>- name: Restore data sets from backup stored in the UNIX file /tmp/temp_backup.dzp.
    Use z/OS username as new HLQ.

ketankelkar (changes requested):
overall looks pretty good, just a few finishing touches required since the default behavior is changing.

plugins/modules/zos_backup_restore.py
AndreMarcel99: We already check [lines before](https://github.com/ansible-collections/ibm_zos_core/blob/10ac10cd7d08ec6cbcde8380ff501f1274ed6b52/plugins/modules/zos_backup_restore.py#L733)  if the HLQ is in upper case if not we already tell the user that is invalid

plugins/modules/zos_backup_restore.py
```diff
@@ -186,15 +186,15 @@
...
     description:
       - Override the default high level qualifier (HLQ) for temporary and backup
         data sets.
-      - The default HLQ is the Ansible user that executes the module and if
-        that is not available, then the value of C(TMPHLQ) is used.
+      - The default is the original HLQ fo the data sets if that is not available,
+        then the value of C(TMPHLQ) is used.
```
ketankelkar, line 197: we were talking about the default behavior of the tmp hlq offline, just wanted to confirm that this statement is accurate. 

also `fo` should be `of`.

plugins/modules/zos_backup_restore.py
```diff
@@ -270,7 +270,7 @@
     space_type: g
 
 - name: Restore data sets from backup stored in the UNIX file /tmp/temp_backup.dzp.
-    Use z/OS username as new HLQ.
+    With the original high level qualifiers.
```
ketankelkar, line 273: Maybe add some details to the 2nd sentence:
> Restore data sets from a backup stored in the UNIX file /tmp/temp_backup.dzp. Restore the data sets with the original high level qualifiers.

ketankelkar (changes requested):
this is shaping up very nicely, i had a few more suggestions:

ketankelkar:
looks like backup restore func tests are passing with the changes made here
image
image
image

ketankelkar (approved):
LGTM - thanks for making all the requested changes!!

rexemin (approved):

plugins/modules/zos_backup_restore.py
```diff
@@ -186,15 +186,15 @@
...
     type: str
     required: false
   tmp_hlq:
     description:
       - Override the default high level qualifier (HLQ) for temporary and backup
         data sets.
-      - The default HLQ is the Ansible user that executes the module and if
-        that is not available, then the value of C(TMPHLQ) is used.
+      - The default is the original HLQ of the data sets.
+        If that is not available, then the value of C(TMPHLQ) is used.
```
ddimatos, line 197: I am a bit confused by this and its not your doing, I think a long time ago when the module did not support a user entry for `hlq`, we would honor the ZOAU environment var `TMPHLQ` while we sort of do but not in this case I don't believe. 

So this:
```
 If that is not available, then the value of C(TMPHLQ) is used.
```
sort of contradicts this:
```
The default is the original HLQ of the data sets.
```

I would imagine if 'If that is not available' then the action is no HLQ is used meaning it falls back to `The default is the original HLQ of the data sets.`

plugins/modules/zos_backup_restore.py
```diff
@@ -282,8 +282,8 @@
     space: 1
     space_type: g
 
-- name: Restore data sets from backup stored in the UNIX file /tmp/temp_backup.dzp.
-    Use z/OS username as new HLQ.
+- name: Restore data sets from a backup stored in the UNIX file /tmp/temp_backup.dzp.
```
ddimatos, line 285: Shouldn't the behavior be that you restore data sets with ever is configure for `hlq`?

Something like:
```
Restore the data sets with the HLQ configured for option C(hlq), if the option was not set, the original HQL is used for restoration. 
```

ddimatos (changes requested):
@AndreMarcel99 - couple of doc related questions. Let me know if you have questions.

plugins/modules/zos_backup_restore.py
ketankelkar: i think the point here is to highlight two things, firstly that backups can be stored in USS,  secondly, since no hlq is set, the default behavior kicks in (hence the 2nd sentence in the description)
> Restore the data sets with the original high level qualifiers.

plugins/modules/zos_backup_restore.py
ketankelkar: i believe the desired default behavior from the zoau side of the house is somewhat nebulous and a jira has been (or will be). Figured in the mean time it'd be fine to keep the message close to what the original form was and tweak what has changed, ie default is original not user

plugins/modules/zos_backup_restore.py
AndreMarcel99: I agree with @ketankelkar

fernandofloresg (changes requested):

plugins/modules/zos_backup_restore.py
```diff
@@ -1018,11 +996,14 @@ def to_dunzip_args(**kwargs):
...
-    if kwargs.get("hlq"):
+    if kwargs.get("hlq") is None:
+        zoau_args["keep_original_hlq"] = True
+    else:
         zoau_args["high_level_qualifier"] = kwargs.get("hlq")
 
     if kwargs.get("tmp_hlq"):
-        zoau_args["tmphlq"] = str(kwargs.get("tmp_hlq"))
+        zoau_args["high_level_qualifier"] = str(kwargs.get("tmp_hlq"))
+        zoau_args["keep_original_hlq"] = False
```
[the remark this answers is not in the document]
AndreMarcel99: dzip do not support tmphlq.  To use tmphlq required change the variable on the system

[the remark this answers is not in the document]
AndreMarcel99: NAZARE-10728

fernandofloresg (approved):

ddimatos (approved):
@AndreMarcel99 this looks good to me, I think what happened here and partly to blame is we focused on `tmp_hlq`, the issue you are fixing is `hlq`, which addresses issue #1177, thus your changes appear to be sufficient. 

While I do think our docs need updating, you identified JIRA NAZARE-10728 which is related to `tmp_hlq` so lets keep these separate and reviewers can focus on just what the issue states needs correcting. 

I have created two issues, based on the `tmp_hlq`:
#1674 
#1675 

Thank you for your patience and understanding.

PR #1642, opened 2024-08-07
[Enabler][1595]Ensure_portability_zos_encode

### LINKED ISSUE #1595: [Enabler][Portability] [zos_encode] Ensure test suite portability by removing hard coded data.
### Enabler description

#### Issue:
- [ ] Just need to remove the variable assign to MVS_PS and similar, that could mislead us.

### Ansible module

zos_encode

AndreMarcel99:
##### SUMMARY
Remove hardcoded content of files and dataset names.

Fixes #1595 

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
Add date time for files names and delete hardcoded dataset names.

##### ADDITIONAL INFORMATION
Captura de pantalla 2024-08-08 a la(s) 11 14 16 a m
Captura de pantalla 2024-08-08 a la(s) 11 14 27 a m

### DIFF
--- a/tests/functional/modules/test_zos_encode_func.py
+++ b/tests/functional/modules/test_zos_encode_func.py
@@ -30,7 +30,7 @@ FROM_ENCODING = "IBM-1047"
 INVALID_ENCODING = "EBCDIC"
 TO_ENCODING = "ISO8859-1"
 TEMP_JCL_PATH = "/tmp/jcl"
-DATE_TIME = datetime.now().strftime("%H:%M:%S").replace("-", "")
+DATE_TIME = datetime.now().strftime("%H:%M:%S").replace("-", "").replace(":", "")
 TEST_DATA = """0001 This is for encode conversion testing_____________________________________
 0002 This is for encode conversion testing_____________________________________
 0003 This is for encode conversion testing_____________________________________
--- a/tests/functional/modules/test_zos_encode_func.py
+++ b/tests/functional/modules/test_zos_encode_func.py
@@ -14,6 +14,7 @@
 from __future__ import absolute_import, division, print_function
 from os import path
 from shellescape import quote
+from datetime import datetime
 # pylint: disable-next=import-error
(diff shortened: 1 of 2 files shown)


richp405 (approved):

tests/functional/modules/test_zos_encode_func.py
```diff
@@ -22,20 +23,14 @@
 __metaclass__ = type
 
 SHELL_EXECUTABLE = "/bin/sh"
-USS_FILE = "/tmp/encode_data"
 USS_NONE_FILE = "/tmp/none"
-USS_DEST_FILE = "/tmp/converted_data"
 USS_PATH = "/tmp/src"
 USS_DEST_PATH = "/tmp/dest"
```
ketankelkar, line 28: it might be advantageous to make the constants USS_PATH and USS_DEST_PATH unique per test case, or to have the test cases create and delete their own subdirectories. There are places in this test suite where the entire `USS_PATH` or `USS_DEST_PATH` directory are deleted in the finally clause. If we were ever to run multiple tests cases in parallel, we might run into the case where one test done using the directory and deletes it while the other test case is still using it

eg line 291:
```python3
hosts.all.file(path=USS_DEST_PATH, state="absent")
```

`USS_NONE_PATH` is likely a candidate for uniqueness too, it looks like the file is not supposed to exist initially, but then some test cases create it and delete it, which could interfere with other test cases
Same goes for `TEMP_JCL_PATH`

ketankelkar (changes requested):
These changes are shaping up very nicely! Thanks for handling that uss_file uniqueness issue ahead of time. 

Just had a few quick notes about some of the constants defined at the top. The main issue is these hard-coded constants point to files and directories which are removed in their entirety in the clean up phase in the finally clauses, which could lead to issues in running the test cases in parallel on the same zos targets

ketankelkar (approved):
re-ran the pipeline just for validation

image

image

image

PR #1643, opened 2024-08-07
Bugfix/531/add parameter to tolerate enqueue

AndreMarcel99:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/zos_backup_restore.py
+++ b/plugins/modules/zos_backup_restore.py
@@ -502,7 +502,7 @@ def parse_and_validate_args(params):
         full_volume=dict(type=full_volume_type, default=False, dependencies=["volume"]),
         temp_volume=dict(type="volume", required=False, aliases=["dest_volume"]),
         backup_name=dict(type=backup_name_type, required=False),
-        enqueue=dict(type=bool, required=False, default=False, dependencies=["operation"]),
+        enqueue=dict(type="bool", required=False, default=False, dependencies=["operation"]),
         recover=dict(type="bool", default=False),
         overwrite=dict(type="bool", default=False),
         sms_storage_class=dict(type=sms_type, required=False),
@@ -964,7 +964,7 @@ def to_dzip_args(**kwargs):
         zoau_args["tmphlq"] = str(kwargs.get("tmp_hlq"))
 
     if kwargs.get("enqueue"):
-        zoau_args["enqueue"] = "-f"
+        zoau_args["force"] = True
 
     return zoau_args
 
--- a/plugins/modules/zos_backup_restore.py
+++ b/plugins/modules/zos_backup_restore.py
@@ -122,6 +122,11 @@ options:
       - GDS relative names are supported when I(operation=restore).
     type: str
     required: True
+  enqueue:
+    description:
+      - When I(operation=backup), specifies potentially recoverable errors should be tolerated.
+    type: bool
+    required: False
   recover:
     description:
       - Specifies if potentially recoverable errors should be ignored.
@@ -369,6 +374,7 @@ def main():
         full_volume=dict(type="bool", default=False),
         temp_volume=dict(type="str", required=False, aliases=["dest_volume"]),
         backup_name=dict(type="str", required=True),
+        enqueue=dict(type="bool", required=False),
         recover=dict(type="bool", default=False),
         overwrite=dict(type="bool", default=False),
         sms_storage_class=dict(type="str", required=False),
@@ -388,6 +394,7 @@ def main():
         full_volume = params.get("full_volume")
         temp_volume = params.get("temp_volume")
         backup_name = params.get("backup_name")
+        enqueue = params.get("enqueue")
         recover = params.get("recover")
         overwrite = params.get("overwrite")
         sms_storage_class = params.get("sms_storage_class")
@@ -410,6 +417,7 @@ def main():
                 sms_storage_class=sms_storage_class,
                 sms_management_class=sms_management_class,
                 tmp_hlq=tmp_hlq,
+                enqueue=enqueue,
             )
         else:
             restore(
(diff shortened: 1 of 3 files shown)


richp405 (approved):
I wonder if you can run multiple commands in a shell call.. you have a few duplicate commands, like dtouch and echo>.... I wonder if && would speed it up, since there would be less hops to the test system.  Not a deal-breaker... just a thought on optimizing.

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -795,3 +821,40 @@ def test_backup_into_gds(ansible_zos_module, dstype):
     finally:
         hosts.all.shell(cmd=f"drm ANSIBLE.* ")
 
+
+def test_backup_tolerate_enqueue(ansible_zos_module):
+    hosts = ansible_zos_module
+    default_data_set_name_1 =  get_tmp_ds_name()
+    default_data_set_name_2 =  get_tmp_ds_name()
+    temp_file = f"/tmp/MEM2"
```
fernandofloresg, line 829: Looks like this variable is not used.

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -795,3 +821,40 @@ def test_backup_into_gds(ansible_zos_module, dstype):
...
+    data_sets_hlq = "ANSIBLE.**"
+    data_sets_backup_location = get_tmp_ds_name()
+    try:
+        hosts.all.shell(cmd="dtouch {0}".format(default_data_set_name_1))
+        hosts.all.shell(cmd="dtouch {0}".format(default_data_set_name_2))
+        hosts.all.shell(cmd="""decho "HELLO WORLD" "{0}" """.format(default_data_set_name_1))
+        hosts.all.shell(cmd="""decho "HELLO WORLD" "{0}" """.format(default_data_set_name_2))
+        hosts.all.file(path="/tmp/disp_shr/", state="directory")
```
fernandofloresg, line 837: Can you use temporary directories for testing? Since we are trying to improve portability and parallel execution this will make it easier.

tests/functional/modules/test_zos_backup_restore.py
```diff
@@ -30,6 +31,31 @@
...
+    sleep(300);
+    fclose(member);
+    return 0;
+}
+"""
+
+call_c_jcl="""//PDSELOCK JOB MSGCLASS=A,MSGLEVEL=(1,1),NOTIFY=&SYSUID,REGION=0M
+//LOCKMEM  EXEC PGM=BPXBATCH
+//STDPARM DD *
+SH /tmp/disp_shr/pdse-lock '{0}'
```
fernandofloresg, line 53: can you use the temp folder here?

fernandofloresg (commented):
by saying  "Our code already use flag -f in dzip function just weren't working properly." do you mean it was not working ZOAU and they fixed it now? So this test case will fail with 1.3.2?

fernandofloresg (changes requested):
Request above changes

AndreMarcel99:
Captura de pantalla 2024-08-13 a la(s) 3 26 36 p m
Captura de pantalla 2024-08-13 a la(s) 3 26 50 p m

fernandofloresg:
> This may seem overly picky, while I understand why you made it trivial as we really did no module changes, only added test case to validate the ZOAU fix, but if your a user how will you know this is fixed as `trivial` don't get pulled into our [CHANGELOG.rst](https://github.com/ansible-collections/ibm_zos_core/blob/dev/CHANGELOG.rst) or release_nores.rst?

Agree, while we are not doing the fix ourselves the requirements will force the user to use the ZOAU version that contains this fix, so the at the end of the day, the fix will be reflected in the product and is better to show that to the user.

changelogs/fragments/1643-Validate_parameter_recover_to_tolerate_enqueue.yml
fernandofloresg, line 1: I think this should be changed to bugfixes, also phrasing feels weird, using the bugfixes style might fit this better, first describe the problem in past tense and then the fix in present tense.

changelogs/fragments/1643-Validate_parameter_recover_to_tolerate_enqueue.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - zos_backup_restore - Module would fail when recover option is True not allowing
+    recoverable errors. Fix allow potentially recoverable errors should be ignored
```
fernandofloresg, line 3: just one suggestion
```suggestion
    recoverable errors. Fix allow potentially recoverable errors to be ignored
```

changelogs/fragments/1643-Validate_parameter_recover_to_tolerate_enqueue.yml
fernandofloresg, line 1: Can this be changed to :

```
When a recoverable error was encountered and ``recover=True``, the module would ignore the option and fail. Fix now does not fail when a recoverable error is raised and ``recover=True``.

```

fernandofloresg (changes requested):

fernandofloresg (approved):

rexemin (approved):

plugins/modules/zos_backup_restore.py
```diff
@@ -124,7 +124,7 @@
     required: True
   recover:
     description:
-      - Specifies if potentially recoverable errors should be ignored.
+      - When I(operation=backup) specifies if potentially recoverable errors should be ignored.
```
ddimatos, line 127: My comment is to expand the docs clarity, this comment really applies to all doc,  imagine you bring no experience and are reading this for the first time is how I approach doc. 
I see you referenced I(operation=backup) in the option which is great, my suggestion is below. 

Hence:
```
  recover:
    description:
      - When I(recover=true) and I(operation=backup) then potentially recoverable errors will be ignored.
    type: bool
    default: False
```

ddimatos (changes requested):
This is good work , I asked for one doc change , the test case addition is great, more work to create the test then anything else here :). 

I reviewed ZOAUs approach to this fix, seems like there is nothing for us to do other than validate it which you do in the test case, ZOAU is no longer delete dump data sets on non-zero return codes. So this looks good.

changelogs/fragments/1643-Validate_parameter_recover_to_tolerate_enqueue.yml
ddimatos: Resolving comment for @fernandofloresg - I validated the change was made.

ddimatos (approved):
Looks great, thanks for the micro edits I suggested.

PR #1644, opened 2024-08-08
[Bugfix] [v1.11.0 beta.1] [zos_apf] Backup when using special symbols

rexemin:
##### SUMMARY
Fixes a bug where the backup function would use outdated utilities to get the type of a dataset and fail when the name has a `$` in its name.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_apf

### DIFF
--- a/plugins/modules/zos_apf.py
+++ b/plugins/modules/zos_apf.py
@@ -292,6 +292,7 @@ backup_name:
     type: str
 '''
 
+import os
 import re
 import json
 from ansible.module_utils._text import to_text
@@ -312,7 +313,7 @@ except Exception:
 
 
 # supported data set types
-DS_TYPE = ['PS', 'PO']
+DS_TYPE = data_set.DataSet.MVS_SEQ.union(data_set.DataSet.MVS_PARTITIONED)
 
 
 def backupOper(module, src, backup, tmphlq=None):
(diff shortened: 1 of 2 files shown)


rexemin:
Screenshot 2024-08-08 at 9 10 40
Screenshot 2024-08-08 at 9 11 07

plugins/modules/zos_apf.py
```diff
@@ -340,11 +341,15 @@ def backupOper(module, src, backup, tmphlq=None):
...
+    file_type = None
+    if data_set.is_data_set(src):
+        file_type = data_set.DataSet.data_set_type(src)
+    else:
+        if os.path.exists(src):
+            file_type = 'USS'
+
     if file_type != 'USS' and file_type not in DS_TYPE:
-        message = "{0} data set type is NOT supported".format(str(file_type))
+        message = "{0} data set type of {1} is NOT supported".format(str(file_type), src)
```
ketankelkar, line 352: i think your format args _might_ be out of order:
- {0} --> `str(file_type)`. Say it resolves to VSAM,
- {1} --> `src`. Say it resolves `MY.VSAM.DATASET`, 

then the error message would be: 
> VSAM data set type of MY.VSAM.DATASET is NOT supported.

I'm guessing the intended error message would be:
> MY.VSAM.DATASET of type VSAM is NOT supported.

ketankelkar (changes requested):
one potential minor detail in the error message

fernandofloresg (approved):
other than ketan comment is ok for me

ketankelkar (approved):
lgtm

PR #1657, opened 2024-08-14
[Enabler][1594]test_fetch_portability

### LINKED ISSUE #1594: [Enabler][Portability] [zos_fetch] Ensure test suite portability by removing hard coded data.
### Enabler description

#### Issue:
- [ ] Test test_fetch_use_data_set_qualifier uses a hardcoded HLQ, it should be done dynamically.

### Ansible module

zos_fetch

AndreMarcel99:
##### SUMMARY

Fixes #1594 

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
zos_fetch

Captura de pantalla 2024-08-14 a la(s) 4 00 58 p m
Captura de pantalla 2024-08-14 a la(s) 4 01 11 p m


rexemin (approved):

richp405 (approved):

tests/functional/modules/test_zos_fetch_func.py
```diff
@@ -158,15 +167,16 @@ def test_fetch_uss_file_not_present_on_local_machine(ansible_zos_module):
...
+    with open(tmp_path, "w",encoding="utf-8") as file:
         file.close()
     hosts = ansible_zos_module
     params = {
         "src":"/etc/profile",
         "dest":"/tmp/",
         "flat":True
     }
-    dest_path = "/tmp/profile"
+    dest_path = tmp_path
```
fernandofloresg, line 179: Since dest_path is no longer `/tmp/profile` but `src` is `/tmp/profile` this now leaves behind a temporary file in `/tmp/profile`, you could just remove `/tmp/profile` but what if other test is using that ?

You may want to create a temporary file in the remote note or do this under a temporary folder in the local host. I strongly suggest the second option.

tests/functional/modules/test_zos_fetch_func.py
```diff
@@ -182,12 +192,12 @@ def test_fetch_uss_file_replace_on_local_machine(ansible_zos_module):
 
 def test_fetch_uss_file_present_on_local_machine(ansible_zos_module):
     hosts = ansible_zos_module
+    dest_path = get_unique_uss_file_name()
     params = {
         "src":"/etc/profile",
-        "dest":"/tmp/",
+        "dest": dest_path,
```
fernandofloresg, line 198: With this new change the test is no longer doing what it is supposed to, name is `test_fetch_uss_file_present_on_local_machine`, but the file that copies into is not present.

tests/functional/modules/test_zos_fetch_func.py
```diff
@@ -696,13 +705,13 @@ def test_fetch_partitioned_data_set_replace_on_local_machine(ansible_zos_module)
...
     hosts = ansible_zos_module
-    dest_path = "/tmp/profile"
+    dest_path = get_unique_uss_file_name()
     with open(dest_path, "w",encoding="utf-8") as dest_file:
         dest_file.close()
     os.chmod(dest_path, stat.S_IREAD)
     params = {
         "src":"/etc/profile",
-        "dest":"/tmp/",
+        "dest": dest_path,
```
fernandofloresg, line 714: When creating temporary files that are going to be used in the local node is better that we use `tempfile.NamedTemporaryFile` with `delete=True` and `delete_on_close=False`, that will delete the file once the context is closed, make sure to remove the file since is not automatically removed.

fernandofloresg (changes requested):
Just minor details, when using local files we can leverage tempfile instead of using `get_unique_uss_file_name` function, I only added the comment once but make sure to change it in other occurrences too.

AndreMarcel99:
Captura de pantalla 2024-08-22 a la(s) 2 25 25 p m
Captura de pantalla 2024-08-22 a la(s) 2 25 46 p m

fernandofloresg (approved):
lgtm!

PR #1664, opened 2024-08-23
[zos_copy] Enable zos_copy test cases portability

fernandofloresg:
##### SUMMARY

Updated zos_copy test cases to make sure that tests can be executed in any z/OS remote node and/or be executed in parallel. To enable test case parallelism created a function that returns a random valid unix name, it is used both for files and folders.
 

Also updated the following test suites : 
tests/functional/modules/test_zos_backup_restore.py
tests/functional/modules/test_zos_encode_func.py
tests/functional/modules/test_zos_fetch_func.py
tests/functional/modules/test_zos_job_submit_func.py

Those tests introduced a new function which was defined in each different case, now, these execute the same function with different prefixes and suffixes.

Fixes 1587

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME

zos_copy 

##### ADDITIONAL INFORMATION

```paste below

```


fernandofloresg:
Screenshot 2024-08-23 at 9 37 26 AM

zos_copy test cases.

fernandofloresg:
Screenshot 2024-08-23 at 12 21 16 PM

Screenshot 2024-08-23 at 12 21 39 PM

Executed other test suites that also moved to a this new utils function.

rexemin:
Since when do zos_copy tests take less than an hour to run? That's a miracle

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -426,7 +422,7 @@ def test_job_submit_pds(ansible_zos_module, location):
         results = None
         hosts = ansible_zos_module
         data_set_name = get_tmp_ds_name()
-        temp_path = get_unique_uss_file_name()
+        temp_path = get_random_file_name(dir=TMP_DIRECTORY, suffix='.dzp')
```
rexemin, line 425: I'm curious why you are using a suffix in this test suite that looks like you're creating compressed file

AndreMarcel99 (approved):
It will required the utils.py a new issue for adapt fetch and job submit test suite

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -2433,79 +2420,16 @@ def test_copy_ps_to_existing_uss_dir(ansible_zos_module):
...
 
     try:
-        hosts.all.zos_data_set(name=dest, state="absent")
+        hosts.all.shell(cmd=f"decho 'this is a test line' '{src_ds}'")
         copy_res = hosts.all.zos_copy(src=src_ds, dest=dest, remote_src=True)
         verify_copy = hosts.all.shell(
             cmd="cat \"//'{0}'\"".format(dest), executable=SHELL_EXECUTABLE
         )
 
-        # Copying the remote loadlibs in USS to a local dir.
```
rexemin, line 2446: Do we need to test this flow on a new test or can we just leave it out now? I remember this came about from a support issue and it was a bit contentious at the time, I could be wrong though

rexemin (commented):
Looks really good, thanks for taking Marcel's helper function and putting it in a helper that all test suites can now use. I just left two questions

tests/functional/modules/test_zos_job_submit_func.py
fernandofloresg: Hey sorry this should not be in zos_job_submit but rather in zos_backup_restore, will change it an run regression again

tests/functional/modules/test_zos_copy_func.py
fernandofloresg: oh it is tested in test_copy_local_pds_loadlib_to_pds_loadlib but this looks like a merge error, so I removed it.

rexemin (approved):

fernandofloresg:
> It will required the utils.py a new [issue](https://github.com/ansible-collections/ibm_zos_core/issues/1666) for adapt fetch and job submit test suite

Fetch and job submit are both in this PR, do you think this will also fix your issue #1666 ?

PR #1684, opened 2024-09-10
[bugfix][1678]Add validation for maker begin end

### LINKED ISSUE #1678: [Bug]Zos_blockinfile_not use marker_begin and end

### Bug description

When using a new marker_begin and marker_end in insert when removing the new markers are no passing to the absent case.
So these case will fail.
```
    - name: Remove text from Scripts
      zos_blockinfile:
        marker_begin: "NEW BEGIN"
        marker_end: "NEW END"
        marker: "* {mark} ANSIBLE MANAGED BLOCK"
        src: "{{ tmp_ds }}"
        state: absent
      ignore_errors: True

```

v1.11.0

v1.2.5 (default)

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

_No response_

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

AndreMarcel99:
##### SUMMARY
Add documentation, test case and validation for user

Fixes #1678 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Add validation
Captura de pantalla 2024-09-10 a la(s) 12 28 42 p m
Captura de pantalla 2024-09-10 a la(s) 12 28 58 p m

### DIFF
--- a/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
+++ b/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
@@ -1,4 +1,4 @@
 bugfixes:
   - zos_blockinfile - User could set same values for marker begin and end generate issue on ZOAU python API.
-    Change now catch the error and throw to the user a message with the problem.
+    Change now validates that ``marker_begin`` and ``marker_end`` are different and fails if they are the same.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1684).
--- a/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
+++ b/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - zos_blockinfile - User could set same values for marker begin and end generate issue on ZOAU python API.
-    Change now validates that ``marker_begin`` and ``marker_end`` are different and fails if they are the same.
+  - zos_blockinfile - Previously module was not able to delete a block when 'marker_begin' and 'marker_end'
+    were set to the same value. Fix introduces a requirement for 'marker_begin' and 'marker_end' to have different values.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1684).
--- a/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
+++ b/changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - zos_blockinfile - User could set same values for marker begin and end generate issue on ZOAU python API.
+    Change now catch the error and throw to the user a message with the problem.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1684).
--- a/plugins/modules/zos_blockinfile.py
+++ b/plugins/modules/zos_blockinfile.py
@@ -96,7 +96,7 @@ options:
   marker_begin:
     description:
     - This will be inserted at C({mark}) in the opening ansible block marker.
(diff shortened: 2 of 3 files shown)


richp405 (approved):

tests/functional/modules/test_zos_blockinfile_func.py
fernandofloresg, line 1: I don't think there is need for this new test case, looks like is doing the same as test_uss_block_absent_custommarker

fernandofloresg (changes requested):
Thank you for the quick addition for the validation, just a few corrections and don't think the test case is necessary as it is, rather, I would like to see a negative test case that checks the validation when both markers are the same, no need to setup a uss env, just call the module and validate the error.

plugins/modules/zos_blockinfile.py
```diff
@@ -96,13 +96,15 @@
   marker_begin:
     description:
     - This will be inserted at C({mark}) in the opening ansible block marker.
+    - Required unique value different from marker_end.
```
fernandofloresg, line 99: Thinking about this again, I don't think we should use word "required" because it can mislead the user to think this option is required when in fact is not. 

How about changing to ? 


```suggestion
    - Value needs to be different from marker_end.
```

fernandofloresg (changes requested):
Sorry, asking for a redundant change.

changelogs/fragments/1684-Add_validation_for_marker_begin_end.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - zos_blockinfile - User could set same values for marker begin and end generate issue on ZOAU python API.
+    Change now validates that ``marker_begin`` and ``marker_end`` are different and fails if they are the same.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1684).
```
ketankelkar, line 4: An Ansible user won't care particularly about bugs in ZOAU (or any other dependencies). Better to use this space to describe the bug as experienced by the Ansible user. 

> User could previously set the same values for module options 'marker_begin' and 'marker_end', which would result in <..whatever the issue was..>. Fix now introduces a requirement for 'marker_begin' and 'marker_end' to have different values.

Even better would be if you could bring the <..whatever the issue was..> to the beginning, 
> Previously <..issue..> happened when 'marker_begin' and 'marker_end' were set to the same value. Fix introduces a requirement for 'marker_begin' and 'marker_end' to have different values.

plugins/modules/zos_blockinfile.py
```diff
@@ -96,13 +96,15 @@
...
     - This will be inserted at C({mark}) in the opening ansible block marker.
+    - Value needs to be different from marker_end.
     required: false
     type: str
     default: BEGIN
   marker_end:
     required: false
     description:
     - This will be inserted at C({mark}) in the closing ansible block marker.
+    - Required unique value different from marker_begin.
```
ketankelkar, line 107: seems kind of strange to have a different description here than in marker_begin. I'd say pick one or the other and use the same line in both module options

If you go with "Value needs to be different from marker_end." I'd say use stronger language "Value must be different from marker_end."

fernandofloresg (changes requested):
Agree with @ketankelkar comments and request a change at least in the docs.

fernandofloresg (changes requested):

fernandofloresg (approved):
Requested a correction for a wrong dot, now committed it from GitHub to not have you get back to this.

PR #1695, opened 2024-09-18
[Bugfix] [module_utils] [multiple modules] tmp_hlq use in utils whenever mvscmd gets called

### LINKED ISSUE #912: [Bug] [module_utils] data_set_exists() does not allow for tmp_hlq

v1.2.3

### IBM Enterprise Python

v3.9.x

v1.7.0-beta.1

### ansible-version

v2.15.x

### z/OS version

v2.5

### Ansible module

_No response_

### Bug description

When looking over discussion #834 I noticed that the module_utils method `data_set_exists()` invokes `mvscmd` and passes commands for it to execute. When mvsmcd/mvscmdauth is invoked where no data sets are provided, mvscmd/mvscmdauth will place the command into a temporary data set on behalf of the caller (ansible). This poses a problem for users who's HLQ are limited. 

We currently support `tmp_hlq`,  likely it was overlooked that we needed to pass tmp_hlq when ever mvscmd is used without us passing data sets. 

This issue should also perform a quick scan of all code that invokes mvscmd or mvscmdauth to determine if tmp_hlq needs to be passed to the function. 

Testing:
This is likely outside the scope of this task (we might need a new issue for this),  ideally we should dynamically create a user (omvs segement) with limited privileges that is not permitted to create a HLQ to test this code. 

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

rexemin:
##### SUMMARY
Fixes #912.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
- module_utils/data_set.py

##### ADDITIONAL INFORMATION
Includes a lot of small changes just to pass the tmp_hlq used by modules all the way down to either `DataSet` or `DataSetUtils` where it can be used every time we make a call to `mvscmd`.


richp405 (approved):
Nice and thorough.

plugins/modules/zos_data_set.py
```diff
@@ -1415,9 +1415,9 @@ def perform_data_set_operations(data_set, state, replace, tmp_hlq, force):
     elif state == "absent":
         changed = data_set.ensure_absent()
     elif state == "cataloged":
-        changed = data_set.ensure_cataloged()
+        changed = data_set.ensure_cataloged(tmp_hlq=tmp_hlq)
     elif state == "uncataloged":
-        changed = data_set.ensure_uncataloged()
+        changed = data_set.ensure_uncataloged(tmp_hlq=tmp_hlq)
     return changed
```
ketankelkar, line 1421: how come the rest of these `ensure_*` methods don't get the `tmphlq` passed in? I noticed for example for `ensure_absent`, you added a default value of `tmphlq=None` in the function signature, but what if the user wants to define one, how does that value make it to the function call?

changelogs/fragments/1695-tmp_hlq_when_calling_mvscmd.yml
```diff
@@ -0,0 +1,31 @@
+bugfixes:
+  - zos_apf - The value of ``tmp_hlq`` would not be used when running
```
ketankelkar, line 2: this might be too technical of a description to read in the changelog / release notes. Think about how the user-experience has changed:

> The ``tmp_hlq`` module option was previously ignored and default values were used instead, fix now honors the value set in the module option.

^ Please edit the above for accuracy and even wordiness - the only point i'm trying to make is referencing the underlying mvs commands in the context of most of the modules here feels a bit too technical/deep in the code.

ketankelkar (changes requested):
really good work here, i liked your core fix to leveraging the `-Q` CLI option for specifying temporary high level qualifiers while we wait for the python api to become availble.
image

Change requests are:
- a note on the focus of the changelog 
- a question about potentially overlooked function calls in the data set module

plugins/modules/zos_data_set.py
rexemin: Are you only referring to the calls in zos_data_set or also to the functions in the data_set.py util?

The calls in zos_data_set only have `tmp_hlq` when they deal with datasets, the calls that check members don't end up using `mvscmd` and thus don't include the HLQ. For the specific call to `data_set.ensure_absent`, I included the HLQ at first, but then that caused errors while testing, so I removed it.

changelogs/fragments/1695-tmp_hlq_when_calling_mvscmd.yml
rexemin: Thanks for the correction, I wasn't sure myself of the original fragment, I changed it now for every module affected.

plugins/modules/zos_data_set.py
ketankelkar: I was referring to:
- `ensure_present`
- `ensure_absent`
- `ensure_cataloged` <-- you've got this one covered
- `ensure_uncataloged` <-- you've got this one covered

For `ensure_absent`, it calls `attempt_catalog_if_necessary_and_delete` which calls `data_set_cataloged` which calls `mvscmd`. 

For `ensure_present`, it calls `attempt_catalog_if_necessary` which calls `data_set_cataloged` which calls `mvscmd`.  

To me, it seems like both `ensure_present` and `ensure_absent` should pass the`tmphlq` all the way down the chain, otherwise, the module option would still be ignored for certain operations. 
Any chance that the errors you got were because the chain of calls wasn't completely updated?

ketankelkar (changes requested):
to me, it seems like `ensure_present` and `ensure_absent` in the data set module needs to accomodate tmphlq

plugins/modules/zos_data_set.py
rexemin: I'm not sure what other updates I made to the code, but now the calls don't end up with errors.

ketankelkar (approved):
LGTM - thanks for making the requested changes!

rexemin:
Screenshot 2024-10-02 at 11 20 51
Screenshot 2024-10-02 at 11 21 04
Screenshot 2024-10-02 at 11 21 20
Screenshot 2024-10-02 at 11 21 31
Screenshot 2024-10-02 at 11 21 43

rexemin:
Screenshot 2024-10-02 at 15 26 55
Screenshot 2024-10-02 at 15 27 05
Screenshot 2024-10-02 at 15 27 16

ketankelkar (approved):

PR #1723, opened 2024-09-26
[v1.9.3][zos_job_query] Return system and subsystem values when querying a job.

### LINKED ISSUE #1681: [Bug] [zos_job_query] Subsystem not returned in v1.8.0(and v1.11.0-beta.1) but 1.4.1 does return it

### Bug description

Job query was changed after 1.4.0, previously it would issue a REXX command and gather different fields, system(SYSNAME) and subsystem(ESYSID) were just a couple,  now that is retrieved using a parsing mechanism in the following line.

https://github.com/ansible-collections/ibm_zos_core/blob/42158ff2ed651dfe4f95b8af0bb1bfcfd915b916/plugins/module_utils/job.py#L346

We can get this output when using `zos_job_submit`, since version v1.8.0 `return_output` option was introduced in `zos_job_submit` it changed the way we get some values from system adding `dd_scan` into get_job_status interface, in short, we need to set the flag to true in the following line:

https://github.com/ansible-collections/ibm_zos_core/blob/ef7069ea7dc0ff743c8391270e358168d1fc07b2/plugins/module_utils/job.py#L243

and here 

https://github.com/ansible-collections/ibm_zos_core/blob/ef7069ea7dc0ff743c8391270e358168d1fc07b2/plugins/module_utils/job.py#L255

but that is a hot fix only, it will impact performance for sure, we should think if we want a new option in zos_job_query for getting more info about the job or just update this and make `zos_job_query` less performant but with a more complete output.

v1.11.0-beta.1

v1.3.2

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

zos_job_output, zos_job_query

---
- hosts: zvm
  collections:
    - ibm.ibm_zos_core
  gather_facts: false
  vars:
    ZOAU: "/zoau/v1.2.5.8" # to test with 1.4.1
    PYZ: "/allpython/3.11-3/usr/lpp/IBM/cyp/v3r11/pyz"

  tasks:
    # Test role
    - name: get job details
      zos_job_query:
        job_id: "JOB00339"
      register: debug_msg
    
    - name: Print out debug msg
      debug: 
        msg : "{{ debug_msg }} "

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

Enabled the dd scan when fetching a job status, while this is less performant it enables us to fetch both SYSTEM and NODE values from JES output.

We know that this won't be the same final implementation that will go into `dev` branch since we have requested ZOAU to provide these values when fetching the job list instead of us parsing the JES output.

Fixes #1681 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_job_query

### DIFF
--- a/plugins/modules/zos_job_query.py
+++ b/plugins/modules/zos_job_query.py
@@ -119,6 +119,16 @@ jobs:
          Unique job identifier assigned to the job by JES.
       type: str
       sample: JOB01427
+    system:
+      description:
+         The job entry system that MVS uses to do work.
+      type: str
+      sample: STL1
+    subsystem:
+      description:
+         The job entry subsystem that MVS uses to do work.
+      type: str
+      sample: STL1
     ret_code:
       description:
          Return code output collected from job log.
@@ -129,17 +139,6 @@ jobs:
             Return code or abend resulting from the job submission.
           type: str
           sample: CC 0000
-        msg_code:
-          description:
-            Return code extracted from the `msg` so that it can be evaluated.
-            For example, ABEND(S0C4) would yield "S0C4".
-          type: str
-          sample: S0C4
-        msg_txt:
-          description:
-             Returns additional information related to the job.
-          type: str
-          sample: "No job can be located with this job name: HELLO"
         code:
           description:
              Return code converted to integer value (when possible).
--- a/plugins/modules/zos_job_query.py
+++ b/plugins/modules/zos_job_query.py
@@ -139,6 +139,17 @@ jobs:
             Return code or abend resulting from the job submission.
           type: str
           sample: CC 0000
+        msg_code:
+          description:
+            Return code extracted from the `msg` so that it can be evaluated.
+            For example, ABEND(S0C4) would yield "S0C4".
+          type: str
+          sample: S0C4
+        msg_txt:
+          description:
+             Returns additional information related to the job.
+          type: str
+          sample: "No job can be located with this job name: HELLO"
         code:
           description:
              Return code converted to integer value (when possible).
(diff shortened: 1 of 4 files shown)


fernandofloresg:
Screenshot 2024-09-30 at 1 40 51 PM

plugins/modules/zos_job_query.py
```diff
@@ -129,17 +139,6 @@
             Return code or abend resulting from the job submission.
           type: str
           sample: CC 0000
-        msg_code:
```
rexemin, line 132: I'm curious why these attributes were removed from the module doc, were they never returned to begin with?

fernandofloresg: That is the reason, as of now, we do not return these values. So, for version 1.9.x I am removing them since we won't have any enhancements for this version. However, we can let them in the docs and if somebody needs them then add them as a bug fix.

Now that I'm reading this I'm tempted to keep them just in case.

@ketankelkar  @ddimatos  FYI

fernandofloresg:
@richp405 You are correct, this is not ideal, I'm glad you brought this up. We've discussed this during the last week, currently we have task #1694 to cover this implementation in the "correct" way, meaning, ZOAU will get this value and make it available for us in `jls`, that way we will not be scanning dds from our end.


This fix is for v1.9.x, and will be "temporary" for our dev branch if we put it in v1.13.0-beta.1, we can discuss either having this added to v1.13.0-beta.1 or wait for ZOAU implementation to not punish users with long output data or non-unicode chars.


Even, value for system are subject to change when ZOAU implements this. This fix is not getting the correct system name but we are ok with it because providing the correct system name would add a heavy overhead and might not even be correct, that is because we can not infer that the system doing the query is the same that executed the job. @ddimatos Has more details in this.

We can discuss this in 10/09 team call.

plugins/modules/zos_job_query.py
fernandofloresg: Added this back.

AndreMarcel99 (approved):

ketankelkar (approved):
nice work,

PR #1739, opened 2024-09-27
[Bugfix][zos_copy] Honor remote_tmp variable from ansible.cfg when creating temporary files

### LINKED ISSUE #1481: [Bug][zos_copy] Temporary files are set up in '/tmp/' with no means to override

### Bug description

TL;DR: `zos_copy` will always copy temporary files to `/tmp/`. If a user cannot read _and_ write there, the module fails.

The module builds the temporary path (`temp_path`) with a call to `gettempprefix()`, which is built-in to the python `tempfile` library ([python docs](https://docs.python.org/3/library/tempfile.html#tempfile.gettempprefix)) and always returns `'tmp'`.

```python3
temp_path = "/{0}/{1}".format(gettempprefix(), _create_temp_path_name())
```
[Link to src](https://github.com/ansible-collections/ibm_zos_core/blob/v1.8.0/plugins/action/zos_copy.py#L295)

Even when the `ansible_remote_tmp` ([doc](https://docs.ansible.com/ansible/latest/collections/environment_variables.html#envvar-ANSIBLE_REMOTE_TEMP)) variable is set in `ansible.cfg`, the module logic fails to check that configuration and builds the tempfile(s) under `/tmp`. 

The module should honor the configuration set in the `ansible_remote_tmp` variable and build all temporary dirs/files at the configured location. 

Find a way to set up `temp_path` so that the  value of `ansible_remote_tmp` is honored instead of effectively hard-coding `/tmp/...` into `temp_path`.

Reported in IBM internal SLACK thread `C0YPDHTGB/p1713974037809719`. 

v1.8.0 (default)

v1.2.5 (default)

### IBM Enterprise Python

v3.12.x

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

zos_copy

See SLACK thread (verbosity dump may contain some internal IBM info)

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

AndreMarcel99:
##### SUMMARY
Check the variable remote_tmp that can be accessed by any user to use on tmp files to copy and ensure creation of the folder.
Fixes #1481 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Check options of configuration and execute file module to ensure creation of folder.

##### ADDITIONAL INFORMATION
Captura de pantalla 2024-09-27 a la(s) 2 36 27 p m
Captura de pantalla 2024-09-27 a la(s) 2 36 37 p m

### DIFF
--- a/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
+++ b/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - zos_copy - Previously, the module ignored the variable of remote_tmp of the configuration file and used the
+    tmp folder. Change now use the variable remote_tmp variable. If it is not given by default, the value is .ansible/tmp.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1739).
--- a/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
+++ b/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
@@ -1,4 +1,4 @@
 bugfixes:
   - zos_copy - Previously, the module ignored the variable of remote_tmp of the configuration file and used the
-    tmp folder. Change now use the variable remote_tmp variable. If it is not given by default, the value is .ansible/tmp.
+    tmp folder. Fix now uses the variable remote_tmp variable. If it is not given, default value is ~/.ansible/tmp.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1739).
--- a/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
+++ b/changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
@@ -1,4 +1,5 @@
 bugfixes:
-  - zos_copy - Previously, the module ignored the variable of remote_tmp of the configuration file and used the
-    tmp folder. Fix now uses the variable remote_tmp variable. If it is not given, default value is ~/.ansible/tmp.
+  - zos_copy - Previously, the module ignored the value of ``remote_tmp`` set in Ansible configuration file
+    and used the ``/tmp/`` directory. Fix now uses the value of ``remote_tmp`` or the default value ``~/.ansible/tmp``
+    if none is given.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1739).
--- a/plugins/action/zos_copy.py
+++ b/plugins/action/zos_copy.py
@@ -267,6 +267,9 @@ class ActionModule(ActionBase):
         # Erasing all rendered Jinja2 templates from the controller.
         if template_dir:
             shutil.rmtree(template_dir, ignore_errors=True)
+        # Remove temporary directory from remote
+        path = os.path.normpath(f"{self.tmp_dir}/ansible-zos-copy")
+        self._connection.exec_command(f"rm -rf {path}*")
 
         if copy_res.get("note") and not force:
             result["note"] = copy_res.get("note")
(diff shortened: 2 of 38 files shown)


fernandofloresg:
This changes, at high level, ensure that the remote temporary folder used by the module is either the default used by ansible `~/.ansible/tmp` or whatever the user specified in `ansible.cfg` as `remote_tmp`.


### Updates to `plugins/action/zos_copy`
Modified copy_to_remote to get the proper temporary directory and make sure that it exists and user has access to it. If not, module will fail.
Also, this action plugin will now be cleaning any temporary files it creates.

### Updates to `plugins/module_utils/copy.py`.
Previously, the module used the `mvs_cmd.ikjeft01` to execute `OGET` command, the problem was that internally ZOAU would create a data set to store the command, resulting in the text being truncated. 

So, if user provided a temporary directory that was long i.e. more than 70 chars long, it would break.

Now, using `tsocmd` to execute the code will prevent this.

### Updates to `plugins/modules/zos_copy.py`

Now the module will get the temporary dir provided by the module init, using `module.tmpdir` it will access the class property. It will also set the env variable TMPDIR with that directory, hopefully, other utilities would use this, even if not, our code now uses it whenever we create temporary files on the remote.

fernandofloresg:
Screenshot 2024-10-04 at 2 16 30 PM

fernandofloresg:
with latest changes 
Screenshot 2024-10-04 at 3 14 03 PM

changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - zos_copy - Previously, the module ignored the variable of remote_tmp of the configuration file and used the
```
rexemin, line 2: I would change this fragment to:
> Previously, the module ignored the value of ``remote_tmp`` set in Ansible's configuration file and used the ``/tmp/`` directory. Fix now uses the value of ``remote_tmp`` or the default value ``~/.ansible/tmp`` if none is given.

plugins/action/zos_copy.py
```diff
@@ -293,8 +299,13 @@ def run(self, tmp=None, task_vars=None):
...
         """Copy a file or directory to the remote z/OS system """
-
-        temp_path = "/{0}/{1}/{2}".format(gettempprefix(), _create_temp_path_name(), os.path.basename(src))
+        self.tmp_dir = self._connection._shell._options.get("remote_tmp")
+        rc, stdout, stderr = self._connection.exec_command("cd {0} && pwd".format(self.tmp_dir))
+        if rc > 0:
+            msg = f"Failed to resolve remote temporary directory {self.tmp_dir}. Ensure that the directory exists and user has proper access."
+            return self._exit_action({}, msg, failed=True)
+        self.tmp_dir = stdout.decode("utf-8").replace("\r", "").replace("\n", "")
+        temp_path = "{0}/{1}/{2}".format(self.tmp_dir, _create_temp_path_name(), os.path.basename(src))
```
rexemin, line 308: Could this leave us with extra forward slashes (i.e. `/user/tmp//directory/src_basename`) if the user specifies something like this for the tmp dir: `/user/tmp/`?

If that is not an issue, this code is fine, otherwise I suggest using `temp_path = os.path.join(self.tmp_dir, _create_temp_path_name(), os.path.basename(src))`

plugins/modules/zos_copy.py
```diff
@@ -3140,7 +3140,8 @@ def normalize_line_endings(src, encoding=None):
         src_tag = encoding["from"]
 
     if src_tag != "IBM-037":
-        fd, converted_src = tempfile.mkstemp()
+        # fd, converted_src = tempfile.mkstemp(dir=os.environ['TMP'])
```
rexemin, line 3143: We could get rid of this identical commented line

rexemin (changes requested):
Looks really good, just left a couple of comments, but really only the fragment change is needed from me

plugins/modules/zos_copy.py
fernandofloresg: done

plugins/action/zos_copy.py
fernandofloresg: done

fernandofloresg:
Screenshot 2024-10-04 at 8 37 38 PM


Now with dev merged in :)

fernandofloresg:
Screenshot 2024-10-04 at 10 35 57 PM

now with dependency finder :)

richp405 (approved):
Other than Ivan's change to the description, this looks fine.

changelogs/fragments/1739-tmp_files_not_use_tmp_folder.yml
fernandofloresg: Done but changed Ansible's to Ansible

rexemin (approved):

Issue #1737, opened 2024-09-27
[Enhancement] [zos_job_submit] Update module using `jls` latest NOEXEC status for  jobs that run with TYPRUN=SCAN


fernandofloresg:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enhancement or feature description

ZOAU improved the `jls` status when jobs contain `TYPRUN=SCAN` where `jls` now returns a status of `NOEXEC` , prior it returned a status of `?`. 

They way our code currently supports `TYPRUN=SCAN` is by viewing a few heuristics with correct placement of the code in our jobs.py. In this issue, we can further solidify the detection of `TYPRUN=SCAN` by adding a check for status that equals `NOEXEC` along with our existing heuristics and code location. 
 
For reference, you can see the original PR #1283.
For reference, you can see issue #1272 for TYPRUN=COPY

ZOAU's new NOEXEC status is not enough to differentiate between TYPRUN=COPY and TYPRUN=SCAN, what I have noticed is the DDs vary between a COPY and SCAN, today for SCAN job we look for `JESJCL`, where if its a COPY, we can look to see that `JESJCL` does NOT exist (you can't look for JESJCLIN, ZOAU does not show it and it depends on the input queue). 

So, 
TYRUN=SCAN if `jls` status is NOEXEC and it DOES have `JESJCL`.
TYRUN=COPY if `jls` status is NOEXEC and it does NOT have `JESJCL`.

SCAN
```
Requests that the system scan this job’s JCL for syntax errors, without executing the job or allocating devices.
TYPRUN=SCAN checks the JCL only through the converter, not the interpreter. 
```

So it gets put on the input queue and then immediately written to sysout, thus it ends up with DDs `JESJCLIN` , `JESMSGLG` , `JESJCL`  and `$INTTEXT` when viewed in SDSF.

DD:
```
JESJCL	Indicates all job control statements in the input stream, that is, all JCL statements and JES2 control statements, plus all procedure statements from any in-stream or cataloged procedure a job step calls, plus all messages about job control statements.
```

COPY (JES2 only)
```
Requests that JES2 copy the input job stream, as submitted, directly to a sysout data set and schedule the sysout data set for output processing. The system does not schedule the job for execution.....
```
So it gets put on the input queue and then immediately written to sysout, thus it ends up with DDs `JESJCLIN` and `JESMSGLG` when viewed in SDSF.

DDs:
```
[...]

PR #1746, opened 2024-10-07
[v1.13.0][Enhancement] Add deploy and forget capability to zos_job_submit

### LINKED ISSUE #715: [Enhancement] [zos_job_submit] Deploy and forget capability
Currently, there is no provision to submit a job and continue on without waiting for a return code. This enhancement adds a provision to have the module submit the target JCL and only wait for confirmation of job submission before returning module success -- probably sufficient to capture the job number and return (the module already captures the job number). The user would leverage `zos_job_query` to keep tabs on job status as needed. Initial proposal is to have a user set `wait_time_s` to `0` to indicate that the module is to submit the job and not wait for a return code/job completion. The return structure design/values would likely need to change or be modified to handle module exits without a fully populated return code block -- perhaps it makes more sense to leave it out altogether rather than report back a temporary status

fernandofloresg:
##### SUMMARY

Add deploy and forget capability to zos_job_submit. When `wait_time_s` is equal to 0, the module will only submit the job and not wait to pick additional details like job name, ret_code and so on.

We will still return the values in the result dictionary, for that we create an empty schema to make sure we return those values as empty when we don't explicitly modify them.

Created a new test case that calls the module with `wait_time_s` as 0 (this would previously fail) and asserts that we the schema is in place.

Also modified a few things that changed when ZOAU implemented the TYPRUN=SCAN enhancement, however, they set NOEXEC status for both TYPRUN=SCAN and TYPRUN=COPY. So had to change the test case and add the entry into the docs.

Left some comments in the test case for TYPRUN that are intended to be addressed in #1737 

NOTE: **The PR seems a bit strange in GitHub but there are not as many changes as it may seem, I only indented some portions of the code with under an if. Read carefully and you may notice that I didn't remove much code other that a couple of lines, if it seems like it do search for the same line in the new added sections. is just an illusion.**

 
Fixes #715 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

zos_job_submit 

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1746-deploy-forget-zos_job_submit.yml
+++ b/changelogs/fragments/1746-deploy-forget-zos_job_submit.yml
@@ -1,0 +1,4 @@
+minor_changes:
+  - zos_job_submit - Add deploy and forget capability. Now when wait_time_s is 0, the module will
+    submit the job and will not wait to get the job details or content, returning only the job id.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1746).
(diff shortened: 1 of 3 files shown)


rexemin (approved):

fernandofloresg:
Screenshot 2024-10-15 at 12 32 00 PM

PR #1759, opened 2024-10-18
[Bug]Return system and subsystem in zos_job_query

### LINKED ISSUE #1681: [Bug] [zos_job_query] Subsystem not returned in v1.8.0(and v1.11.0-beta.1) but 1.4.1 does return it

fernandofloresg:
##### SUMMARY

Enabled the dd scan when fetching a job status, while this is less performant it enables us to fetch both SYSTEM and NODE values from JES output.

We know that this won't be the same final implementation that will go into dev branch since we have requested ZOAU to provide these values when fetching the job list instead of us parsing the JES output.

Same as #1723  but this time for dev branch.

Fixes #1681 

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1759-system-subsystem-job_query.yml
+++ b/changelogs/fragments/1759-system-subsystem-job_query.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - zos_job_query - Module was not returning values for system and subsystem. Fix now returns these values.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1759).
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -240,7 +240,6 @@ def job_status(job_id=None, owner=None, job_name=None, dd_name=None):
         job_id=job_id,
         owner=owner,
         job_name=job_name,
-        dd_scan=False
     )
 
     if len(job_status_result) == 0:
@@ -252,7 +251,6 @@ def job_status(job_id=None, owner=None, job_name=None, dd_name=None):
             job_id=job_id,
             owner=owner,
             job_name=job_name,
-            dd_scan=False
         )
 
     return job_status_result
--- a/plugins/modules/zos_job_query.py
+++ b/plugins/modules/zos_job_query.py
@@ -124,6 +124,16 @@ jobs:
          Type of address space used by the job.
       type: str
       sample: STC
+    system:
+      description:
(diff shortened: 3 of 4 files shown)


fernandofloresg:
Screenshot 2024-10-18 at 3 31 21 PM

richp405 (approved):

PR #1771, opened 2024-10-27
Staging v1.12.0 beta.1

ddimatos:
##### SUMMARY
Release version 1.12.0-beta.1 merge into main, addresses #1699


docs/source/modules/zos_fetch.rst
```diff
@@ -126,13 +126,13 @@ tmp_hlq
 
 
 ignore_sftp_stderr
-  During data transfer through sftp, the module fails if the sftp command directs any content to stderr. The user is able to override this behavior by setting this parameter to ``true``. By doing so, the module would essentially ignore the stderr stream produced by sftp and continue execution.
+  During data transfer through SFTP, the SFTP command directs content to stderr. By default, the module essentially ignores the stderr stream produced by SFTP and continues execution. The user is able to override this behavior by setting this parameter to ``false``. By doing so, any content written to stderr is considered an error by Ansible and will have module fail.
```
richp405, line 129: Last section: will have THE module fail, or will CAUSE THE module TO fail.

richp405 (approved):
A grammatical issue in some docs... no show-stoppers.  You are a merge maniac.

docs/source/modules/zos_fetch.rst
ddimatos: Thanks Rich, sending this up in a min.

ddimatos: Same text was also in zos_copy, fixed both.

ddimatos: Done.

plugins/modules/zos_blockinfile.py
```diff
@@ -96,13 +96,15 @@
   marker_begin:
     description:
     - This will be inserted at C({mark}) in the opening ansible block marker.
+    - Value needs to be different from marker_end.
```
fernandofloresg, line 99: I think this was mentioned by Ketan before, but wording is different between `marker_begin` and `marker_end`. The later says "Value must be different from marker_end" It would be nice to have the same wording.

fernandofloresg (approved):
What a piece of PR... left one comment which does not need to be changed, but rather is a nice to have.

plugins/modules/zos_blockinfile.py
ddimatos: Thanks for pointing this out, I fixed it and actually another doc issue.

PR #1780, opened 2024-11-01
[1.12.0][Bug][zos_mvs_raw] Bugfix/zos mvs raw verbose 1.12

fernandofloresg:
##### SUMMARY

Looks like mvccmd utility won't return anything in the stderr even if a program executed fails, so we only changed and if statement.

Changing that statement from 'and' to 'or' will in theory, fail any non-zero programs, which is better than having them fail only when verbose=True. Anyways, there is a task in the backlog to add a max_rc option to the module, so we can polish this mechanism there.

While testing this noticed that the module will hide the return code coming from the program if it fail, returning 8 as return code instead, fixed that and had to change some test cases that were expecting a wrong return code.

**NOTE: This was not originally added into the beta release because there was no time to test the fix with latest code. Good thing that we did, because we found and edge case. **

##### ISSUE TYPE

- Bugfix Pull Request
- 
##### COMPONENT NAME

zos_mvs_raw

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
+++ b/changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
@@ -2,7 +2,7 @@ bugfixes:
   - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed.
     Whereas, if the program failed and verbose was true the module would fail.
     Fix now has a consistent behavior and fails in both cases.
-    (https://github.com/ansible-collections/ibm_zos_core/pull/1774).
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1780).
   - zos_mvs_raw - Module would obfuscate the return code from the program when failing returning 8 instead.
     Fix now returns the proper return code from the program.
-    (https://github.com/ansible-collections/ibm_zos_core/pull/1774).
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1780).
--- a/changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
+++ b/changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
@@ -1,5 +1,5 @@
 bugfixes:
-  - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed.
+  - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed (false positive).
     Whereas, if the program failed and verbose was true the module would fail.
     Fix now has a consistent behavior and fails in both cases.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1780).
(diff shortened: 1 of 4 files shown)


fernandofloresg:
Screenshot 2024-11-01 at 4 31 13 PM
Screenshot 2024-11-01 at 4 30 57 PM

changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
```diff
@@ -0,0 +1,8 @@
+bugfixes:
+  - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed.
```
ketankelkar, line 2:
```suggestion
  - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed (false positive).
```

i'm not sure if this suggested adds any clarity (keep it if it does 😄)

changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
```diff
@@ -0,0 +1,8 @@
...
+    Whereas, if the program failed and verbose was true the module would fail.
```
ketankelkar, line 3: Does this sentence add anything? The module is supposed to fail when rc is nonzero, so this isn't describing a bug.
```suggestion
```

changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
```diff
@@ -0,0 +1,8 @@
...
+    Fix now has a consistent behavior and fails in both cases.
```
ketankelkar, line 4: If we do end up removing the 'whereas...' sentence, this sentence here could be:
```suggestion
    Fix now fails the module for all instances where a program has a non-zero return code.
```

ketankelkar (changes requested):
code changes look good, i had a few suggestions for the changelog, a few for comments in the changelog, and a potential mis-configured parameter. 

It's interesting that I've seen these changes 3 times now (from the other review) and only just now thought of suggestions here

changelogs/fragments/1780-zos_mvs_raw-verbose-fail.yml
fernandofloresg: it does! thanks

fernandofloresg:
@ketankelkar Thanks for your suggestions I ended up adding them all. Will run the test suite again and ask for a review after that is done.

richp405 (approved):
We have a similar issue in other functions, where we ended up letting the user select an 'acceptable' error code (I believe it was in submit).  What you did makes sense for mvs_raw.

fernandofloresg:
Screenshot 2024-11-04 at 11 18 39 AM

fernandofloresg:
closed it by accident

ddimatos (approved):
Looks good to me, looks like you found a legacy bug and fixed a new one and went with neither `and` and `or`. Thanks a bunch Fernando.

ketankelkar (approved):
LGTM - thanks for making the requested changes!

Issue #1748, opened 2024-10-14
[Enhancement] [zos_job_submit] Add support for Ansible async controls


rexemin:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enhancement or feature description

After the research done in #1743, it is feasible to implement support for async controls in zos_job_submit. This task should do the following:
- [x] Add async controls to zos_copy (as it is used to copy over local files in zos_job_submit)
- [x] Add async controls to zos_job_submit
- [x] Move cleanup from zos_job_submit's module to its action plugin
- [x] Remove the use of `ansible.builtin.tempfile` from zos_job_submit's action plugin

### Ansible module

zos_job_submit

rexemin:
Another sub-item for epic #1249.

fernandofloresg:
Fixed by #1789

PR #1786, opened 2024-11-04
[Enhancement] [zos_job_submit] Async support for zos_job_submit

### LINKED ISSUE #1249: [Enhancement] [Epic] [zos_job_submit,  zos_job_query, zos_job_output] support Ansible async controls

### Ansible module

zos_job_submit

The collection does not explicitly support Ansible `async` , when trying to use it with zos_job_submit the error `async is not supported for this task` appears. Further evaluation, to support `async` plugins specifically have to declare `self._supports_async = True` but its more than simply this entry, the module itself has to  deal with running in the async setting. 

This epic should have issues created:
- Research usage and impact of `self._supports_async = True` and answer the question if our current architecture can support this and if not what needs to change.
- Research how supporting `async` might impact a future enhancement where module `zos_job_submit` could accept a list of jobs to submit , the JSON responses were designed to accommodate more than one job submission, it would be good to understand if such a the multi job submission feature were might be limited by `ansync`, for example if both async and multiple jobs were supported we might need to disable async when running more than one job else asynchronously looping over a list of jobs might not lend itself to [poll](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_async.html#asynchronous-actions-and-polling). 
- Issues to track the development if research determines architecture supports asynch

For reference:
- see the [services.py](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/service.py) module: 
- [messages](https://groups.google.com/g/ansible-devel/c/cOFra1OTiII) from Brian
- Comments on [this](https://github.com/ansible/ansible/pull/47677/files) PR
- [Documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_async.html) using async
- Review [lopp_var](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#defining-inner-and-outer-variable-names-with-loop-var) controls , will help with evaluation 

An example which I can't fully evaluate because of the error:
zos_job_submit_loop_async.yml
```
---
- hosts: zvm
  collections:
    - ibm.ibm_zos_core
  gather_facts: false

  tasks:

    - name: Create 14 more copies of 1.jcl for demonstration
      ansible.builtin.copy:
        dest: "/tmp/{{ item }}.jcl"
        src: /tmp/1.jcl
      with_sequence: start=2 end=15
      delegate_to: localhost

    - name: Submit jobs asynchronously
      vars:
        jobs:
          - /tmp/1.jcl
          - /tmp/2.jcl
          - /tmp/3.jcl
          - /tmp/4.jcl
          - /tmp/5.jcl
          - /tmp/6.jcl
          - /tmp/7.jcl
          - /tmp/8.jcl
          - /tmp/9.jcl
          - /tmp/10.jcl
          - /tmp/11.jcl
          - /tmp/12.jcl
          - /tmp/13.jcl
          - /tmp/14.jcl
          - /tmp/15.jcl
        job: "{{ item }}"
      include_tasks: submit_jobs.yml
      loop: "{{ jobs | batch(5) | list }}"
```

submit_jobs.yml
```

    - name: Submit local jcl found on controller
      zos_job_submit:
        src: "{{ async_item }}"
        location: LOCAL
        encoding:
          from: ISO8859-1
          to: IBM-1047
        wait_time_s: 1
        use_template: false
      async: 600
      poll: 0
      loop: "{{ job }}"
      loop_control:
        loop_var: "async_item"
      ignore_errors: true
      register: result

    # - name: Wait until job completes
    #   zos_job_query:
    #     job_id: "{{ result.job_id }}"
    #   register: result
    #   until: result.jobs[0].ret_code.code == 0
    #   retries: 100
    #   delay: 2

    # - name: Results
    #   debug:
    #     var: result
    #   when: result.jobs[0].ret_code.code == 0
```

Error:
```
failed: [zvm] (item=/tmp/3.jcl) => {
    "ansible_loop_var": "async_item",
    "async_item": "/tmp/3.jcl",
    "msg": "async is not supported for this task."
}
```

Related issues:
- [ ] Issue [Enabler] [Research] [zos_job_submit, zos_job_query, zos_job_output] support Ansible async controls #1743

rexemin:
##### SUMMARY
Fixes #1249 and #1748.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- zos_job_submit

##### ADDITIONAL INFORMATION
In the end, zos_copy was not needed to be changed, zos_job_submit's action plugin can run zos_copy and block the rest of the execution flow until the local JCL file is uploaded to the remote host.


rexemin:
Results for Jenkins. First, zos_job_submit by itself, then the rest of the modules affected by the change.

Screenshot 2024-11-15 at 13 26 32
Screenshot 2024-11-15 at 13 26 41
Screenshot 2024-11-15 at 13 26 53
Screenshot 2024-11-15 at 13 27 05
Screenshot 2024-11-15 at 13 27 18

richp405 (approved):
Nice Upgrade!

plugins/action/zos_job_submit.py
fernandofloresg, line 1: Can you explain why the way a tmp dir is created was changed? Just want to understand the change.

plugins/action/zos_job_submit.py
```diff
@@ -129,36 +129,41 @@ def run(self, tmp=None, task_vars=None):
 
                 source_full = rendered_file
 
-            remote_path = None
-            remote_path = self._transfer_file(source_full, tmp_src)
-
-            if remote_path:
-                self._fixup_perms2((tmp, remote_path))
+            # remote_path = None
```
fernandofloresg, line 132: These can be removed rather than commented

plugins/action/zos_job_submit.py
```diff
@@ -101,9 +101,9 @@ def run(self, tmp=None, task_vars=None):
             # if self._connection._shell.path_has_trailing_slash(dest):
             #     dest_file = self._connection._shell.join_path(dest, source_rel)
             # else:
-            self._connection._shell.join_path(dest_path)
+            # self._connection._shell.join_path(dest_path)
 
-            tmp_src = self._connection._shell.join_path(tmp, "source")
+            # tmp_src = self._connection._shell.join_path(tmp, "source")
```
fernandofloresg, line 106: these can be removed too

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -1113,3 +1157,70 @@ def test_zoau_bugfix_invalid_utf8_chars(ansible_zos_module):
...
+            encoding='utf-8'
+        )
+
+        assert result.returncode == 0
+        assert "ok=2" in result.stdout
+        assert "changed=2" in result.stdout
+        assert result.stderr == ""
+
+    finally:
+        os.remove("inventory.yml")
```
fernandofloresg, line 1224: wondering if these two could have been NamedTemporaryFiles ?

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -381,6 +383,48 @@
...
+      ibm.ibm_zos_core.zos_job_submit:
+        src: {3}
+        location: local
+      async: 45
+      poll: 0
+      register: job_task
+
+    - name: Query async task.
+      async_status:
+        jid: "{{{{ job_task.ansible_job_id }}}}"
```
fernandofloresg, line 413: Can you share how a non async status would fail in this case?

fernandofloresg (changes requested):
Code is great. Is mostly questions I have

plugins/action/zos_job_submit.py
rexemin: The tempfile module cannot be run in async mode, so I had to change how we generate the tmp dir. It also means the module now uses the config file defined by the user.

fernandofloresg: ack!

fernandofloresg (approved):
lgtm

tests/functional/modules/test_zos_job_submit_func.py
rexemin: Yes, they can be, and here's the Jenkins pipeline testing that change
Screenshot 2024-11-19 at 10 01 47

AndreMarcel99 (approved):

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -1113,3 +1157,65 @@ def test_zoau_bugfix_invalid_utf8_chars(ansible_zos_module):
...
+    tmp_file = tempfile.NamedTemporaryFile(delete=True)
+    with open(tmp_file.name, "w",encoding="utf-8") as f:
+        f.write(JCL_FILE_CONTENTS)
+
+    # Getting all the info required to run the playbook.
+    path = get_config
+    with open(path, 'r') as file:
+        enviroment = yaml.safe_load(file)
+
+    ssh_key = enviroment["ssh_key"]
```
ddimatos, line 1173: Just an FYI, this will only work with the `zinventory` fixture because a user populates Jenkins with an SSH key, while fixture `zinventory-raw` has no support SSH keys currently so this test can not be part of the Concurrent executor (CE.py).

To support ssh keys, CE.py can be updated to take that as an argument and then update `zinventory-raw` to handle it, but honestly don't feel that is the right solution. Keys already exists on the host that can be looked up and passed along without user intervention or added to the hosts file which is what users.py does (actually generates a key and then adds it and removes it).

PR #1799, opened 2024-11-07
[v1.11.1][Bug][zos_mvs_raw] Multiple fixes

### LINKED ISSUE #1797: [v1.11.1][Bug][zos_mvs_raw] stdout is being replaced by stderr when rc=0 and verbosity is true

### Bug description

Code is replacing `stdout` with `stderr` when verbosity is true and return code is 0.

v1.11.0

v1.3.2

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

_No response_

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

### LINKED ISSUE #1798: [v1.11.1][Bug][zos_mvs_raw] Module false positive when RC is not zero and verbosity is false

### Bug description

If a program executes successfully with return code 0 and verbosity is enabled then the module will fail. This is happening with version 1.9.3 and was not discovered during testing.

This new test case has to be added.

```yaml
- hosts : zvm
  collections :
    - ibm.ibm_zos_core
  gather_facts: False
  vars:
    ZOAU: "/zoau/v1.3.3"
    PYZ: "/allpython/3.10/usr/lpp/IBM/cyp/v3r10/pyz"
    RC_0_DATA_SET_NAME: "USER.WHAT.EVER.ONE"
    DATA_SET_TO_LISTCAT: "USER.WHAT.EVER"
    IDCAMS_STDIN: " LISTCAT ENTRIES('{{ DATA_SET_TO_LISTCAT | upper }}')"
    # " LISTCAT NTRIES('OMVSADM.TEST')"

  tasks:

    - name: create some data set
      block:
        - name: create data set to LISTCAT
          zos_data_set:
            name: "{{ DATA_SET_TO_LISTCAT }}"
            type: seq
            state: present
            replace: yes

        - name: mvs raw with verbose with rc=0
          zos_mvs_raw:
            verbose: True
            program_name: idcams
            auth: True
            dds:
              - dd_data_set:
                  dd_name: SYSPRINT
                  data_set_name: "{{ RC_0_DATA_SET_NAME }}"
                  disposition: new
                  type: seq
                  return_content:
                    type: text
              - dd_input:
                  dd_name: SYSIN
                  content: "{{ IDCAMS_STDIN }}"
          register: output_rc_0
          ignore_errors: True

        - name: print output
          debug:
            msg:
              - "{{ output_rc_0 }}"

        - name: mvs raw with verbose with rc=0
          zos_mvs_raw:
            verbose: False
            program_name: idcams
            auth: True
            dds:
              - dd_data_set:
                  dd_name: SYSPRINT
                  data_set_name: "{{ RC_0_DATA_SET_NAME }}"
                  disposition: old
                  type: seq
              - dd_input:
                  dd_name: SYSIN
                  content: "{{ IDCAMS_STDIN }}"
          register: output_rc_0

        - name: print output
          debug:
            msg:
              - "{{ output_rc_0 }}"

      always:
        - name: delete data sets
          zos_data_set:
            name: "{{ item }}"
            state: absent
          loop:
            - "{{ DATA_SET_TO_LISTCAT }}"
            - "{{ RC_0_DATA_SET_NAME }}"
```

v1.11.0

v1.3.3

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

zos_mvs_raw

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

Fix False positive when RC != 0 and verbosity is false, fix not obfuscate program return codes when failing and stdout replaced by stderr.

## First bug: Module would return stderr in stdout when verbosity true and rc = 0

This fix was originally planned to be in v1.11.1 just to remove the following lines from the code:

```python
        if rc == 0 and verbose:
            out = err
```

This issue was reported by the CICS collection, after testing the module I was not able to find a case that returned some output in `stdout` when the command `mvscmd` ran successfully. So why was CICS experiencing issues with this change? That is because they are using `mvscmd.execute` function directly, now, our collection is not intended to be used as a library, but if they are using it is fine, we are not reverting this change to support them, we are reverting this change because it is legitimately a bug from users perspective, to get `stderr` content in `stdout` when `verbose` is true and return code is 0.

Because of that, I was not able to create a test case that didn't involve checking that `stdout` was not containing results from `stderr`, in my opinion that would be a redundant test case, because we always return the `stdout` in some `dd_output` inside `dd_names`.

However, I got Andrew Hughes from CICS collection team to test this branch and they are not having issues with it anymore, hence I'm not adding a test case for this. If you think differently, let me know and I can find a way to create a test case.

## Second bug 

Looks like mvccmd utility won't return anything in the stderr even if a program executed fails, so we only changed and if statement.

Changing that statement from 'and' to only fail any non-zero programs, is better than having them fail only when verbose=True. Anyways, there is a task in the backlog to add a max_rc option to the module, so we can polish this mechanism there.

## Third bug

While testing this noticed that the module will hide the return code coming from the program if it fail, returning 8 as return code instead, fixed that and had to change some test cases that were expecting a wrong return code.

Fixes #1797 
Fixes #1798 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_mvs_raw

### DIFF
--- a/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
+++ b/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
@@ -6,6 +6,6 @@ bugfixes:
   - zos_mvs_raw - Module would obfuscate the return code from the program when failing returning 8 instead.
     Fix now returns the proper return code from the program.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
-  - zos_mvs_raw - The module would return the stderr content in stdout when verbose was true and return code was 0.
+  - zos_mvs_raw - Module would return the stderr content in stdout when verbose was true and return code was 0.
     Fix now does not replace stdout content with stderr.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
--- a/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
+++ b/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
@@ -1,0 +1,8 @@
+bugfixes:
+  - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed.
+    Whereas, if the program failed and verbose was true the module would fail.
+    Fix now has a consistent behavior and fails in both cases.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
+  - zos_mvs_raw - Module would obfuscate the return code from the program when failing returning 8 instead.
+    Fix now returns the proper return code from the program.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
--- a/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
+++ b/changelogs/fragments/1799-zos_mvs_raw-verbose-fail.yml
@@ -1,8 +1,11 @@
 bugfixes:
   - zos_mvs_raw - If a program failed with a non-zero return code and verbose was false, the module would succeed.
-    Whereas, if the program failed and verbose was true the module would fail.
+    Whereas, if the program failed and verbose was true the module would fail(false positive).
     Fix now has a consistent behavior and fails in both cases.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
   - zos_mvs_raw - Module would obfuscate the return code from the program when failing returning 8 instead.
     Fix now returns the proper return code from the program.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1799).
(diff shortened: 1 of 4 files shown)


fernandofloresg:
Screenshot 2024-11-08 at 10 46 28 AM
Screenshot 2024-11-08 at 10 46 45 AM

AndreMarcel99 (approved):

rexemin (approved):

PR #1800, opened 2024-11-11
[v1.12.0][Bug][zos_mvs_raw] Remove stdout being replaced by stderr when RC=0 and verbosity is true

### LINKED ISSUE #1796: [v1.12.0][Bug][zos_mvs_raw] stdout is being replaced by stderr when rc=0 and verbosity is true

### Bug description

Code is replacing `stdout` with `stderr` when verbosity is true and return code is 0.

v1.12.0-beta.1

v1.3.2

### IBM Enterprise Python

v3.11.x (default)

### ansible-version

v2.16.x (default)

### z/OS version

v2.5 (default)

### Ansible module

_No response_

_No response_

### Ansible configuration.

_No response_

### Contents of the inventory

_No response_

_No response_

fernandofloresg:
##### SUMMARY

**This is the same fix as** #1794

Module would return stderr in stdout when verbosity true and rc = 0

This fix was originally planned to be in v1.11.1 just to remove the following lines from the code:

```python
        if rc == 0 and verbose:
            out = err
```

This issue was reported by the CICS collection, after testing the module I was not able to find a case that returned some output in `stdout` when the command `mvscmd` ran successfully. So why was CICS experiencing issues with this change? That is because they are using `mvscmd.execute` function directly, now, our collection is not intended to be used as a library, but if they are using it is fine, we are not reverting this change to support them, we are reverting this change because it is legitimately a bug from users perspective, to get `stderr` content in `stdout` when `verbose` is true and return code is 0.

Because of that, I was not able to create a test case that didn't involve checking that `stdout` was not containing results from `stderr`, in my opinion that would be a redundant test case, because we always return the `stdout` in some `dd_output` inside `dd_names`.

However, I got Andrew Hughes from CICS collection team to test this branch and they are not having issues with it anymore, hence I'm not adding a test case for this. If you think differently, let me know and I can find a way to create a test case.

Fixes #1796 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_mvs_raw

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1800-zos_mvs_raw-stdout-replaced-by-stderr.yml
+++ b/changelogs/fragments/1800-zos_mvs_raw-stdout-replaced-by-stderr.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - zos_mvs_raw - Module would return the stderr content in stdout when verbose was true and return code was 0.
+    Fix now does not replace stdout content with stderr.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1800).
+
--- a/plugins/module_utils/zos_mvs_raw.py
+++ b/plugins/module_utils/zos_mvs_raw.py
@@ -60,8 +60,6 @@ class MVSCmd(object):
             MVSCmd._build_command(pgm, dds, parm),
         )
         rc, out, err = module.run_command(command, errors='replace')
-        if rc == 0 and verbose:
-            out = err
         return MVSCmdResponse(rc, out, err)
 
     @staticmethod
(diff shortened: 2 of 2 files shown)


fernandofloresg:
Screenshot 2024-11-11 at 5 05 26 PM
Screenshot 2024-11-11 at 5 05 34 PM

richp405 (approved):

AndreMarcel99 (approved):

PR #1810, opened 2024-11-20
[Enhancement] [zos_copy] [zos_job_submit] [zos_script] Add option to disable autoescape for Jinja templates

### LINKED ISSUE #1198: [Enhancement] Disable jinja2 autoescape feature for some modules that use Ansible template

### Ansible module

zos_job_submit

With the `zos_job_submit` module when we use a template as source, HTML/XML special characters will be escaped because the jinja2 autoescape feature is enabled in the code. So this code for example:

Vars:
```
template_values:
  jobcard: "TESTTE JOB 'WD-TEST',MSGLEVEL=(1,1),MSGCLASS=R"
```

Template
```
//{{template_values.jobcard}}
```

Will be converted to:
```
//TESTTE JOB &#39;WD-TEST&#39;,MSGLEVEL=(1,1),MSGCLASS=R
```

We need to explicitly disable autoescaping with:

```
//{{template_values.jobcard | safe }}
```

Or

```
{% autoescape false -%}
//{{template_values.jobcard}}
{%- endautoescape %}
```

For  `zos_job_submit` autoescape should be disabled by default. Or we should have a way to disable it. In template_parameters for example.

rexemin:
##### SUMMARY
Fixes #1198.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- module_utils/template.py
- zos_copy
- zos_job_submit
- zos_script

##### ADDITIONAL INFORMATION
This adds a new option `autoescape` to all the modules that currently support templates. Its default value is `True` just to err on the side of caution, but it allows users to disable autoescaping whenever they need to.


rexemin:
Results for the pipeline

Screenshot 2024-11-22 at 18 13 48
Screenshot 2024-11-22 at 18 14 00

rexemin:
Screenshot 2024-11-22 at 18 13 48
Screenshot 2024-11-22 at 18 14 00

plugins/module_utils/template.py
```diff
@@ -213,26 +218,33 @@ def __init__(
...
-        )
+            'auto_reload': auto_reload
+        }
+
+        # Setting autoescape this way so bandit understands we're following best
+        # practices in regards to jinja autoescaping.
+        if autoescape:
+            self.templating_env = jinja2.Environment(autoescape=True, **environment_args)
+        else:
+            self.templating_env = jinja2.Environment(autoescape=jinja2.select_autoescape(), **environment_args)
```
ddimatos, line 247: Question, the branched logic here, `else`, is relying on `jinja2.select_autoescape()` which will return false as it is today, would it be safer to be explicit since this could change in the future where maybe it would be `true` thus if/else would return true should the behavior change.

Also note the documentation:
image

ddimatos (changes requested):
General questions:
1.  How does this impact users who have been piping values to safe?
2. We probably want to update the changelog fragment so it also includes a behavior change so the one releasing can properly highlight this. This could be a known issues type fragment. 

```
//{{template_values | safe }}
```

rexemin:
This change should not affect users already using the `safe` filter. Tested without issues with a template that used it.

plugins/module_utils/template.py
rexemin: Explicit in what way? I could change the function so it specifies the kinds of files that will commonly be templated, but just putting `false` in `autoescape` generates an error in Bandit.

fernandofloresg: @rexemin Is that the case if you use the following?

```suggestion
            self.templating_env = jinja2.Environment(autoescape=autoescape, **environment_args)
```
Just set autoescape parameter value to whatever autoescape variable has?

tests/functional/modules/test_zos_job_submit_func.py
```diff
@@ -158,6 +158,18 @@
 /*
 //SYSUT2   DD SYSOUT=*
 //
+""",
+    "Autoescape": """//{{ pgm_name }}    JOB (T043JM,JM00,1,0,0,0),{{ parameter }},CLASS=R,
+//             MSGCLASS=X,MSGLEVEL=1,NOTIFY=S0JM
```
fernandofloresg, line 163: In the issue related users complained both about `&` and `'` could you add those as part of the tests?

fernandofloresg: oh I see single quotes is added down there so just `&`

fernandofloresg (changes requested):
Just requesting a little addition to test

richp405 (approved):
lgtm, other than the document issue Demetrios mentioned.

plugins/module_utils/template.py
fernandofloresg: yup, is gonna trigger that scan issue no matter what. This is ok for me as long as we have a test which tests autoescape False and True. That means, if this gets changed in the future, then will be caught in the test cases.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1379,6 +1379,11 @@ def test_copy_template_file(ansible_zos_module, encoding):
             src=temp_template,
             dest=dest_path,
             use_template=True,
+            template_parameters={
+                # This should make sure that we don't get stuff like '&#39' when
+                # rendering quotes.
+                "autoescape": False
```
fernandofloresg, line 1385: Since we are adding a functionality that changes the file contents (escaping) we should also validate the file contents afterwards, for both escaping and not escaping.

fernandofloresg (approved):
lgtm

fernandofloresg:
Please check why some fragments are being added, maybe the branch has diverted a bit from dev

rexemin:
> Please check why some fragments are being added, maybe the branch has diverted a bit from dev

Done, it looks like it just needed to merge recent changes in dev.

plugins/module_utils/template.py
ddimatos: Sorry I had missed the alert to reply to your comment, it as @fernandofloresg had written , that is what I meant by explicit , forcing in a default value to the method rather than relying on what might change in the future. 

Optionally if we know of sure this is a false positive from our SAST (bandit) we can instruct bandit to ignore it.

ddimatos (approved):
LGTM

PR #1830, opened 2024-12-10
[Enabler][1589]volume_init_test_portability

### LINKED ISSUE #1589: [Enabler][Portability] [zos_volume_init] Ensure test suite portability by removing hard coded data.

### Enabler description

#### Issue:

- [ ] Use of IMSTESTL hlq data sets.

### Ansible module

zos_volume_init

AndreMarcel99:
##### SUMMARY
Remove hard coded volumes

Fixes #1589 

##### ISSUE TYPE

- Enabler Pull Request

##### ADDITIONAL INFORMATION

Right now there are few volumes could be use for the test case, but right now is available to check what volume have similar characteristics as USER02

Captura de pantalla 2024-12-09 a la(s) 7 30 25 p m
Captura de pantalla 2024-12-09 a la(s) 7 30 36 p m


rexemin (approved):

tests/functional/modules/test_zos_volume_init_func.py
```diff
@@ -16,37 +16,46 @@
...
     hosts = ansible_zos_module
-
     # remove all data sets from target volume. Expected to be the following 3
-    hosts.all.zos_data_set(name="IMSTESTL.IMS01.SPOOL1", state="absent")
-    hosts.all.zos_data_set(name="IMSTESTL.IMS01.SPOOL2", state="absent")
-    hosts.all.zos_data_set(name="IMSTESTL.IMS01.SPOOL3", state="absent")
+    volumes = Volume_Handler(volumes_on_systems)
+    volume_1 = volumes.get_available_vol()
+    volume_2 = volumes.get_available_vol()
+    clear_volume(hosts, volume_1)
```
richp405, line 48: do you need to check that volume_1 is not the same as volume_2?

tests/functional/modules/test_zos_volume_init_func.py
```diff
@@ -100,25 +101,62 @@ def test_index_param(ansible_zos_module, params):
...
+    hosts.all.zos_operator(cmd=f"vary {volume_2},online")
+
+
+def test_index_param_index_sms_false(ansible_zos_module, volumes_on_systems):
+    hosts = ansible_zos_module
+
+    volumes = Volume_Handler(volumes_on_systems)
+    volume_1 = volumes.get_available_vol()
+    volume_2 = volumes.get_available_vol()
+
```
richp405, line 113: Same question here: is there any issue if volume_1 == volume_2?  We used to have issues when we only had 1 testable volume.

richp405 (commented):
I'm not going to 'request changes', but point out the issue, because there might be a problem in the testing if only 1 volume is available at the time.  (you write to vol 1 after vary off of vol 2... just not sure if that's an issue here.)

tests/functional/modules/test_zos_volume_init_func.py
AndreMarcel99: At line 22 I already add comment that even if is not hardcoded, these two volumes that we use are still need to be added in our test_config file, because are so specific the characteristics of the volume

tests/functional/modules/test_zos_volume_init_func.py
AndreMarcel99: Same as the comment on the top, but also I create a [issue](https://github.com/ansible-collections/ibm_zos_core/issues/1832) for a new tool for this specific test suite

tests/functional/modules/test_zos_volume_init_func.py
richp405: I didn't notice the comment.  My bad.

richp405 (approved):

Issue #1837, opened 2024-12-16
[Module][zos_stat] Develop module for listing system stats 3


fernandofloresg:
### Is there an existing issue for this?

- [X] There are no existing issues.

### New module description

The new zos_stat module will be the z/OS counterpart to the stat module that comes built-in into Ansible. It will primarily deal with querying and returning the attributes associated with an MVS data set allocated in a remote host.

This issue spans development and testing work for the new module. Development will follow the proposed interface described in prior research.

### User story

The user will be able to write tasks using zos_stat to retrieve information about a data set in their remote system.

PR #1896, opened 2025-01-28
[Module] [zos_stat] New module zos_stat

### LINKED ISSUE #1835: [Module][zos_stat] Develop module for listing system stats 1

### New module description

The new zos_stat module will be the z/OS counterpart to the stat module that comes built-in into Ansible. It will primarily deal with querying and returning the attributes associated with an MVS data set allocated in a remote host, an aggregate mounted on it, or a file in USS.

This issue spans development and testing work for the new module. Development will follow the proposed interface described in prior research.

### User story

The user will be able to write tasks using zos_stat to retrieve information about a data set in their remote system.

rexemin:
##### SUMMARY
Fixes #1835, #1836 and #1837.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
zos_stat

##### ADDITIONAL INFORMATION
Please review the research document available in the dev-docs repo.


plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: zos_stat
+version_added: '1.14.0'
+author:
+  - "Ivan Moreno (@rexemin)"
+short_description: Retrieve facts from z/OS
```
fernandofloresg, line 25: This description is very similar to `zos_gather_facts` module,  which states "Gather z/OS system facts." I would recommend adding "from MVS data sets and USS files." or something more ad hoc to this module specifics

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+description:
+  - The L(zos_stat,./zos_stat.html) module retrieves facts from resources
+    stored in a z/OS system.
+  - Resources that can be queried are files, data sets and aggregates.
+options:
+  name:
```
fernandofloresg, line 31: In other modules we tend to use `src` as main file/data set option, so I think it would be nice if we use that one here too. Maybe we can keep name as alias to be similar to what the community does.

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+    description:
+        - Name of a data set or aggregate, or a file path, to query.
```
fernandofloresg, line 33: This sounds a bit weird to me, maybe we can change it to something like
```suggestion
        - Name of a data set, aggregate, or a file path to query.
```

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+        - Data sets can be sequential, partitioned (PDS), partitioned
+          extended (PDSE), VSAMs, generation data groups (GDG) or
+          generation data sets (GDS).
+    type: str
+    required: true
+  volume:
+    description:
+        - Name of the volume where the data set will be searched on.
+        - Required when getting facts from a data set. Ignored otherwise.
```
fernandofloresg, line 42: We can add here that multi-volume data sets are not supported.

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+        sets.
+      - The default HLQ is the Ansible user used to execute the module and
+        if that is not available, then the environment variable value
+        C(TMPHLQ) is used.
+    type: str
+    required: false
+
+notes:
+  - When querying data sets, the module will create a temporary data set
+    that requires around 4 kilobytes of available space on the remote host.
```
fernandofloresg, line 81: Changing this to be more ansible-like language.
```suggestion
    that requires around 4 kilobytes of available space on the managed node.
```

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+    This data set will be removed before the module finishes execution.
+  - Sometimes, the system could be unable to properly determine the
+    organization or record format of the data set or the space units used
+    to represent its allocation. When this happens, the values for these
+    fields will be null.
+  - When querying a partitioned data set (PDS), if the Ansible user has
+    RACF READ authority on it, the last referenced date will be updated by
+    the query operation.
+
+seealso:
```
fernandofloresg, line 91: We can also add `zos_gather_facts` here

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+stat:
+  description:
+    - Dictionary containing information about the resource.
+    - Attributes that don't apply to the current resource will still be
+      present on the dictionary with null values, so as to not break
+      automation that relies on certain fields to be available.
+  returned: success
+  type: dict
+  contains:
+    name:
```
fernandofloresg, line 140: If we change this to `src` make sure to update it here too.

plugins/modules/zos_stat.py
fernandofloresg: when looking down below the returned values I noticed "volumes" so looks like this is supported right ?

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+        seq_type:
+          description:
+            - Type of sequential data set (when it applies).
+            - Value can be either one of 'BASIC', 'LARGE' or 'EXTENDED'.
+          returned: success
+          type: str
+          sample: BASIC
+        data:
+          description:
+            - Dictionary containing attributes for the DATA part of a VSAM.
```
fernandofloresg, line 426: In VSAM these parts are called components.
```suggestion
            - Dictionary containing attributes for the DATA component of a VSAM. 
```

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,1685 @@
...
+            spanned:
+              description:
+                - Whether the data set allows records to be spanned
+                  across control intervals.
+              returned: success
+              type: bool
+              sample: false
+        index:
+          description:
+            - Dictionary containing attributes for the INDEX part of a VSAM.
```
fernandofloresg, line 473:
```suggestion
            - Dictionary containing attributes for the INDEX component of a VSAM. 
```

fernandofloresg (changes requested):
I went through the documentation, will review the code next time.

plugins/modules/zos_stat.py
rexemin: It will be supported, I'm adding that after I add files

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,2157 @@
...
+            'checksum_algorithm': {
+                'type': 'str',
+                'required': False,
+                'default': 'sha1',
+                'choices': ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
+            }
+        },
+        required_if=[
+            # Forcing a volume when querying data sets.
+            ('type', 'data_set', ('volume',))
```
fernandofloresg, line 2051: Was forcing a volume decided in the design phase of the module? I would think normal operation would be, we use the volume in the master catalog if the data set is cataloged.

docs/source/modules/zos_stat.rst
```diff
@@ -0,0 +1,670 @@
...
+      | **type**: str
+      | **sample**: PS
+
+    type
+      Type of the data set.
+
+      | **returned**: success
+      | **type**: str
+      | **sample**: LIBRARY
+
```
richp405, line 191: I think this should be lower case ("library" instead of "LIBRARY").

docs/source/modules/zos_stat.rst
```diff
@@ -0,0 +1,670 @@
...
+    record_format
+      Record format of a data set.
+
+      | **returned**: success
+      | **type**: str
+      | **sample**: VB
+
```
richp405, line 198: Same here... we changed parameters to be lower case.

docs/source/modules/zos_stat.rst
```diff
@@ -0,0 +1,670 @@
...
+      | **type**: str
+      | **sample**: 3390
+
+    space_units
+      Units used to describe sizes for the data set.
+
+      | **returned**: success
+      | **type**: str
+      | **sample**: TRACK
+
```
richp405, line 330: Sorry but the case thing is jumping out at me.  Also, I believe we used 'trk' for track measure in archive, backup, copy and others that take sizes.

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,2157 @@
...
+          description: Generic device type where the data set resides.
+          returned: success
+          type: str
+          sample: "3390"
+        space_units:
+          description: Units used to describe sizes for the data set.
+          returned: success
+          type: str
+          sample: TRACK
+        primary_space:
```
richp405, line 313: Again, I think this option should be lower-case.

richp405 (changes requested):
This is an AWESOME start!  The thing that jumped out at me we the case of some options... last I heard, we are turning thlngs like "size type" to all-lower (for choices), and upper-casing them where needed.

plugins/modules/zos_stat.py
rexemin: Yes, since not forcing the volume would entail first finding the data set, at which point the module starts to duplicate the functionality of zos_find. The expected workflow if a user doesn't know which volume a data set resides in, is to first use zos_find and then zos_stat.

plugins/modules/zos_stat.py
```diff
@@ -0,0 +1,2179 @@
...
+            'required': True,
+            'aliases': ['src']
+        },
+        'volume': {
+            'arg_type': 'str',
+            'required': False
+        },
+        'type': {
+            'arg_type': 'str',
+            'required': False
```
surendrababuravella, line 2090: "type" kept as optional but it is a mandatory field as per processing. And type=None scenario has not handled in get_facts_handler().

docs/source/modules/zos_stat.rst
rexemin: Thanks for reminding me of this, I changed the code last week so the module would return everything in lowercase, but I forgot about updating the documentation too.

richp405 (approved):
Looks better. Thanks!

richp405 (commented):
Thanks for the case changes!  I saw a comment about 'name should be source', which I would agree needs fixed.

fernandofloresg (changes requested):
just a request to add something to the docs and then I'll approve

fernandofloresg (approved):

fernandofloresg (approved):
LGTM ! Great to see this new module

PR #1900, opened 2025-01-31
[Enhancement][zos_job_query] Map system and subsystem from ZOAU reponse

surendrababuravella:
##### SUMMARY
Mapping system and subsystem values directly from zoau to zos_job_query response #1694 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_job_query

##### ADDITIONAL INFORMATION
zos_job_query response
```
{
        "changed": false,
        "failed": false,
        "jobs": [
            {
                "asid": 0,
                "content_type": "JOB",
                "creation_date": "2025-01-31",
                "creation_time": "3:12:05",
                "execution_time": "00:00:02",
                "job_class": "R",
                "job_id": "JOB00273",
                "job_name": "SLEEP",
                "owner": "USER",
                "priority": 1,
                "program_name": "BPXBATCH",
                "queue_position": 28,
                "ret_code": {
                    "code": 0,
                    "msg": "CC"
                },
                "subsystem": "JES2",
                "svc_class": null,
                "system": "ZOSSERVER"
            }
        ],
        "message": ""
    }
}
```

### DIFF
--- a/changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
+++ b/changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - zos_job_query - system and subsystem values are captured directly from zoau response and mapped in the modules response.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1900).
--- a/changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
+++ b/changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - zos_job_query - system and subsystem values are captured directly from zoau response and mapped in the modules response.
+  - zos_job_query - System and Subsystem are now retrieved from JES.
     (https://github.com/ansible-collections/ibm_zos_core/pull/1900).
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -376,11 +376,11 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, dd_scan=T
             job["ret_code"]["steps"] = []
             job["ddnames"] = []
             job["duration"] = duration
-            if hasattr(entry,"execution_time"):
+            if hasattr(entry, "execution_time"):
                 job["execution_time"] = entry.execution_time
-            if hasattr(entry,"system"):
+            if hasattr(entry, "system"):
                 job["system"] = entry.system
-            if hasattr(entry,"subsystem"):
+            if hasattr(entry, "subsystem"):
                 job["subsystem"] = entry.subsystem
 
             if dd_scan:
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -376,9 +376,12 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, dd_scan=T
             job["ret_code"]["steps"] = []
             job["ddnames"] = []
             job["duration"] = duration
-            job["execution_time"] = entry.execution_time
-            job["system"] = entry.system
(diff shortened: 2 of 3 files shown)


richp405 (approved):

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -124,12 +124,14 @@ def test_zos_job_name_query_multi_wildcards_func(ansible_zos_module):
...
             assert result.get("jobs")[0].get("ret_code").get("code") == 0
+            assert result.get("jobs")[0].get("execution_time") is not None
+            assert result.get("jobs")[0].get("system") is not None
+            assert result.get("jobs")[0].get("subsystem") is not None
 
             jobname = "HE*L*"
             qresults = hosts.all.zos_job_query(job_name=jobname, owner="*")
             for qresult in qresults.contacted.values():
                 assert qresult.get("jobs") is not None
-                assert qresult.get("jobs")[0].get("execution_time") is not None
```
fernandofloresg, line 132: We should keep this assertion because we are implementing this enhancement in `zos_job_query`, is a good thing that you added those assertions inside the `zos_job_submit` response, keep them, but make sure that we validate them in `zos_job_query` response too.

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -95,7 +96,6 @@ def test_zos_job_id_query_multi_wildcards_func(ansible_zos_module):
                 assert qresult.get("jobs") is not None
                 assert qresult.get("jobs")[0].get("system") is not None
                 assert qresult.get("jobs")[0].get("subsystem") is not None
-                assert qresult.get("jobs")[0].get("execution_time") is not None
```
fernandofloresg, line 98: Same comment as below, we should keep the assertions in `zos_job_query` response too.

plugins/module_utils/job.py
```diff
@@ -374,7 +376,12 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, dd_scan=T
             job["ret_code"]["steps"] = []
             job["ddnames"] = []
             job["duration"] = duration
-            job["execution_time"] = entry.execution_time
+            if hasattr(entry, "execution_time"):
+                job["execution_time"] = entry.execution_time
+            if hasattr(entry, "system"):
+                job["system"] = entry.system
+            if hasattr(entry, "subsystem"):
```
fernandofloresg, line 383: In this part looks like If we call the function with dd_scan=True it would now return a different value for system and subsystem like when we use it from `job_output` function, used by `zos_job_output` module.

That would create some inconsistencies in the results from one module to another, I suggest we add a check like if we already have populated "system" or "subsystem" do not fetch it from the dd_scan log.

fernandofloresg: This in line 488.

changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - zos_job_query - system and subsystem values are captured directly from zoau response and mapped in the modules response.
```
fernandofloresg, line 2: I would rephrase this since is a customer facing change.
```suggestion
  - zos_job_query - System and Subsystem are now retrieved from JES.
```

fernandofloresg (changes requested):

changelogs/fragments/1900-zos_job_query-map_system_subsystem.yml
ddimatos: You can review how to write a changelog fragement and its guidelines here. https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

fernandofloresg (approved):
LGTM!

Issue #1898, opened 2025-01-29
[Bug] Wrongly behavior for force_lock parameter to copy in PDS with $ in the name


vcorfmatibm:
### Is there an existing issue for this?

- [x] There are no existing issues.

### Bug description

There is an issue with zos_copy module with parameter force_lock set to False, when I want to copy in a PDS that is being used and if this PDS has $ in his name. In this case, the copy wrongly success even if the PDS is being used, but it should fail.
The same case with a PDS without $ in his name fails with error "Unable to write to dest '..........' because a task is accessing the data set."
Regarding the code, I saw that the data_set_locked(dataset_name) method returns False instead of True. The dataset_name in command_dgrs should be formatted by adding \ before $. I tested with this correction and it works.

### IBM z/OS Ansible core Version

v1.12.0

### IBM Z Open Automation Utilities

v1.3.4

### IBM Enterprise Python

v3.13.x

### ansible-version

v2.15.x

### z/OS version

v2.5

### Ansible module

zos_copy

### Ansible configuration.

```YAML
ansible [core 2.15.8]
  config file = None
  configured module search path = ['/home/vcorfmat/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/vcorfmat/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/vcorfmat/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/vcorfmat/.local/bin/ansible
  python version = 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

### Contents of the inventory

```YAML

```

### Contents of `group_vars` or `host_vars`

```YAML

```

fernandofloresg:
Fixed by #1908

PR #1907, opened 2025-02-05
[Bugfix][1.12.1][1898][zos_copy]Fix zos_copy failing when force_lock is true and data set name contains a special character

AndreMarcel99:
##### SUMMARY
When force_lock option is set to False (Default) use function to validate if the data set is locked, when the dataset name include special characters it fails. Change now escape special characters.

Fixer #1898 
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Escape dataset name and add simple quotes for calls on the shell for test suite.

Captura de pantalla 2025-02-05 a la(s) 2 01 26 p m
Captura de pantalla 2025-02-05 a la(s) 2 01 40 p m
Captura de pantalla 2025-02-05 a la(s) 2 35 40 p m

### DIFF
--- a/changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
+++ b/changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - zos_copy: Previously if the dataset name includes special characters such as $ fail when force_lock is false
+  - zos_copy - Previously if the dataset name includes special characters such as $ fail when force_lock is false
       in the validation. Change now allow the use of special characters.
       (https://github.com/ansible-collections/ibm_zos_core/pull/1907)
--- a/changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
+++ b/changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - zos_copy - Previously if the dataset name includes special characters such as $ fail when force_lock is false
-      in the validation. Change now allow the use of special characters.
+  - zos_copy - Previously, if the dataset name included special characters such as $, validation would fail when force_lock was false.
+      This has been changed to allow the use of special characters when force_lock option is false.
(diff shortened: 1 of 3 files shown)


changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
fernandofloresg, line 1: I suggest a rephrase to correct grammar mistakes : 

Previously, if the dataset name included special characters such as $, validation would fail when force_lock was true. This has been changed to allow the use of special characters when force_lock option is true.

tests/functional/modules/test_zos_copy_func.py
```diff
@@ -1976,7 +1976,7 @@ def test_ensure_copy_file_does_not_change_permission_on_dest(ansible_zos_module,
 def test_copy_dest_lock(ansible_zos_module, ds_type, f_lock ):
     hosts = ansible_zos_module
     data_set_1 = get_tmp_ds_name()
-    data_set_2 = get_tmp_ds_name()
+    data_set_2 = get_tmp_ds_name(symbols=True)
```
fernandofloresg, line 1979: symbols=True has the problem of being random, so you won't know for sure which special characters you are getting I would suggest to use something like

```suggestion
    data_set_2 = get_tmp_ds_name(llq_size=4)
    data_set_2 = f"{data_set_2}$#@"
```

Will will ensure every time we run this test case we are always testing the same special characters.

fernandofloresg (changes requested):
Nice solution, just requested a couple of minor changes

changelogs/fragments/1917-wrong_behaviour_force_lock_parameter.yml
AndreMarcel99: The problem is when force_lock was false

AndreMarcel99:
Captura de pantalla 2025-02-05 a la(s) 7 01 04 p m
Captura de pantalla 2025-02-05 a la(s) 7 01 23 p m

fernandofloresg (approved):
LGTM!

richp405 (approved):

Issue #1839, opened 2024-12-16
[Bug] [zos_job_query] Appears to use a wildcard when a job_id is used.


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Bug description
Part of Epic: [Enabler] [Epic] [zos_job_query] Performance evaluation and issue related tasks. #1841

This is a slight variation of the issue reported in #1764 , after reading the log it appears that `*` or just no value is sent to ZOAU causing all jobs to be quarried. At the time of writing this issue it appears this way from the log, I have not reviewed code, or forced an stdout to verify. 

In #1764, the task is below.
```
- name: "STC running check"
  delegate_to: zceeZosLpar
  ibm.ibm_zos_core.zos_job_query:
    job_id: "JOB04719"
  register: bcResult
```
The log confirmed the task arguments as:
```
{"module_args": {"job_name": "EDMCARD1", "owner": null, "job_id": null}
```

Yet ZOAU claimed it ran on the modules behalf:
```
jls -j -o owner,name,id,status,ccode,jobclass,serviceclass,priority,asid,creationdate,creationtime,queueposition,jobtype,programname --
```

If ZOAU is correct (else they have a bug) then our module is overly compensating and not only impacting performance, it could trigger SAF violations trying to access jobs that were not intended or instructed to do so. 

This issue will be part of a larger epic. 

### IBM z/OS Ansible core Version

v1.11.0

### IBM Z Open Automation Utilities

v1.3.1

### IBM Enterprise Python

v3.11.x

### ansible-version

v2.14.x

### z/OS version

v2.5

richp405:
There was an old existing behavior that we need to change, once I determine what versions of zoau can handle it... have code that specifically pulls jobs, then discards the ones not needed.  I believe this was an outgrowth of adding wildcards to the system.

I will add tracing to _get_job_status and lower, to make sure we can safely remove it ... checking jls shows that there is now support for wildcards.

richp405:
I found the gap... in job.py @300, the call to fetch-muiltiple was missing parameters, so the zoau code had no choice but to assume the search was for all jobnames & owners.

I corrected the code, and it runs, but I'm trying to enable logging to PROVE it is passing correctly now.

richp405:
There were 2 gaps (same issue)... this appears to tighten the query up a bit.  Will create PR tomorrow.

richp405:
[...]

PR #1916, opened 2025-02-10
[Bug] 1839: zos job query was using a wildcard

richp405:
##### SUMMARY
Customer reported issues where a job query would query *, which was against their policy.

Reading through line 300+ in module_utils/job.py showed that two of the parameters were ignored during the request, leading to inefficiency.  If you asked for a job_id, that is all that would return, but if you asked for job_owner or job_name, we asked for *, then filtered out what wasn't needed.  On a large system, this was a time killer.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_job_query was the outer-facing module, called by zos_job_submit et al.
The internal was module_utils/job.py

### DIFF
--- a/plugins/module_utils/job.py
+++ b/plugins/module_utils/job.py
@@ -320,23 +320,17 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, dd_scan=T
 
     final_entries = []
 
-    with open("/tmp/log.txt", "a") as fh:
-        basic = "\r\nTRACEIN 332: id {}, owner {}, job name{} temp {}\r\n"
-        fmt = basic.format(job_id, owner, job_name, job_id_temp )
-        fh.write( fmt )
-
-
-
     # In 1.3.0, include_extended has to be set to true so we get the program name for a job.
-    entries = jobs.fetch_multiple(job_id=job_id_temp, include_extended=True)
+    # entries = jobs.fetch_multiple(job_id=job_id_temp, include_extended=True)
+
+    # expanding > 1.3.0 of zoau, to include all params
+    entries = jobs.fetch_multiple(job_id=job_id_temp, job_owner=owner, job_name=job_name, include_extended=True)
 
     while ((entries is None or len(entries) == 0) and duration <= timeout):
         current_time = timer()
         duration = round(current_time - start_time)
         sleep(1)
         entries = jobs.fetch_multiple(job_id=job_id_temp, include_extended=True)
-
-    # print( "\r\nTRACEIN 332: id {job_id}, owner {owner}, job name{job_name} temp {job_id_temp}\y\n")
 
     if entries:
         for entry in entries:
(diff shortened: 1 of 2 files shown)


richp405:
Related to #1839

richp405:
image
image

plugins/module_utils/job.py
fernandofloresg, line 1: Please update the copyright year here too.

plugins/modules/zos_job_query.py
```diff
@@ -38,6 +38,8 @@
        - A job name can be up to 8 characters long.
        - The I(job_name) can contain include multiple wildcards.
        - The asterisk (`*`) wildcard will match zero or more specified characters.
+       - Note that using this value will query the system for '*' and then return just matching values
```
fernandofloresg, line 41: Added some points for consistency.
```suggestion
       - Note that using this value will query the system for '*' and then return just matching values.
```

plugins/modules/zos_job_query.py
```diff
@@ -38,6 +38,8 @@
...
+       - This may lead to security issues if there are read-access limitations on some users or jobs
```
fernandofloresg, line 42:
```suggestion
       - This may lead to security issues if there are read-access limitations on some users or jobs.
```

fernandofloresg: Plus question, what are the security issues? I imagine if a user don't have access to a job it would just not appear in the list returned by ZOAU. This is not a blocker or change request just a question.

fernandofloresg (changes requested):
Looks like we were missing this param for a long time, glad you added it. Can you also add a changelog entry? Also, be sure to test other zos_job* modules.

plugins/module_utils/job.py
richp405: Totally forgot this one.

plugins/modules/zos_job_query.py
richp405: Thanks for seeing that.

plugins/modules/zos_job_query.py
richp405: That was part of the original complaint: user got security warnings because the query was looking at jobs that were restricted from the ansible account.  It turned out when job.py was asking zoau for job_name="something", it was reduced to job_name='*'.  It's rare, but some job classes are not supposed to be viewed by all users (some kind of SAF setting).

changelogs/fragments/1916-job-status-query-on-job-name.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - module_utils/job.py - When searching for a job name, module performed a '*' (find all), then filtered the results.
```
fernandofloresg, line 2: Since this will be customer-facing is better that we don't refer to the internal utilities but rather, have an entry for each module like
```suggestion
  - zos_job_output - When searching for a job name, module performed a '*' (find all), then filtered the results.
```

and same for zos_job_query and zos_job_submit. Is just a change to avoid doing this rephrasing during release phase.

fernandofloresg (changes requested):

fernandofloresg (approved):
Thanks for updating the code. I really liked the description you gave in the changelog.

rexemin (approved):

PR #1933, opened 2025-02-18
When zos_blockinfile tries to introduce a block with double quotes generates a false negative 

### LINKED ISSUE #1258: [Bug]When zos_blockinfile tries to introduce a block with double quotes generates a false negative

Rohitcodes28:
##### SUMMARY

Fixes #1258.

The double quotes were not being handled properly so few new functions were introduced to the blockinfile code as a measure to handle only double quotes, Now that is fixed by the zoau team so this PR is to revert the changes made to handle double quotes. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_blockinfile.py

##### ADDITIONAL INFORMATION

A known issue is that the code for blockinfile is working as expected but the json output is giving a stderr 
[ZOAUResponse]
	rc: 0
	response_format: UTF-8
	stdout_response: {"data":{"changed":true,"found":0,"commands":["/eof/a\\newest \\\" quotes \\\" /$","$ a\\newest \\\" quotes \\\" "]},"program":"dmod"}

	stderr_response: BGYSC5309E Failed to insert new entry.
BGYSC5309E Failed to insert new entry.

	command: dmod -b -j  -s -e "/eof/a\newest \" quotes \" /$" -e "$ a\newest \" quotes \" " "omvsadm.test.seq" 

A separate ticket to handle the same is created by the zoau team NAZARE-10844

```paste below

```


Rohitcodes28:
Pipeline Executed successfully
Screenshot 2025-02-18 at 3 40 08 PM

richp405 (approved):
looks okay from here.

changelogs/fragments/1933-zos_blockinfile-false_negative_double_quotes.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
```
rexemin, line 1: I would change this fragment to a ``trivial`` one, since this change is not really affecting the current user experience (you're changing things behind the scenes, so to speak).

changelogs/fragments/1933-zos_blockinfile-false_negative_double_quotes.yml
```diff
@@ -0,0 +1,4 @@
...
+   - zos_blockinfile - Previously, double quotes were being handled separately as it was not supported by zoau.
```
rexemin, line 2: Text looks good, I would just write zoau in all caps, as ZOAU, since it's a product's name and an initialism.

rexemin (changes requested):
Changes look good, just had a couple of comments for the changelog fragment.

changelogs/fragments/1933-zos_blockinfile-false_negative_double_quotes.yml
fernandofloresg: I agree with Ivan's suggestion, since behavior is not going to change for the user we should set it as trivial. That doesn't make it less impactful though, you've done changes that will improve the codebase maintainability.

fernandofloresg (approved):

fernandofloresg (approved):
Looking good.

AndreMarcel99 (approved):

rexemin (approved):

Issue #1692, opened 2024-09-17
[Enhancement] [zos_script] Support automatic removal of line breaks [CR LF, LF, and CR]


ddimatos:
### Is there an existing issue for this?

- [X] There are no existing issues.

### Enhancement or feature description

The `zos_scripit` module does not automatically or even optionally handle removing line breaks:
CR LF  = `\r\n`
CR       = `\r`
LF        = `\n`

While another module `zos_copy` does by design handle line breaks, it does so normalizing the source on a preferred encoding, than scans for line breaks and removes them , returning the encoding back should it be not the optimized encoding used by the module. 

Why `zos_copy` is significant here is becasue `zos_script`'s plugin uses `zos_copy` to transfer content from the control node to the managed z/OS node and because `zos_copy` automatically handles line breaks, this should be inherent, yet `zos_copy` does have an issue open where line breaks don't get removed. 

The result of this enhancement should be that the module handle line breaks when source originates on the control node or USS. We may want to consider discussing if this should be automatic or not, how it affects python scripts and shell scripts. 

This issue should also include documentation explaining the experience one can expect when handling line breaks. 

### Ansible module

zos_script

fernandofloresg:
Fixed by #1954

PR #1954, opened 2025-02-25
[zos_script][zos_copy] Support automatic removal of carriage return characters (CR) when copying files 

surendrababuravella:
##### SUMMARY
automatic removal of line breaks(CR) while copying files using zos copy #1692

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
zos_copy, zos_script

##### ADDITIONAL INFORMATION
-> Implemented logic to remove CR while copying a file to remote directory.
-> Fixed issue related to the carriage return characters were being removed from only first 1KB of a file.
-> Added new test case in zos_script test cases for this use case.

### DIFF
--- a/changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
+++ b/changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - zos_copy - Carriage return will be removed automatically while copying files to remote server.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/1954).
--- a/changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
+++ b/changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
@@ -1,3 +1,7 @@
 minor_changes:
-  - zos_copy - Carriage return will be removed automatically while copying files to remote server.
-    (https://github.com/ansible-collections/ibm_zos_core/pull/1954).
+   - zos_script - Support automatic removal of carriage return line breaks [CR, CRLF] when copying local files to USS.
+     (https://github.com/ansible-collections/ibm_zos_core/pull/1954).
+bugfixes:
+   - zos_copy - the carriage return characters were being removed from only first 1024 bytes of a file. Now fixed that issue to support 
+     removal of the carriage return characters from the complete file content if the file size is more than 1024 bytes.
+      (https://github.com/ansible-collections/ibm_zos_core/pull/1954).
--- a/plugins/modules/zos_copy.py
+++ b/plugins/modules/zos_copy.py
@@ -1485,45 +1485,6 @@ class CopyHandler(object):
                 stderr=to_native(err)
             )
 
-    def remove_cr_endings(self, src):
-        """Creates a temporary file with the same content as src but without
-        carriage returns.
-
-        Parameters
-        ----------
-        src : str
-            Path to a USS source file.
-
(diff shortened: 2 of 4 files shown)


plugins/modules/zos_copy.py
```diff
@@ -1482,6 +1484,44 @@ def create_temp_with_lf_endings(self, src):
...
+        CopyOperationError
+            If the conversion fails.
+        """
+        try:
+            fd, converted_src = tempfile.mkstemp(dir=os.environ['TMPDIR'])
+            os.close(fd)
+
+            with open(converted_src, "wb") as converted_file:
+                with open(src, "rb") as src_file:
+                    chunk = src_file.read(1024)
```
fernandofloresg, line 1512: This is currently using 1024 bytes chunk (1KB), we've decided to read in chunks to avoid issues with large files but I think we can use something closer to MB instead. Maybe around 10-100MB would be a good start.

changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - zos_copy - Carriage return will be removed automatically while copying files to remote server.
```
fernandofloresg, line 2: Suggested a description more alike the issue description.

```suggestion
  - zos_script -Support automatic removal of carriage return line breaks [CR, CRLF] when copying local files to USS.
```

fernandofloresg: And I suggest you add another entry in this changelog for bugfixes where you state a bugfix for `zos_copy` that it was only removing the carriage return characters in the first 1024 bytes or first KB.

tests/functional/modules/test_zos_script_func.py
```diff
@@ -560,3 +574,32 @@ def test_rexx_script_with_args_remote_src(ansible_zos_module):
             os.remove(script_path)
         if os.path.exists(local_script):
             os.remove(local_script)
```
fernandofloresg, line 576: Can you add a newline here?

fernandofloresg (changes requested):
Requested a  couple of minor changes

plugins/modules/zos_copy.py
surendrababuravella: Changed size from 1KB to 32 MB

changelogs/fragments/1954-zos_copy-cr-removal-while-copying-file-to-remote.yml
surendrababuravella: Updated changeling as suggested

surendrababuravella:
Jenkins job details:
Screenshot 2025-02-27 at 5 09 33 PM

richp405 (approved):
looks okay from here.

fernandofloresg (approved):
Looks good to me. This new function inclusion may seen redundant but in tests surendra had issues with the original function due to encoding.

fernandofloresg:
Surendra is reviewing if he can avoid adding duplicate code.

fernandofloresg:
This first section tests everything that is dependent on zos_copy but zos_mount.

Screenshot 2025-03-02 at 9 54 21 PM

Screenshot 2025-03-02 at 9 54 36 PM

Screenshot 2025-03-02 at 9 56 31 PM


This other Jenkins run tests zos_mount only.

Screenshot 2025-03-02 at 9 59 50 PM

Screenshot 2025-03-02 at 9 57 34 PM
Screenshot 2025-03-02 at 9 59 29 PM

PR #1956, opened 2025-02-27
[zos_operator] Update zos_operator test cases with new way to get zoau version

Rohitcodes28:
##### SUMMARY

Test_zos_operator_func.py and test_zos_apf_func.py need the zoau version for a few test cases. 
As the test cases run locally and it does not have access to the zoautil_py module we can not use the zoau_version_checker.py included in the plugins/modue_utils to execute the test cases as well. 
A new helper is defined in the tests names version.py. Similar to the zoau_version_checker.py but it uses the command zoaversion to fetch the version info rather than the zoautil_py module. 
It help in local and pipeline testing. 

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME

- test_zos_apf_func.py
- test_zos_operator_func.py
- tests/helpers/version.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1956-zoau-version-support-for-test-cases.yml
+++ b/changelogs/fragments/1956-zoau-version-support-for-test-cases.yml
@@ -1,7 +1,7 @@
 trivial:
 
   - tests/functional/modules/test_zos_operator_func.py- test function test_zos_operator_positive_verbose_blocking
-    is enbled and working as expected now. 
+    is enabled and condition to version check is operational now. 
     (https://github.com/ansible-collections/ibm_zos_core/pull/1956).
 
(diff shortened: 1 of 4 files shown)


Rohitcodes28:
Pipeline tested and executed successfully

richp405 (approved):
I don't see any breaking code... I'm just a little confused about the premise... is the point of using the zoauversion call to FORCE the system to reach to a target site?

fernandofloresg (approved):
Looks good to me. We answered Rich question and yes, that is because we don't have ZOAU installed locally and we don't want to do it, because we only care about the version in the managed node.

AndreMarcel99 (approved):

PR #1969, opened 2025-03-10
[Documentation][copy]Update_documentation_about_alias_and_executable_on_copy

### LINKED ISSUE #1957: [Documentation] zos-copy - aliases needed for exectable module copy

Following discussion with Oscar Fernando Flores

" used zos_copy with executable: true 

Something I noticed and have to go back to the team is that in my tests, aliases were not copied unless aliases was specified, we would need to review and update the documentation accordingly."

PLease update documentation

v1.12.0

### Ansible module

zos_copy

AndreMarcel99:
##### SUMMARY
Update documentation on zos_copy option C(executable) do not copy aliases.

Fixes #1957 
##### ISSUE TYPE

- Documentation Pull Request

Captura de pantalla 2025-03-10 a la(s) 11 13 10 a m
Captura de pantalla 2025-03-10 a la(s) 11 13 20 a m


plugins/modules/zos_copy.py
```diff
@@ -229,14 +229,12 @@
   executable:
     description:
       - If set to C(true), indicates that the file or library to be copied is an executable.
-      - If the C(src) executable has an alias, the alias information is also copied. If the
```
fernandofloresg, line 232: We shouldn't remove the Unix info in this statement. Only the first sentence, and the rest makes no sense to keep it here but rather to add below line 246.

plugins/modules/zos_copy.py
```diff
@@ -229,14 +229,12 @@
...
-        C(dest) is Unix, the alias is not visible in Unix, even though the information is there and
-        will be visible if copied to a library.
       - If I(executable=true), and C(dest) is a data set, it must be a PDS or PDSE (library).
       - If C(dest) is a nonexistent data set, the library attributes assigned will be
         Undefined (U) record format with a record length of 0, block size of 32760 and the
         remaining attributes will be computed.
       - If C(dest) is a file, execute permission for the user will be added to the file (``u+x``).
+      - If the C(src) executable has an alias, the alias do not be copy unless C(aliases=true).
```
fernandofloresg, line 237:
```suggestion
      - If the C(src) executable has an alias, the alias will not be copied unless C(aliases=true).
```

fernandofloresg (changes requested):
Requested a couple of changes, there is no need to run the test cases in its entirety, you can just run one test case to make sure the file is not malformed.

richp405 (approved):

plugins/modules/zos_copy.py
AndreMarcel99: Done, but I think is better to work pass that information to the alias section

fernandofloresg: yeah that's what I meant by adding it to line 243

fernandofloresg (approved):
lgtm!

PR #1986, opened 2025-03-18
[Enabler] [zos_copy] Add retries when performing dest lock tests

### LINKED ISSUE #1972: [Enabler] Add retry for test copy_dest_lock in zos_copy

### Enabler description

When running zos_copy test suite the test `copy_dest_lock` might fail with the error reported in #1665. A mitigation technique we could use is to add a wrap the whole test case and make it so it can retry up to 3-5 times when we detect that error.
So, if error message is Unable to copy source and dest are identical we retry the whole copy_dest_lock test scenario up to 3 times. That way if we introduce a legit failure it will fail those 3 times. 

### Ansible module

zos_copy

rexemin:
##### SUMMARY
Fixes #1972.

##### ISSUE TYPE

- Enable Pull Request

##### COMPONENT NAME
tests/functional/modules/test_zos_copy_func.py

##### ADDITIONAL INFORMATION
When getting the error message described in #1972, the test suite will attempt up to another 4 times (for a total of 5 attempts) each dest lock test.


fernandofloresg (approved):
LGTM!

AndreMarcel99 (approved):

Issue #1959, opened 2025-02-28
[Enabler] Validate async support in modules throughout the collection


fernandofloresg:
### Is there an existing issue for this?

- [x] There are no existing issues.

### Enabler description

We had originally planned to add async support for many modules in the collection, turns out that modules that does not have an action plugin can use the `async` keyword and work asynchronously without any changes.

So, instead of us going through each module task and validate I created this task so one person can validate if all modules without action plugin work with `async` keyword.

Next, is the table of all modules and if they support or not the async mode 
| Module | Async supported out of the box | 
| --- | --- | 
| zos_apf – Add or remove libraries to Authorized Program Facility (APF) | No  |
| zos_archive – Archive files and data sets on z/OS. | No  |
| zos_backup_restore – Backup and restore data sets and volumes | No |
| zos_blockinfile – Manage block of multi-line textual data on z/OS | No |
| zos_copy – Copy data to z/OS | WIP | 
| zos_data_set – Manage data sets | No
| zos_encode – Perform encoding operations. | No |
| zos_fetch – Fetch data from z/OS | No |
| zos_find – Find matching data sets | No  |
| zos_gather_facts – Gather z/OS system facts.| No  |
| zos_job_output – Display job output | No  |
| zos_job_query – Query job status | No |
| zos_job_submit – Submit JCL | Yes  |
| zos_lineinfile – Manage textual data on z/OS | No  |
| zos_mount – Mount a z/OS file system. | No |
| zos_mvs_raw – Run a z/OS program. | No  |
| zos_operator – Execute operator command | No |
| zos_operator_action_query – Display messages requiring action | No |
| zos_ping – Ping z/OS and check dependencies. | No | 
| zos_script – Run scripts in z/OS | Yes |
| zos_tso_command – Execute TSO commands | No |
| zos_unarchive – Unarchive files and data sets in z/OS. | No |
| zos_volume_init – Initialize volumes or minidisks. | No |
| zos_zfs_resize | No |

zos_ping will probably not work with `async` because is not a python file, so if after validation it is still not working don't need to implement the support.

New playbooks will be written to validate this task, so the person validating this should upload those playbooks into the `dev-docs` repository.

- [ ] Uploaded docs into dev-docs repository.

Rohitcodes28:
The available modules have been tested for the async.  The ones added No (need ticket ) are the ones that are not supporting async and does not have a action plugin so needs to have. a ticket to resolve and add async support if required.

[...]

PR #1987, opened 2025-03-19
adding the async and action documentation to all modules.

Rohitcodes28:
##### SUMMARY

Adding the documentation regarding async and action support in all modules.

```
attributes:
  action:
    support: full
    description: Indicates this has a corresponding action plugin so some parts of the options can be executed on the controller.
  async:
    support: full
    description: Supports being used with the ``async`` keyword.
```

Modules having the action plugin support are marked support : full for action, rest all marked as none whereas its validated in [Async Validation](https://github.com/ansible-collections/ibm_zos_core/issues/1959)  that all modules not even having action plugins supports the async functionality. 

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


changelogs/fragments/1987-async_documentation.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
```
AndreMarcel99, line 1: For consistency, would be better to add for every module even is the same text in everyone
```
zos_apf - Add documentation for async and action support (<link for PR>)
zos_archive - ....
```

fernandofloresg (changes requested):
LGTM!

richp405 (approved):
Nice!

PR #2043, opened 2025-04-09
reducing the volume size so it can execute on any volume

Rohitcodes28:
##### SUMMARY

As the volume size is more than required in the test case,  this PR is to reduce the size of the cylinder.  Initially the volume size is 200 cylinders and we need minimum of 2 cylinders to test the test_basic_mount. We are keeping it to 10 so that the volumes can easily allocated  to the test case.

##### ISSUE TYPE

- Enabler Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/functional/modules/test_zos_mount_func.py
+++ b/tests/functional/modules/test_zos_mount_func.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) IBM Corporation 2020, 2024
+# Copyright (c) IBM Corporation 2020, 2025
 # Apache License, Version 2.0 (see https://opensource.org/licenses/Apache-2.0)
 
 from __future__ import absolute_import, division, print_function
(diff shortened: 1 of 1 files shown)


changelogs/fragments/2043-zos_mount-volume-size-resize.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
+  - test_zos_mount_func.py - Modifying the volume cylinder size, so test case does not fail due to volume unavailability.
```
fernandofloresg, line 2:
```suggestion
  - test_zos_mount_func.py - Modifying the volume cylinder size, so test case does not fail due to volume space constraints.
```

fernandofloresg (approved):
Suggested a minor change but approved.

changelogs/fragments/2043-zos_mount-volume-size-resize.yml
Rohitcodes28: Done !!.

Rohitcodes28:
Pipeline tested successfully for  10 cylinders [pipeline link ](https://billies1.fyre.ibm.com/job/ansible-zos-core-dev-1/613/parameters/)

richp405 (approved):

PR #2044, opened 2025-04-09
[New Module] zos_replace

### LINKED ISSUE #1994: [Epic] [zos_replace] Module development

### Enabler description

This epic represents the work of developing module `zos_replace` where the following user stories should be implemented.

- [ ] **Searching for strings**: As a user I want the ability to search for specific strings within a data set such that I can identify how frequently a string occurs
- [ ] **Replacing a string**: As a user I want the ability to replace an existing string with a newly provided text string such that a data set can be updated with new information
- [ ] **Replace a string globally**: As a user I want the ability to perform updates to all existing, matching strings given a back-reference like \g or \2 such that a data set with multiple instances of the same string can be universally updated.
- [ ] **Seeking before and after markers**: As a user I want the ability to restrict my search to certain sections within a data set based on a provided string such that I can leave portions of a data set untouched as needed.
- [ ] **Creating backups**: As a user I want the ability to take a quick snapshot of the data set before any actions are performed such that I can go back to the original state of the data set.
- [ ] **Performing validations**: As a user I want the ability to validate the modified data set before it replaces the original data set such that I can have certainty the replace operation is handled to my satisfaction.

Proposed module interface

after:

before:

encoding: from: to:

backup:

backup_name:

src:

regexp:

replace:

### Ansible module

_No response_

AndreMarcel99:
##### SUMMARY
Develop the module zos_replace to generate outputs as the community base on re python library and add options to the user for replace regular strings.

Fixes #1994 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
Module zos_replace with is code flow to open a file on binary mode to avoid any problem with special characters and maintain consistency  on the files given by the users.

##### ADDITIONAL INFORMATION
The test case ds_backup_no_name works locally, but not on the pipeline, work in progress.

Captura de pantalla 2025-05-12 a la(s) 3 24 46 p m
Captura de pantalla 2025-05-12 a la(s) 3 24 59 p m


richp405 (approved):
Why not use some of the test entries as examples in the doc?  Just a thought.

rexemin (changes requested):
Thanks for working on this, I'm apologizing in advance for the amount of comments I've made so far. Would also like to see the RST file for this module, if you have problems with using sphinx to generate the doc file, I can help you with that

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+            dataset = data_set.MVSDataSet(
+                name=name
+            )
+        except Exception:
+            messageDict = dict(msg="Unable to resolve name of data set {name}.")
+            module.fail_json(**messageDict, **results)
+
+        name = dataset.name
+
+        if data_set.DataSet.is_gds_relative_name(name):
```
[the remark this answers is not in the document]
AndreMarcel99: I have tested with (0) or (-1) and it works

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+        lit_rex = True if "regexp" in literal else False
+
+    if not bool(after) and not bool(before):
+        new_full_text, replaced = replace_text(content=decode_list, regexp=regexp, replace=replace, literal=lit_rex)
+        return new_full_text, replaced, new_full_text
+
+    begin_block_code, end_block_code, match = search_bf_af(text=decode_list, literal=literal, before=before, after=after)
+
+    if not match:
+        module.fail_json(msg="Pattern for before/after params did not match the given file.")
```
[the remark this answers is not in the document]
AndreMarcel99: The community module fail as a error.

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+    src = resolve_src_name(module=module, name=src, results=result, tmphlq=tmphlq)
+    uss = True if "/" in src else False
+    after = module.params.get("after")
+    before = module.params.get("before")
+    regexp = module.params.get("regexp")
+    replace = module.params.get("replace")
+    encoding = module.params.get("encoding")
+    backup = module.params.get("backup")
+    if parsed_args.get('backup_name') and backup:
+        backup = parsed_args.get('backup_name')
```
fernandofloresg, line 575: We don't have a reason to reuse `backup` instead of `backup_name`, it adds complexity down the road because we don't know if this is of type `str` or `bool` and hence we need to check everytime. Please use `backup_name` for this.

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+    result["target"] = src
+
+    if len(literal) > 0:
+        if "after" in literal and not after:
+            module.fail_json(msg="To use the option literal required the after parameter.", **result)
+        if "before" in literal and not before:
+            module.fail_json(msg="To use the option literal required the before parameter.", **result)
+
+    if backup:
+        if isinstance(backup, bool):
```
fernandofloresg, line 587: No need to do this if we use `backup_name`.

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+    Returns
+    ----------
+        list : List with the new text with the replace expected.
+    """
+    decode_list = open_file(file, encoding, uss)
+
+    lit_rex = False
+
+    if len(literal) > 0:
+        lit_rex = True if "regexp" in literal else False
```
fernandofloresg, line 497: I would change this variable name to:
```suggestion
        is_regex = True if "regexp" in literal else False
```

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+  zos_replace:
+    target: "/tmp/source"
+    regexp: '^MOUNTPOINT*'
+    after: export ZOAU_ROOT
+
+- name: Replace on SEQ full text
+  zos_replace:
+    target: "SEQ.SOURCE"
+    regexp: '^MOUNTPOINT*'
+    register: output
```
fernandofloresg, line 142: This should not be here.

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+
+- name: Replace words with backref on PDSE to delete some calls of SYSTEM
+  zos_replace:
+    target: PDS.SOURCE(MEM)
+    regexp: '^(.*?SYSTEM.*?)SYSTEM(.*)'
+    replace: '\1\2'
+    after: IEE133I PENDING *
+    before: IEF456I JOB12345 *
+
+- name: Replace using after on USS with no utf-8 characters
```
fernandofloresg: It doesn't make any difference in the task definition the fact that the USS has no utf-8 characters so no point in mentioning. Rather you could explain what this regexp is looking for

fernandofloresg (changes requested):
Left some more comments mainly about docs

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,646 @@
...
+    return modified_list, times_replaced
+
+
+def merge_text(original, replace, begin, end):
+    """Function to generate the new full text with the replace
+        text inside of it.
+    Args
+    ----------
+        original : list
+            Tue full original text on list
```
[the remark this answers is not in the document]
AndreMarcel99: The function open_file get the content on a list

tests/functional/modules/test_zos_replace_func.py
```diff
@@ -0,0 +1,1630 @@
...
+            assert result.get("changed") == True
+            assert result.get("target") == ds_full_name
+            assert result.get("found") == 1
+        results = hosts.all.shell(cmd="cat \"//'{0}'\" ".format(params["target"]))
+        for result in results.contacted.values():
+            assert result.get("stdout") == TEST_DISABLE_REGEX_CONTENT_AFTER_REGEXP
+    finally:
+        remove_ds_environment(ansible_zos_module, ds_name)
+
+# @pytest.mark.ds
```
rexemin, line 1337: Why is this one commented out?

rexemin (changes requested):

richp405 (approved):

plugins/modules/zos_replace.py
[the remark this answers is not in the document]
AndreMarcel99: In this case after resolve the name, even the alias name, I check once again to validate if is not solve it mean is still a invalid data set name.

surendrababuravella (approved):
Looks fine

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,714 @@
...
+            os.remove(tmp_file)
+
+    else:
+        full_text, replaced, fragment = replace_func(file=src, regexp=regexp, replace=replace, module=module, uss=uss,
+                                                     encoding=encoding, after=after, before=before, literal=literal)
+        try:
+            # zoau_io.zopen on mode w allow delete all the content inside the dataset allowing to write the new one
+            with zoau_io.zopen(f"//'{src}'", "w", encoding, recfm="*") as ds:
+                pass
+            for line in full_text:
```
fernandofloresg, line 657: If full_text is a list of ordered strings that make up the file contents can they be joined before writing into the data set ? this approach seems I/O intensive 

Like this 
```
content = [line.rstrip() for line in full_text] 
content = "\n".join(content)
rc_write = datasets.write(dataset_name=src, content=full_text, append=True)
```

didn't tried this but my point is, IO tends to be the costliest operation in computation, so we try to avoid it or minimize it

AndreMarcel99: zoau_io.zopen is better option we have right now to delete the content on a dataset to not remove the full dataset

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,716 @@
...
+        of a USS file or data set. If this parameter is not provided, this
+        module assumes that USS file or data set is encoded in IBM-1047.
+      - Supported character sets rely on the charset conversion utility (iconv)
+        version; the most common character sets are supported.
+    required: false
+    type: str
+    default: IBM-1047
+  disable_regex:
+    description:
+      - A list or string that allows the user to specify choices "before", "after", or "regexp" as regular strings instead of regex patterns.
```
ddimatos, line 79: I was a bit confused by this module option, the example cleared it up some, here are my thoughts but also think more needs to be put into this description, maybe when I get to the code I will have a better understanding. 

I suggest don't speak in third person "A list or string that allows the user to specify choices" , instead "A list or string that enables you to..." or just "enables....". 

Don't say choices when its not really a choice given its type is `raw` , find another way have the user use only one of these 3 strings `"before", "after", or "regexp"`, see below at my attempt to convey this. 

Since your type is raw, what happens if a user does this "disable_regex:"before, after, regexp", which is probably what you want them to think in terms of choices, likely this is not supported so we might need to be a bit more clear on this. 

Here is my attempt with not fully understanding this option, it's a bit confusing because I believe (without having looked at code) what this option is doing is disabling the regular expression processing, for example, you want to match this line but not expand the `*`, "regexp: //SYSPRINT DD SYSOUT=*"


```
      - A list or string that disables the regular expression processing for special characters. 
      - This is useful when the I(regex) to match is a string that contains special characters such as `*` that should not be expanded as a regular expression.
      - The supported options are "before", "after", or "regexp" as regular strings.
```

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,716 @@
...
+    required: false
+    type: raw
+  target:
+    description:
+      - The location can be a UNIX System Services (USS) file,
+        PS (sequential data set), member of a PDS or PDSE, PDS, PDSE.
```
ddimatos, line 85:
```suggestion
      - The location can be a UNIX System Services (USS) file,
        PS (sequential data set), PDS, PDSE, member of a PDS or PDSE.
```

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,716 @@
...
+      - The USS file must be an absolute pathname.
+      - It is possible to use a generation data set (GDS) relative name of generation already.
```
ddimatos, line 87: Remove period after already
```suggestion
      - It is possible to use a generation data set (GDS) relative name of generation already
```

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,716 @@
...
+    regexp: '^(.*?SYSTEM.*?)SYSTEM(.*)'
+    replace: '\1\2'
+    after: IEE133I PENDING *
+    before: IEF456I JOB12345 *
+"""
+
+RETURN = r"""
+backup_name:
+    description: Name of the backup file or data set that was created.
+    returned: if backup=true
```
ddimatos, line 176: This is a bit confusing, returned if backup=true yet in the documentation it states it will "be returned on either success or failure"
```
      - The backup file name will be returned on either success or failure of
        module execution such that data can be retrieved.
```

plugins/modules/zos_replace.py
```diff
@@ -0,0 +1,716 @@
...
+    returned: always
+    type: bool
+    sample: 1
+found:
+    description: Number of matches found
+    returned: success
+    type: int
+    sample: 5
+msg:
+    description: Error messages from the module
```
ddimatos, line 191: Is `msg` really used for **error** messages? I have not read down through the code but does not align to the Ansible doc for a module https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#msg

ddimatos (changes requested):
This looks great @AndreMarcel99 , I really like what you have done. All my comments are minor but should be addressed.

plugins/modules/zos_replace.py
fernandofloresg: @ddimatos Multi option in this is supported. Maybe a clearer way to say what this module does it that treats the fields as literal instead of regex expressions, that way if you have a complex regex that you don't want to waste time  escaping or don't have the skills to do so just disable the regex for those fields and the text searched for in the text "as is".

fernandofloresg: That could be the case, but remember `regex` is not the only value that accepts regular expressions, it is used to find the content to replace, but both `after` and `before` options allow the use of regex, so this way you can use all three options as literal or really any combination you want.

ddimatos: I see the name changed to `literal` that is good, it still feels like we might want to explain when a literal is a good choice to use, maybe not exactly this but something like below. 

This is useful when the I(regex) to match is a string that contains special characters such as `*` that should not be expanded as a regular expression.

Issue #2037, opened 2025-04-04
[Documentation] [zos_backup_restore] Add return values into zos_backup_restore


fernandofloresg:
### Is there an existing issue for this?

- [x] There are no existing issues.

### Documentation issue description

The return values for zos_backup_restore are missing in the documentation. You can take the existing values from the return output when running the module successfully.

![Image](https://github.com/user-attachments/assets/bc368381-796a-4f19-9e39-92e37ab426cd)

### IBM z/OS Ansible core Version

v1.13.0

### Ansible module

zos_backup_restore

PR #2049, opened 2025-04-10
[zos backup restore]Added Return Values in documentation

mayankmani-sde:
##### SUMMARY

Added Return values in documentation

#2037 
##### ISSUE TYPE

- Docs Pull Request
##### COMPONENT NAME

zos_backup_restore
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/2049-zos_backup_restore-added-return-values-in-doc.yml
+++ b/changelogs/fragments/2049-zos_backup_restore-added-return-values-in-doc.yml
@@ -1,4 +1,4 @@
 trivial:
   - zos_backup_restore - Updated documentation on zos_backup_restore to notify about return
-    values of Backup and restore operation
+    values of Backup and restore operation.
      (https://github.com/ansible-collections/ibm_zos_core/pull/2049).


richp405 (approved):

plugins/modules/zos_backup_restore.py
```diff
@@ -351,6 +349,28 @@
...
+backup_name:
+  description:
+    - The backup file or data set name that was processed.
+    - Matches the I(backup_name) parameter provided as input.
+  returned: always
+  type: str
+  sample: "/u/oeusr03/my_backup.dzp"
+message:
+  description:
+    - Empty ("") for successfull and failure operations.
```
fernandofloresg, line 369: Even though this is true at the moment, we don't want to tell users "oh by the way this does nothing". I think is better to say:
```suggestion
    - Returns any important messages about the modules execution, if any.
```

At this point it might be empty, but could be used for something in the future. The truth is it won't be, this looks like an attempt to define `msg` this module is old and original developer may have made a mistake. This is something we can remove in the future releases as a deprecation.

But at this point lets use this phrase instead.

plugins/modules/zos_backup_restore.py
```diff
@@ -351,6 +349,28 @@
...
+changed:
+  description:
+    - Indicates if the operation made changes.
+    - C(true) when backup/restore was successful, C(false) otherwise.
+  returned: always
+  type: bool
+  sample: true
+backup_name:
+  description:
+    - The backup file or data set name that was processed.
```
fernandofloresg, line 362:
```suggestion
    - The USS file name or data set name that was used as a backup.
```

fernandofloresg: I think this should be clearer

fernandofloresg (changes requested):
Requested a couple of phrasing changes

plugins/modules/zos_backup_restore.py
mayankmani-sde: Changes have been added .

richp405 (approved):

fernandofloresg (approved):
lgtm!

PR #2059, opened 2025-04-21
Enabler/1819/gh/sanity2-18

Rohitcodes28:
##### SUMMARY

Testing the ac sanity for 2.18 

##### ISSUE TYPE

- Enabler Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/ac-ansible-test-sanity.yml
+++ b/.github/workflows/ac-ansible-test-sanity.yml
@@ -51,23 +51,23 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.11
+        python-version: 3.12
   
     - name: Set up venv
       run: |
         python -m pip install --upgrade pip
         pip install virtualenv
         mkdir venv
-        virtualenv venv/venv-2.16
+        virtualenv venv/venv-2.18
 
     - name: Install dependencies
       run: |
-        source venv/venv-2.16/bin/activate
+        source venv/venv-2.18/bin/activate
         python -m pip install --upgrade pip
-        pip install ansible-core==2.17.6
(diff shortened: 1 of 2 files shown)


richp405 (approved):
While the current is 2.18.4, I think having 2.18 should be fine.  I haven't seen portability notes to say otherwise.

changelogs/fragments/2059-Github-sanity-2-18-fix.yml
```diff
@@ -0,0 +1,4 @@
+trivial:
+  - ac-ansible-test-sanity.yml - Updated the ansible version to 2.18,
+    venv to 2.18, python to 2.12.
```
rexemin, line 3:
```suggestion
    venv to 2.18, python to 3.12.
```
Although it probably doesn't really matter since it's a trivial change.

.github/workflows/ac-ansible-test-sanity.yml
```diff
@@ -51,23 +51,23 @@ jobs:
...
-        virtualenv venv/venv-2.16
+        virtualenv venv/venv-2.18
 
     - name: Install dependencies
       run: |
-        source venv/venv-2.16/bin/activate
+        source venv/venv-2.18/bin/activate
         python -m pip install --upgrade pip
-        pip install ansible-core==2.17.6
+        pip install ansible-core==2.18
```
rexemin, line 67: Maybe we could install the latest minor version of 2.18, but agree with Rich that there shouldn't be much of a difference between versions of Ansible 2.18.

```suggestion
        pip install "ansible-core>=2.18,<2.19"
```

rexemin (approved):

Issue #1940, opened 2025-02-20
[Bug] zos_mount failing during a df call.


richp405:
### Is there an existing issue for this?

- [x] There are no existing issues.

### Bug description

Internal customer noticed zos_mount was failing on certain systems.  Further logging showed an error coming when 'df' was called to enumerate available mounts.

Talking with IBM experts, it appears that the specific error message: FSUMF168 is effectively a warning, and the data in the response should still be complete, meaning we can depend on it for the list of errors.

On a custom copy of zos I modified the mount program:

![Image](https://github.com/user-attachments/assets/985d0711-7887-4147-8d12-0f6984d69b80)

`ibm_zos_core/plugins/modules/zos_mount.py\", line 1075, in run_module\n  File \"<frozen codecs>\", line 332, in decode\nUnicodeDecodeError: 'utf-8' codec can't decode byte 0xe3 in position 3: invalid continuation byte\ndebug3: mux_client_read_packet_timeout: read header failed: Broken pipe\r\ndebug2: Received exit status from master 1\r\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE: No start of json char found\nSee stdout/stderr for the exact error",
    "rc": 1
}`

![Image](https://github.com/user-attachments/assets/1113697a-1f49-45a3-923f-3c896d8e9698)

I'm looking to determine why there's a json conversion in a shell call.

### IBM z/OS Ansible core Version

v1.12.0

### IBM Z Open Automation Utilities

v1.3.0, 1.3.4.1

### IBM Enterprise Python

v3.13.x, 3.12.x

### ansible-version

v2.18.x - asking to try an olver copy

### z/OS version

v3.1

### Ansible module

zos_mount

### Playbook verbosity output.

ansible-playbook [core 2.18.1]
  config file = /Users/tobyk/ansible/ansible.cfg
  configured module search path = ['/Users/tobyk/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/11.1.0_1/libexec/lib/python3.13/site-packages/ansible
  ansible collection location = /Users/tobyk/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible-playbook
[...]

PR #2060, opened 2025-04-21
[Bugfix][zos_mount] added condition to skip fsumf168 during df call

richp405:
##### SUMMARY
added code in zos_mount to ignore an error return, if the return included FSUMF168 in the text of the df call.

Fiixes #1940 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zos_mount.py

##### ADDITIONAL INFORMATION
Speaking with storage experts, it was determined this message was a warning, and that the data was still being retunred.


richp405:
I switched this back to draft while working out a pipeline error.

richp405:
The pipeline error was my fault.  EC was mis-configured.

image

changelogs/fragments/2060-zos_mount-skip_fsumf168_from_df.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
```
rexemin, line 1: I feel like this one should be labeled as a bugfix

rexemin (changes requested):
Pretty straightforward. Looks good, just wondering if the changelog fragment should be a bugfix so that users are notified of this change

changelogs/fragments/2060-zos_mount-skip_fsumf168_from_df.yml
richp405: you are correct.

changelogs/fragments/2060-zos_mount-skip_fsumf168_from_df.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - zos_mount - Added handling to the df call, so that FSUMF168 are ignored.
```
fernandofloresg, line 2: We try to stick to the ansible development process for change logs https://docs.ansible.com/ansible/latest/community/development_process.html . In this case this is the format we use 
```
Fixes that resolve issues. SHOULD not be used for minor enhancements (use minor_change instead). Use past tense to describe the problem and present tense to describe the fix.
```

And since this bugfix will be client-facing it should be less technical more like "In some cases, df command would raise error FSUMF168 as a warning for certain cases(can add more details for those cases) and the module failed. Fix now handles this warning message gracefully and continues with the operation."

fernandofloresg: Your comments make perfect sense for these cases :
```
FSUMF168 return in stderror means that the mount dataset wouldn't resolve.
        # While this shows a catalog or volume issue, it should not impact our search for an existing mount
```

fernandofloresg (approved):
LGTM !

PR #2075, opened 2025-05-02
[Enhancement] [zos_stat] Add support to stat migrated data sets

### LINKED ISSUE #1186: [Enhancement] [zos_stat] Support to show migrated data sets

### Ansible module

zos_stat

Whenever the zos_stat module or which ever name is decided for the module, please add support for showing migrated data sets.

rexemin:
##### SUMMARY
Fixes #1186.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zos_stat

##### ADDITIONAL INFORMATION

Tested with a playbook that the feature works. No test so far as we don't have consistent access to DFSMShsm to migrate and recall data sets.

### DIFF
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -2149,7 +2149,7 @@ class DataSet(object):
             Return code, standard output and standard error from
             the HRECALL call.
         """
-        name = name.replace('$', '\$')
+        name = name.replace('$', '\\$')
         recall_cmd = f"""tsocmd "HRECALL '{name}'" """
         rc, stdout, stderr = module.run_command(recall_cmd)
 
--- a/plugins/module_utils/data_set.py
+++ b/plugins/module_utils/data_set.py
@@ -16,7 +16,7 @@ __metaclass__ = type
 import re
 import tempfile
 import traceback
-from os import path, walk
+from os import path, walk, environ
 from random import sample
 from string import ascii_uppercase, digits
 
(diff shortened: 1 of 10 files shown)


plugins/module_utils/data_set.py
```diff
@@ -2102,6 +2102,59 @@ def _build_volume_string_iehprogm(volumes):
...
+        -------
+        bool
+            Whether the data set has been migrated.
+        """
+        has_been_migrated = False
+
+        non_migrated_list = datasets.list_dataset_names(name, migrated=False)
+        migrated_list = datasets.list_dataset_names(name, migrated=True)
+
+        if name in migrated_list and name not in non_migrated_list:
```
richp405, line 2125: This comparison will not work if 'name' contains wild cards.  Check if you can see if wildcards could possibly get here, and consider fnmatch functions If needed.  https://docs.python.org/3/library/fnmatch.html

richp405 (changes requested):
Please check to make sure that either wild cards are blocked, or that something like fn match is used, instead of 'in'.

plugins/module_utils/data_set.py
fernandofloresg: At this point variable `name` is not a wildcard, inside the data set handler there is only one data set name.

fernandofloresg (approved):
LGTM!

surendrababuravella (approved):
Looks fine

plugins/module_utils/data_set.py
rexemin: The module only accepts querying a single resource. If there's a `*` or `?` present in `name`, the arg parser already rejects it. Adding wording to the documentation to emphasize this.

richp405 (approved):

PR #2136, opened 2025-06-16
Adds option to skip encoding for zos_unarchive module

Rohitcodes28:
##### SUMMARY

In the zos_unarhcive module  like the exclude option is there to exclude any dataset or file from the unarchive operation similarly skip encoding is added to the module so as user can skip  the encoding of certain files post unarchiving these files. 

##### ISSUE TYPE

- Enhancement Pull Request

##### COMPONENT NAME

zos_unarchive
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/2136-zos_unarchive_skip_encoding_support.yml
+++ b/changelogs/fragments/2136-zos_unarchive_skip_encoding_support.yml
@@ -1,3 +1,4 @@
 trivial:
-   - zos_unarchive - Adds the skip_encoding parameter to the unarchive module.
-     (https://github.com/ansible-collections/ibm_zos_core/pull/2136)
+  - zos_unarchive - Adds support for skipping encoding in zos_unarchive module.
+    This allows users to skip encoding for certain files after unarchiving them.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/2136)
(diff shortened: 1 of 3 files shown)


richp405 (approved):

Rohitcodes28:
image
image
Pipeline results

changelogs/fragments/2136-zos_unarchive_skip_encoding_support.yml
```diff
@@ -0,0 +1,3 @@
+trivial:
+   - zos_unarchive - Adds the skip_encoding parameter to the unarchive module.
```
fernandofloresg, line 2: Lets keep it similar to the archive one: "Adds support for skipping encoding in zos_unarchive module.
    This allows users to skip encoding for certain files after unarchiving them."

plugins/modules/zos_unarchive.py
```diff
@@ -342,6 +342,12 @@
           - The destination I(dest) character set for the files to be written as.
         required: false
         type: str
+      skip_encoding:
+        description:
+          - List of names to skip encoding after unarchiving. This is only used if I(Encoding) is set, otherwise is ignored.
```
fernandofloresg, line 347:
```suggestion
          - List of names to skip encoding after unarchiving. This is only used if I(encoding) is set, otherwise is ignored.
```

fernandofloresg (approved):
Looks good to me, requested a minor change which is not a showstopper since this is a trivial change

PR #2261, opened 2025-08-29
[Enabler][2146]update_zos_find_interface

### LINKED ISSUE #2146: [Enhancement] Update zos_find module interface

### Enabler description

As part of the standardization of module interfaces efforts we need to update `zos_find` module. Some modules have options name changes as well as behavior changes, in this case this module we will have behavior changes. Now, `pds_patterns` option will be removed and this will be inferred from the `patterns` value, like doing `OMVSADM.TEST.**(A*)` 

- [ ] Remove option names to the interface.
- [ ] Update module documentation and examples.
- [ ] Write or update tests to reflect the new interface.
- [ ] Update migration guide stating the changes to be performed by users when migrating to the new 2.0 version.
- [ ] Add a deprecation warning in updated options with aliases.

The current module interface is
```yaml
age
age_stamp
contains
excludes
patterns
size
pds_patterns
resource_type
volume
empty
extended
fifo
limit
purge
scratch
```

And the new proposed interface should be
```yaml
age
age_stamp
contains
excludes
patterns
size
resource_type
volume
empty
extended
fifo
limit
purge
scratch
```

How to add a deprecation warning:

```python
marker=dict(
                        type="list",
                        elements="str",
                        required=False,
                        aliases=["comment"],
                        deprecated_aliases=[
                                dict(
                                    name='comment',
                                    version='2.2.0',  # Version when it will be removed
                                    collection_name='ibm.ibm_zos_core',
                                )
                        ],
                    ),
```
In summary, requested changes are:
Options 
- Removes `pds_patterns`

### Ansible module

_No response_

AndreMarcel99:
##### SUMMARY
Fixes #2146

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME
Remove pds_patterns and for exclude members replace function.

### DIFF
--- a/plugins/modules/zos_find.py
+++ b/plugins/modules/zos_find.py
@@ -78,7 +78,6 @@ options:
         names that match at least one of the patterns specified. Multiple patterns
         can be specified using a list.
       - This parameter expects a list, which can be either comma separated or YAML.
-      - If C(pds_patterns) is provided, C(patterns) must be member patterns.
       - When searching for members within a PDS/PDSE, pattern can be a regular expression.
     type: list
     elements: str
@@ -92,17 +91,6 @@ options:
       - Filtering by size is currently only valid for sequential and partitioned data sets.
     required: false
     type: str
-  pds_patterns:
-    description:
-      - List of PDS/PDSE to search. Wildcard is possible.
-      - Required when searching for data set members.
-      - Valid only for C(nonvsam) resource types. Otherwise ignored.
-    aliases:
-      - pds_paths
-      - pds_pattern
-    type: list
-    elements: str
-    required: false
   resource_type:
     description:
       - The types of resources to search.
@@ -237,14 +225,6 @@ EXAMPLES = r"""
     patterns: 'IMS.LIB.*'
     contains: 'hello'
     excludes: '.*TEST'
-
-- name: Find all members starting with characters 'TE' in a given list of PDS patterns
-  zos_find:
-    patterns: '^te.*'
-    pds_patterns:
-      - IMSTEST.TEST.*
-      - IMSTEST.USER.*
-      - USER.*.LIB
 
 - name: Find all data sets greater than 2MB and allocated in one of the specified volumes
   zos_find:
@@ -418,7 +398,7 @@ def content_filter(module, patterns, content):
     return filtered_data_sets
 
 
-def data_set_filter(module, pds_paths, patterns):
+def data_set_filter(module, patterns):
     """ Find data sets that match any pattern in a list of patterns.
 
     Parameters
@@ -442,7 +422,6 @@ def data_set_filter(module, pds_paths, patterns):
         Non-zero return code received while executing ZOAU shell command 'dls'.
     """
     filtered_data_sets = dict(ps=set(), pds=dict(), searched=0)
-    patterns = pds_paths or patterns
     for pattern in patterns:
         rc, out, err = _dls_wrapper(pattern, list_details=True)
         if rc != 0:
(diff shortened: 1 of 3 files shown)


plugins/modules/zos_find.py
```diff
@@ -238,14 +227,6 @@
     contains: 'hello'
     excludes: '.*TEST'
 
-- name: Find all members starting with characters 'TE' in a given list of PDS patterns
```
fernandofloresg, line 241: Since the functionality is not being taken away but rather integrating it to an already existing option we should keep examples that show how to retrieve the members.

plugins/modules/zos_find.py
```diff
@@ -66,6 +66,7 @@
         Multiple patterns can be specified using a list.
       - The pattern can be a regular expression.
       - If the pattern is a regular expression, it must match the full data set name.
+      - Members are affected to be exclude if match on of the patterns.
```
fernandofloresg, line 69: This phrase is grammatically incorrect, the corrected version should be "Members will be excluded if they match one of the patterns.".
But in this scenario I think it would also be useful, e.g. showing an example on how to exclude members, will that be like this? `e.g. excludes: "USER.TEST.PDS(C*)"`?

Or will using `C*` will also exclude members like `USER.TEST.PDS(C*)`?

fernandofloresg: Judging by this line:

```
                            if excludes:
                                members = filter_members(module, members, excludes)
```

It seems the second option is being used, but I believe it's incorrect. Instead, we should intelligently derive the member regex from the excludes list, ensuring it accurately targets the intended members. As an example, if user wants to exclude certain members they could use `excludes: "USER.TEST.P*(C*)"` so that `USER.TEST.P*` will exclude data sets and `(C*)` shall exclude members.

fernandofloresg (changes requested):
Requested some more examples and a change to the exclusion logic. Overall looks good and thank you for adding the keys in the test cases.

plugins/modules/zos_find.py
AndreMarcel99: You're right I change the documentation to _To exclude members, the regular expression or pattern must be enclosed in parentheses._ and make the changes to how you mention before, I really like these design

AndreMarcel99:
Screenshot 2025-09-01 at 3 10 06 p m
Screenshot 2025-09-01 at 3 10 15 p m

plugins/modules/zos_find.py
```diff
@@ -218,6 +207,14 @@
 
 
 EXAMPLES = r"""
+- name: Exclude all members starting with characters 'TE' in a given list datasets patterns
+  zos_find:
+    excludes: '(^te.*)'
```
fernandofloresg, line 212: Could you also add an example on how to exclude members but also having the data set name? I'm afraid only having this example would scare users into thinking that you can only exclude data set names or member names at a time

plugins/modules/zos_find.py
```diff
@@ -66,6 +66,7 @@
         Multiple patterns can be specified using a list.
       - The pattern can be a regular expression.
       - If the pattern is a regular expression, it must match the full data set name.
+      - To exclude members, the regular expression or pattern must be enclosed in parentheses.
```
fernandofloresg, line 69:
```suggestion
      - To exclude members, the regular expression or pattern must be enclosed in parentheses. This expression can be used alongside a pattern to exclude data set names.
```

fernandofloresg (changes requested):
Requested minor documentation changes, no need to run the pipeline again.

plugins/modules/zos_find.py
fernandofloresg: Thank you!

Issue #2155, opened 2025-06-18
[Enabler][SPS] Generate compliance reports on every individual step


fernandofloresg:
### Is there an existing issue for this?

- [x] There are no existing issues.

### Enabler description

One of the main goals of moving our pipelines to SPS is to improve the compliance recordability and report, there is some reports that need to be supplied in order to keep our pipelines compliant and ensure the testing process isn't influenced by external factor.

This task should include
- [ ] Research the necessary reports to be gathered
- [ ] Define a place to store them
- [ ] Implement the reporting process in the pipeline

SME @AndreMarcel99

fernandofloresg:
All scans have been added to the SPS pipeline. Thank you @AndreMarcel99 !!!

PR #2263, opened 2025-09-01
Add_branch_protection_ruleset_and_detect_secrets_baseline

AndreMarcel99:
##### SUMMARY
Add files to ensure running pipeline without errors.
Is part of the task #2155 

**.secrets.baseline**
File base to recognize any secrets can be expose on the repository, for our case the option on mis_raw we use auth=true as an option that was taken as false positive, that's why is include the document in exclusion with the .secret.baseline file to avoid the false positive.


fernandofloresg (approved):
Thanks this looks good to me

PR #2258, opened 2025-08-26
[zos_lineinfile] Avoid leaving behind a temporary file when running test suite

fernandofloresg:
##### SUMMARY

zos_lineinfile tests were leaving behind files like 

```shell
# ls
ANSIBLE.PD88LV4G.T2090681.CVG6JWO8       opt
ANSIBLE.PEF8F16K.T2097663.COHLU8Y4(MEM)  proc
ANSIBLE.PU2T3IZY.T2105964.CLPM6M9A(MEM)  samples
SYSTEM                                   temp
allpython                                temp_file
bin                                      tmp
dev                                      u
dsfs                                     usr
etc                                      var
global                                   zoau
```

turns out that line 
```
cmd=f"iconv -f {encoding} -t IBM-1047 \"{ds_full_name}\" > \"{ds_full_name}\" "
```

Now the issue comes when in the tests we are using dls ANSIBLE.* , what happens is that the shell will expand ansible.* to the file name, making dls fail because the new executed command will be dls ANSIBLE.PD88LV4G.T2090681.CVG6JWO8 that throws an error message BGYSC1103E No datasets match pattern: ANSIBLE.PD88LV4G.T2090681.CVG6JWO8.

This fixes that issue and stabilizes some other unrelated test cases.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_lineinfile

PR #2265, opened 2025-09-01
[Enabler] Stabilize zos_archive tests

fernandofloresg:
##### SUMMARY

Addresses part of the issue of #2258, in that PR I removed a zos_lineinfile code that left some files behind, now after finding some other tests in zos_encode that left files behind I'm making the following changes:
1.- Modify the `zos_encode` test cases so that no temporary files are being left behind.
2.- Modify the way `dls` is executed in `zos_archive` tests, so that if in the future there is a temporary file like `ANSIBLE.TEST.ABC` as a uss file using `dls ANSIBLE.*` won;t get resolved to `dls ANSIBLE.TEST.ABC` but now we are using single quotes around the wildcard (`dls 'ANSIBLE.*'`) so shell doesn't resolve `ANSIBLE.*` making the tests fail.

##### ISSUE TYPE

- Enabler Pull Request

##### COMPONENT NAME

zos_encode
zos_archive

### DIFF
--- a/changelogs/fragments/2265-stabilize-zos_archive-tests.yml
+++ b/changelogs/fragments/2265-stabilize-zos_archive-tests.yml
@@ -1,0 +1,6 @@
+trivial:
+  - test_zos_archive_func.py - Modify the way dls is executed in zos_archive tests
+    adding a single quote wrap around wildcard searches.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/2265).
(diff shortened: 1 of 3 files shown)


fernandofloresg:
Screenshot 2025-09-01 at 3 43 27 p m
Screenshot 2025-09-01 at 3 43 46 p m
Screenshot 2025-09-01 at 3 44 18 p m

PR #2372, opened 2025-10-14
[zos_replace][1.15.1] Write uss files using encoding provided and only write if an occurrence was found

### LINKED ISSUE #2369: [Bug][1.15.1] zos_replace always converts unix file to ISO8859-1

### Bug description

It seems that zos_replace module always converts unix file to ISO8859-1 even when there is no match for given regexp.
IMO zos_replace should keep original encoding (like IBM-1047) of unix file and in case of no match the file should not be modified at all.
I've tested different PYTHONSTDINENCODING settings but the result was always the same - file converted to ISO8859-1.

v1.15.0

v1.3.6

### IBM Enterprise Python

v3.13.x

### ansible-version

v2.17.x

### z/OS version

v2.5

### Ansible module

zos_replace

_No response_

### Ansible configuration.

```YAML

```

### Contents of the inventory

```YAML

```

```YAML

```

fernandofloresg:
##### SUMMARY

Used the encoding value to write into a USS file and changed the logic of both USS and data set to only write if any occurrences were found.

Fix #2369 

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/functional/modules/test_zos_replace_func.py
+++ b/tests/functional/modules/test_zos_replace_func.py
@@ -1974,9 +1974,6 @@ def test_gdg_ds(ansible_zos_module):
         results = hosts.all.shell(cmd="cat \"//'{0}'\" ".format(params["target"]))
         for result in results.contacted.values():
             assert result.get("stdout") == TEST_AFTER
-        results = hosts.all.shell(cmd=f"md5 {params['target']}")
-        for result in results.contacted.values():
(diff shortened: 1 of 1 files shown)


fernandofloresg:
Had to test this in Jenkins because changing the base branch in SPS is not working.

Screenshot 2025-10-15 at 1 22 20 p m
Screenshot 2025-10-15 at 1 22 49 p m

richp405 (approved):

PR #2377, opened 2025-10-21
[New module] zos_started_task

surendrababuravella:
##### SUMMARY
New module zos_started_task which support start, display, modify, cancel, stop and force operations.

##### ISSUE TYPE
New Module Pull Request

##### COMPONENT NAME
zos_started_task

##### ADDITIONAL INFORMATION

### DIFF
--- a/tests/functional/modules/test_zos_started_task_func.py
+++ b/tests/functional/modules/test_zos_started_task_func.py
@@ -16,8 +16,8 @@
 from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
-from ibm_zos_core_startedtask.tests.helpers.dataset import get_tmp_ds_name
-from ibm_zos_core_startedtask.tests.helpers.utils import get_random_file_name
+from ibm_zos_core.tests.helpers.dataset import get_tmp_ds_name
+from ibm_zos_core.tests.helpers.utils import get_random_file_name
 from ibm_zos_core.tests.helpers.dataset import get_tmp_member
 from shellescape import quote
 import re


plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    type: bool
+  identifier_name:
+    description:
+        - Option I(identifier_name) is the name that identifies the task. This name can be up to 8
+          characters long. The first character must be alphabetical.
+    required: false
+    type: str
+    aliases:
+        - identifier
+  get_system_logs:
```
fernandofloresg: Let's change this option's name to `system_logs`, we are trying to avoid the verb options for the new 2.0 version so is better to have the same guidelines applied here.

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    description:
+        - Whether to perform a dump. The type of dump (SYSABEND, SYSUDUMP, or SYSMDUMP)
+          depends on the JCL for the job.
+        - Only applicable when I(state) is C(cancelled), otherwise ignored.
+    required: false
+    type: bool
+  identifier_name:
+    description:
+        - Option I(identifier_name) is the name that identifies the task. This name can be up to 8
+          characters long. The first character must be alphabetical.
```
fernandofloresg: Can we add in which states this is applicable?

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+        - Any appropriate keyword parameter that you specify to override the corresponding parameter in the cataloged
+          procedure. The maximum length of each keyword=option pair is 66 characters. No individual value within this
+          field can be longer than 44 characters in length.
+        - Only applicable when I(state) is C(started), otherwise ignored.
+    required: false
+    type: dict
+  member_name:
+    description:
+        - Name of a member of a partitioned data set that contains the source JCL for the task to be started. The member
+          can be either a job or a cataloged procedure.
```
fernandofloresg: what happens if a user provides both `job_name` and `member_name` at the same time?

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+        - Only applicable when I(state) is C(started), otherwise ignored.
+    required: false
+    type: str
+    aliases:
+        - member
+  parameters:
+    description:
+        - Program parameters passed to the started program.
+        - Only applicable when I(state) is C(started) or C(modified), otherwise ignored.
```
fernandofloresg: Can we add a sample of some parameters of a test task?

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    type: bool
+  state:
+    description:
+        - I(state) should be the desired state of the started task after the module is executed.
+        - If I(state) is C(started) and the respective member is not present on the managed node, then error will be thrown with C(rc=1),
+          C(changed=false) and I(stderr) which contains error details.
+        - If I(state) is C(cancelled), C(modified), C(displayed), C(stopped) or C(forced) and the started task is not running on the managed node,
+          then error will be thrown with C(rc=1), C(changed=false) and I(stderr) contains error details.
+        - If I(state) is C(displayed) and the started task is running, then the module will return the started task details along with
+          C(changed=true).
```
fernandofloresg: I think in this case changed should be False because other than issuing the command the module is not changing the managed node state.
```suggestion
          C(changed=false).
```

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+#     required: false
+#     type: str
+#   volume:
+#     description:
+#         - If I(device_type) is a tape or direct-access device, the serial number of the volume,
+#           mounted on the device.
+#         - Only applicable when I(state) is C(started), otherwise ignored.
+#     required: false
+#     type: str
+  userid:
```
fernandofloresg: Lets make it snake case, that way is more consistent with other option names.
```suggestion
  user_id:
```

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+        - When C(verbose=true), the module will return the started task execution logs.
+    required: false
+    type: bool
+    default: false
+  wait_full_time:
+    description:
+        - If C(wait_full_time=false), the module polls every 5 seconds to check the status of the started task and returns
+          immediately once the task is successfully validated.
+        - When C(wait_full_time=true), the module waits for the duration specified in I(wait_time), even after the started
+          task operation has been successfully validated.
```
fernandofloresg: Please also mention what would happen if I wait full time with different state values, like start, cancel, force.

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    member: "PROCAPP"
+    job_name: "SAMPLE"
+    verbose: True
+- name: Start a started task specifying the subsystem and enabling a reusable ASID.
+  zos_started_task:
+    state: "started"
+    member: "PROCAPP"
+    subsystem: "MSTR"
+    reus_asid: "YES"
+- name: Display a started task using a started task name.
```
fernandofloresg: Please add a started task with wait time so it shows how to wait for it to get all attributes available.

fernandofloresg (changes requested):
Request few minor changes

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+          characters long. The first character must be alphabetical.
+    required: false
+    type: str
+    aliases:
+        - identifier
+  get_system_logs:
+    description:
+        - When C(get_system_logs=true), the module will return system logs that describe the task's execution.
+          This option can return a big response depending on system load, also it could surface other
+          program's activity.
```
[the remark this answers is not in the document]
surendrababuravella: Added

plugins/modules/zos_started_task.py
surendrababuravella: This is applicable for all states

plugins/modules/zos_started_task.py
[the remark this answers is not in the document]
surendrababuravella: It is applicable for all states

plugins/modules/zos_started_task.py
surendrababuravella: Member_name is needed for state=started only.. When both are present then job name will be the task name.

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    type: list
+    elements: str
+#   retry_force:
+#     description:
+#         - Indicates whether retry will be attempted on ABTERM(abnormal termination).
+#         - I(tcb_address) is mandatory to use I(retry_force).
+#         - Only applicable when I(state) is C(forced), otherwise ignored.
+#     required: false
+#     type: bool
+  reus_asid:
```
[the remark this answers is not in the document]
surendrababuravella: updated

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+    description:
+        - When I(reus_asid) is C(True) and REUSASID(YES) is specified in the DIAGxx parmlib member, a reusable ASID is assigned
+          to the address space created by the START command. If I(reus_asid) is not specified or REUSASID(NO) is specified in
+          DIAGxx, an ordinary ASID is assigned.
+        - Only applicable when I(state) is C(started), otherwise ignored.
+    required: false
+    type: bool
+  state:
+    description:
+        - I(state) should be the desired state of the started task after the module is executed.
```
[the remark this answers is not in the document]
surendrababuravella: done

plugins/modules/zos_started_task.py
surendrababuravella: done

plugins/modules/zos_started_task.py
surendrababuravella: done

plugins/modules/zos_started_task.py
```diff
@@ -0,0 +1,1639 @@
...
+        rc=rc,
+        stdout=stdout,
+        stderr=stderr,
+        stdout_lines=stdout.split('\n'),
+        stderr_lines=stderr.split('\n'),
+        system_logs=system_logs,
+        verbose_output=task_output_logs
+    )
+    if msg:
+        result['msg'] = msg
```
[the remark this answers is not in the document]
surendrababuravella: done

fernandofloresg (approved):
Thanks for addressing all comments, making changes after feedback sessions where we went back and forth multiple times, even reverting previously agreed behaviors. 

Appreciate your effort and attention to detail during the whole development process.

PR #2386, opened 2025-10-28
[Role][1625]simple_role_to_get_job_status

### LINKED ISSUE #1625: [Role] A role that can respond with a simple reply if a Job is running or not.

### New module description

A role which can perform a check if a job is running or not, this role would be based on the zos_job_query module.

The idea behind the role is to reduce the parsing effort needed to obtain a simple response on if a job is running. Many times automation needs just a simple reply and in this case here a module is returning a full contract/response that needs excessive parsing. 

So original request which has since been cancelled and migrated to this issue.
https://github.com/ansible-collections/ibm_zos_core/issues/1451

Note: We may returning more than true/false here, it just depends on our research but jobs have states, it might be better return a state.

### User story

As an Ansible user, I can check easily if my job (given its ID) is running without having to look into the `zos_job_query` return values and work out what it means.

AndreMarcel99:
##### SUMMARY
Create a role that execute zos_job_query to parsing the response to return job_status and if is running or not as boolean value. Fixes #1625 

##### ISSUE TYPE

- Role Pull Request
- Bugfix Pull Request

##### COMPONENT NAME
Role that get first value of zos_job_query expecting to be the job_id set by the user. 
Delete unnecessary parsing on zos_job_query that remove the current estatus of the job.
Also adding test method for roles on pytest.

Screenshot 2025-10-27 at 6 55 51 p m

### DIFF
--- a/plugins/modules/zos_job_query.py
+++ b/plugins/modules/zos_job_query.py
@@ -465,7 +465,7 @@ def parsing_jobs(jobs_raw):
     for job in jobs_raw:
         ret_code = job.get("ret_code")
         # Easier to see than checking for an empty string, JOB NOT FOUND was
-        # replaced with None in the jobs.py and msg_txt field describes the job query instead        
+        # replaced with None in the jobs.py and msg_txt field describes the job query instead
         if job.get("ret_code") is None:
             status_raw = "JOB NOT FOUND"
             ret_code["msg"] = status_raw


richp405 (approved):
Nice simple role.  One minor typo in the readme.  A question: should you mention the role in the changelog, or not, since it is a role, not main code?

richp405 (dismissed):

plugins/modules/zos_job_query.py
```diff
@@ -468,38 +468,9 @@ def parsing_jobs(jobs_raw):
...
+            ret_code["msg"] = status_raw
         elif job.get("ret_code").get("msg", "JOB NOT FOUND") is None:
             status_raw = "JOB NOT FOUND"
-        else:
-            status_raw = job.get("ret_code").get("msg", "JOB NOT FOUND")
-
-        if "AC" in status_raw:
-            # the job is active
-            ret_code["msg"] = None
-            ret_code["msg_code"] = None
```
fernandofloresg: if this is not set, will it still be returned as None or not returned at all ?

roles/job_status/README.md
```diff
@@ -0,0 +1,9 @@
+# ibm.ibm_zos_core.job_status
+
+This role gathers job status and if the jos is active about a giving a job_id.
```
fernandofloresg: Would suggest a different phrasing 

```suggestion
This role gathers job status information and checks if a job is active, given a specific job ID.
```

plugins/modules/zos_job_query.py
AndreMarcel99: It returns as None, but right better to be safe

PR #2437, opened 2026-03-17
[Bug][zos_backup_restore] Enable testcase test_backup_and_restore_of_data_set_from_volume_to_new_volume

### LINKED ISSUE #2348: [Enabler][zos_backup_restore] Uncomment test case test_backup_and_restore_of_data_set_from_volume_to_new_volume

### Enabler description

In theory what this test does is perform a backup from one volume to another, in practice this keeps failing in the ECs even though it should work. This work item is meant to investigate how to get this working.

### Ansible module

_No response_

### LINKED ISSUE #2436: [Documentation][zos_backup_restore] Explain usage of temp_volume module option to prevent temp data sets being allocated on source volume

Currently, when backing up/ restoring a data set from a particular volume to another particular volume, if the underlying ZOAU `dzip` mechanism allocates temporary data sets on the same volume as the output volume, the operation fails.
This behavior was revealed by a [test case](https://github.com/ansible-collections/ibm_zos_core/blob/680ac8fab64bac14ca4e7b6c251123543d14bbb1/tests/functional/modules/test_zos_backup_restore_func.py#L934) intermittently failing (thanks to @surendrababuravella for  investigating). 
* See #2348 

This behavior needs to be documented in the module docs. The current work-around is to leverage the already existing `temp_volume` module option and specify a different volume than the target/dest volume. This behavior should be documented under the module option docs as well as under `known issues`

v1.16.0-beta.1

### Ansible module

zos_backup_restore

surendrababuravella:
##### SUMMARY
Fixes #2348 and fixes #2436

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
zos_backup_restore

##### ADDITIONAL INFORMATION

- Enabled test case test_backup_and_restore_of_data_set_from_volume_to_new_volume with `temp_volume` parameter. 
- Updated documentation to clarify `temp_volume` usage and document known issue where operations fail if `temp_volume` matches source volume or is unspecified.

### DIFF
--- a/changelogs/fragments/2348-zos_backup_restore_uncomment_testcase.yml
+++ b/changelogs/fragments/2348-zos_backup_restore_uncomment_testcase.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - zos_backup_restore - Enabled test case test_backup_and_restore_of_data_set_from_volume_to_new_volume
+    with `temp_volume` parameter. Updated documentation to clarify `temp_volume` usage and document
+    known issue where operations fail if `temp_volume` matches source volume or is unspecified.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/2437)
--- a/changelogs/fragments/2348-zos_backup_restore_uncomment_testcase.yml
+++ b/changelogs/fragments/2348-zos_backup_restore_uncomment_testcase.yml
@@ -1,5 +1,0 @@
-bugfixes:
-  - zos_backup_restore - Enabled test case test_backup_and_restore_of_data_set_from_volume_to_new_volume
-    with `temp_volume` parameter. Updated documentation to clarify `temp_volume` usage and document
-    known issue where operations fail if `temp_volume` matches source volume or is unspecified.
-    (https://github.com/ansible-collections/ibm_zos_core/pull/2437)
--- a/changelogs/fragments/2437-zos_backup_restore_uncomment_testcase.yml
+++ b/changelogs/fragments/2437-zos_backup_restore_uncomment_testcase.yml
@@ -1,0 +1,9 @@
+bugfixes:
+  - zos_backup_restore - Updated documentation to clarify `temp_volume` usage and document
+    known issue where operations fail if `temp_volume` matches source volume or is unspecified.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/2437)
+
+trivial:
+  - zos_backup_restore - Enabled test case test_backup_and_restore_of_data_set_from_volume_to_new_volume
+    with `temp_volume` parameter.
+    (https://github.com/ansible-collections/ibm_zos_core/pull/2437)
--- a/plugins/modules/zos_backup_restore.py
+++ b/plugins/modules/zos_backup_restore.py
@@ -156,6 +156,12 @@ options:
         created during the backup and restore process.
       - When I(operation=backup) and I(backup_name) is a data set, specifies the
         volume the backup should be placed in.
+      - When backing up a data set from one volume to another, the operation
+        will fail if I(temp_volume) is set to the same volume as the source volume. Additionally,
+        if I(temp_volume) is not specified, the underlying ZOAU mechanism may allocate temporary
+        data sets on the same volume as the source volume, which can also cause the operation to
+        fail. As a workaround, always use the I(temp_volume) option to specify a different volume
+        than the source volume for temporary data sets.
     type: str
     required: False
     aliases:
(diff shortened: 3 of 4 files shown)


surendrababuravella:
SPS pipeline run:
Screenshot 2026-03-17 at 6 23 15 PM
Screenshot 2026-03-17 at 6 23 15 PM

changelogs/fragments/2348-zos_backup_restore_uncomment_testcase.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - zos_backup_restore - Enabled test case test_backup_and_restore_of_data_set_from_volume_to_new_volume
```
ketankelkar: changelog fragments get turned into changelogs and eventually flow into release notes. reason i'm mentioning this is content in these fragments should be relevant to the user. Test cases, while beneficial to development, do not affect the user experience and hence i'd say should be classified under `trivial`. 

I would split this up into two entries - trivial for test case update and bugfix (as you have it here) for update to docs only bc this particular update addresses a potential bug

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
         created during the backup and restore process.
       - When I(operation=backup) and I(backup_name) is a data set, specifies the
```
ketankelkar, line 157: is it just me or is this line misleading/incorrect for the `temp_volume` option?

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
...
         volume the backup should be placed in.
+      - When backing up a data set from one volume to another, the operation
+        will fail if I(temp_volume) is set to the same volume as the source volume. Additionally,
```
ketankelkar: I think we can simplify this section to describe only the module behavior if the option is not set and how to avoid those 'bad' outcomes by using the temp_volume module option.


```suggestion
        will fail if temporary data sets are created on the same volume as the volume containing the source data set(s). To avoid this, set I(temp_volume) to a different volume than the source volume.
```

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
...
+        if I(temp_volume) is not specified, the underlying ZOAU mechanism may allocate temporary
```
ketankelkar: probably don't need to mention ZOAU as long the module behavior is clear.


```suggestion

```

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
...
+        data sets on the same volume as the source volume, which can also cause the operation to
```
ketankelkar:
```suggestion
```

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
...
+        fail. As a workaround, always use the I(temp_volume) option to specify a different volume
```
ketankelkar: I wouldnt consider this a "work-around" - more like an expected behavior and how to deal with it. 
```suggestion

```

plugins/modules/zos_backup_restore.py
```diff
@@ -156,6 +156,12 @@
...
+        than the source volume for temporary data sets.
```
ketankelkar:
```suggestion

```

plugins/modules/zos_backup_restore.py
```diff
@@ -354,6 +360,12 @@
       the module option without access to the class.
     - If your system uses a different security product, consult that product's
       documentation to configure the required security classes.
+    - B(Known Issue:) When backing up a data set from one volume to another, the operation
```
ketankelkar: I would simplify this similar to in the module option doc:
> When backing up a data set from one volume to another, the operation will fail if temporary data sets are created on the same volume as the volume containing the source data set(s). To avoid this, use the I(temp_volume) option to specify a different volume than the source volume for temporary data sets.

Also probably no need to **bold** the known issue, just following the patterns of other modules (e.g. zos_copy notes section)

changelogs/fragments/2437-zos_backup_restore_uncomment_testcase.yml
ketankelkar, line 1: perhaps the wrong PR number in the changelog fragment file name?

from [docs](https://docs.ansible.com/projects/ansible/latest/community/development_process.html#creating-a-changelog-fragment):
> The file name should include the PR number and a description of the change.

ketankelkar (changes requested):
Good work here, thanks for including these docs changes we'd discussed. I've made a few tweaks to your docs, i think we can get away with not mentioning underlying ZOAU behaviors and instead focus on the user experience - i.e what the user can do to get the module to behave in certain ways.

Otherwise this PR looks good to me

plugins/modules/zos_backup_restore.py
surendrababuravella: This line is valid. When we try to take backup into a dataset, the backup dataset will be created in temp_volume.

ketankelkar (approved):
LGTM

yrana17 (approved):
LGTM

PR #2483, opened 2026-05-05
[Bug] [zos_job_query] Module is passing * for parameter defaults and not using latest ZOAU fetch_multiple interface

### LINKED ISSUE #2367: [Bug] [zos_job_query] Module is passing `*` for job_id when the task specifies None leading to all jobs being queried.

### Bug description

Review the internal discussion here: `<sanitized>/archives/CSLEJ8VGV/p1760138787897389?thread_ts=1754323020.810359&cid=CSLEJ8VGV`

In zos_job_query, there is [this logic](https://github.com/ansible-collections/ibm_zos_core/blob/4682e9e1792abbecba876de0c54e40aae99408bd/plugins/modules/zos_job_query.py#L440) my assumption is this might have been placed for when ZOAU did not handle wild cards but unsure. 

Now that it does handle wild cards the interpretation of None to `*` is detrimental in that it causes the module to query all system jobs , ones that the current user does not have access to view causing RACF violations. 

```
    jobs = []
    if job_id:
        jobs = job_status(job_id=job_id)
    elif owner:
        jobs = job_status(owner=owner, job_name=job_name)
    else:
        jobs = job_status(job_name=job_name)
    if not jobs:
        raise RuntimeError("List FAILED! no such job was found.")
    return jobs
```

While helping a user, I made a minor modification that completely removed the above block of code and from then on the user did not trigger the RACF violations , that change was to remove the above and replace it with (below) essentially overriding the existing logic and allowing for ZOAU talking to the SSI to handle the values passed from the module via ZOAU [fetch_multiple](https://www.ibm.com/docs/en/zoau/1.3.x?topic=apis-jobs#zoautil_pyjobsfetch_multipleemjob_idnone-job_ownernone-job_namenone-include_extendedfalse-kwargsem).
```
jobs = []
jobs = job_status(job_id=job_id, owner=owner, job_name=job_name, dd_name=False)
return jobs
```

To prove this was happening, the second code edit was to force an exception to show me what ZOAU was receiving before my above code change, for that I added to [../module_utils/job.py](https://github.com/ansible-collections/ibm_zos_core/blob/4682e9e1792abbecba876de0c54e40aae99408bd/plugins/module_utils/job.py#L349C1-L350C94) 

From:
```
    # expanding > 1.3.0 of zoau, to include all params
    entries = jobs.fetch_multiple(job_id=job_id_temp, job_owner=owner, include_extended=True)
```

To:
```
    mssg = f"fetch_multiple is passed job_id=[{job_id}], job_name=[{job_name}], owner=[{owner}]]."
    raise Exception(mssg)
```

Which yields for these module options below, noticed that None became --> `*` for `job_id`.

```
       "module_args": {
            "job_id": null,
            "job_name": "BG*",
            "owner": "IBMUSER"
        }
    },
    "message": "",
    "msg": "fetch_multiple is passed job_id=[*], job_name=[BG*], owner=[IBMUSER]]."
```

I believe the logic in the zos_job_query module needs to be reviewed and decided if it really is needed given ZOAU can handle all three options, maybe the right way is to pass all 3 options, job_id, job_name, owner all at once instead of this if/else logic in the zos_job_query module. 

v1.14.0

v1.3.3

### IBM Enterprise Python

v3.11.x

### ansible-version

v2.15.x

### z/OS version

v2.5

### LINKED ISSUE #2368: [Bug] [Bug] [zos_job_query] [module_utils/job.py] is not using the latest ZOAU interface for fetch_multiple

### Bug description
Review the internal discussion here: `<sanitized>/archives/CSLEJ8VGV/p1760138787897389?thread_ts=1754323020.810359&cid=CSLEJ8VGV`

This issue relates, or I should say was discovered as I was trouble shooting this related issue #2367. 

While I don't have a concrete bug related to this discovery at the moment, there may be one as I am still trying to understand why we are getting a CEE dump and this might be the reason, regardless, the API seems to be dated here.

ZOAU API [fetch_multiple](https://www.ibm.com/docs/en/zoau/1.3.x?topic=apis-jobs#zoautil_pyjobsfetch_multipleemjob_idnone-job_ownernone-job_namenone-include_extendedfalse-kwargsem) supports these options `job_id=None, job_owner=None, job_name=None` yet the ../module_utils/job.py [here](https://github.com/ansible-collections/ibm_zos_core/blob/4682e9e1792abbecba876de0c54e40aae99408bd/plugins/module_utils/job.py#L350C20-L350C34) and [here](https://github.com/ansible-collections/ibm_zos_core/blob/4682e9e1792abbecba876de0c54e40aae99408bd/plugins/module_utils/job.py#L356) are only using:

```
entries = jobs.fetch_multiple(job_id=job_id_temp, job_owner=owner, include_extended=True)
```

and

```
entries = jobs.fetch_multiple(job_id=job_id_temp, job_owner=owner, include_extended=True) 
```

By updating the API usage, its possible that the surrounding code could be improved/reduced. I do see this comment but wonder if it is noting that the module is not using job_name, 
```
    # In ZOAU>= 1.3.0, include_extended has to be set to true so we get the program name for a job.
    # Observation shows the job_name parameter is not being used, so we will drop that
```

As I do see the job_name later used, which might be able to see improvement in this code snippet
```
            if owner != "*":
                if owner != entry.owner:
                    continue
            if job_name != "*":
                if not fnmatch.fnmatch(entry.name, job_name):
                    continue
            if job_id_temp is not None:
                if not fnmatch.fnmatch(entry.job_id, job_id):
                    continue
```

What it comes down to is ZOAU fetch_multiple passes these options to JES and JES makes the decision on what to return, it seems like the module util at one time was trying to handle some of this logical processing and might need to be revisited. There is a chance this issue relates to something else I am looking into where a particular user is triggering a CEE dump when ZOAU is called and it might be because the Ansible code is not passing all the options it could, more on that as we dive deeper. 

v1.14.0

v1.3.3

### IBM Enterprise Python

v3.11.x

### ansible-version

v2.15.x

### z/OS version

v2.5

### Ansible module

zos_job_query

_No response_

### Ansible configuration.

```YAML

```

### Contents of the inventory

```YAML

```

```YAML

```

hmillersarah:
##### SUMMARY

Fixes #2367
Fixes #2368 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zos_job_query

##### ADDITIONAL INFORMATION

**#2367 Bug**: 
User-provided task resulting in RACF violations:
```yaml
     zos_job_query:
        job_id: null
        job_name: "HELLO"
        owner: "OMVSADM"
```
The `zos_job_query` module passes `*` as the default for `job_id`, `job_name`, and `owner` parameters if the parameter is null. But as of ZOAU v1.4.1.1, the ZOAU fetch_multiple command ignores job_id="*", so this violation does not create issues. 

However, in the alternate scenario below where job_id="*", `zos_job_query` incorrectly returns all jobs for all owners in the system instead of jobs that match the job_name and owner parameters specified. This issue lies in the `query_jobs` function, where if job_id is specified, it is the only parameter sent to the ZOAU fetch_multiple command even if the other parameters are specified.
```yaml
     zos_job_query:
        job_id: *
        job_name: "HELLO"
        owner: "OMVSADM"
```

**#2368 Bug**: 
User-provided task resulting in CEE dump generation
```yaml
     zos_job_query:
        job_name: "HELLO"
        owner: "OMVSADM"
```
The ZOAU `fetch_multiple` command is not passed the `job_name` parameter from `zos_job_query` and the job_name filtering is handled from the`zos_job_query` job module utility. This results in only the owner being passed to fetch_multiple to query jobs, which may cause access violations if a user does not have access to all the jobs of the specified owner.

**Fixes**: 
`job_name`
- If a null parameter is passed to `job_name`, the module updates the default to `*` in accordance with the module [documentation](https://ibm.github.io/z_ansible_collections_doc/ibm_zos_core/docs/source/modules/zos_job_query.html) stating the `job_name` default is `*`. 
- The `job_name` parameter is now passed to the ZOAU `fetch_multiple` command, as the command can now ingest job_name as a parameter for querying.

`job_id`
- If a null parameter is passed to `job_id`, the parameter defaults to `None`. 
- Removed the previous logic that checked if the`job_id` parameter was passed and if so, prioritized it over the other parameters by passing only the `job_id` to  the`job_status` call from `query_jobs` method in `zos_job_query` module. This fixes the alternate issue of #2368 bug.

`owner`
- If a null parameter is passed to `owner`, the parameter defaults to `None`. 
- Updated the `zos_job_query` documentation to note functionality where if an owner is not passed to the module, all owners will be queried.

**Validation**: 
The following tasks can be used to validate the functionality of zos_job_query:
```yaml
     zos_job_query:
        job_name: "HELLO"
        owner: "OMVSADM"

     zos_job_query:
        job_id: *
        job_name: "HELLO"
        owner: "OMVSADM"

     zos_job_query:
        job_id: null
        job_name: "HELLO"
        owner: "OMVSADM"
```


hmillersarah:
ZIP file contains screenshots of successful SPS pipeline runs for test files: 
[bugfix:2367.zip](https://github.com/user-attachments/files/27451703/bugfix.2367.zip)

plugins/module_utils/job.py
```diff
@@ -167,6 +167,10 @@ def _job_not_found(job_id, owner, job_name, dd_name):
         job_not_found_msg = "{0} with the job_id {1}".format(job_name.upper(), job_id.upper())
     elif job_id != '*':
         job_not_found_msg = "with the job_id {0}".format(job_id.upper())
+    elif job_id is None and job_name is None:
+        job_not_found_msg = "{0} with the job_id {1}".format(job_name, job_id)
```
ketankelkar: wouldn't this result in the message being: "None with the job_id None"?

plugins/modules/zos_job_query.py
```diff
@@ -46,8 +46,7 @@
   owner:
     description:
       - Identifies the owner of the job.
-      - If no owner is set, the default set is 'none' and all jobs will be
-        queried.
+      - If no owner is set, the default set is 'None' and the parameter will not be used for job querying.
     type: str
     required: False
```
ketankelkar, line 51:
```suggestion
    required: False
    default: None
```

plugins/modules/zos_job_query.py
```diff
@@ -59,6 +58,7 @@
         with `S`, `J`, `T` and are followed by 7 digits.
       - The I(job_id) can contain include multiple wildcards.
       - The asterisk (`*`) wildcard will match zero or more specified characters.
+      - If no job_id is set, the default set is 'None' and the parameter will not be used for job querying.
     type: str
     required: False
```
ketankelkar, line 64:
```suggestion
    required: False
    default: None
```

plugins/module_utils/job.py
ketankelkar:
```suggestion
# Copyright (c) IBM Corporation 2019, 2026
```

changelogs/fragments/2367-fix_zos_job_query.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - zos_job_query - All default parameters are set to ``*``, causing the module to query all jobs when a parameter is not set and leading
```
ketankelkar: I would reframe this so it's focused on user impact, not internal behavior/changes.

> zos_job_query – Users would see RACF violations when job_id or owner were not set and the user lacked authority to view all jobs, as job_id and owner defaulted to *. The fix now sets both defaults to None, preventing overly broad job queries.

changelogs/fragments/2367-fix_zos_job_query.yml
ketankelkar, line 1: the filename should reference the PR number (while the link in the changelog references the issue number if available, which you've done), also try for a more descriptive file name than "fixes-<module>" maybe something like "2483-fixes-greedy-job-query-default-behavior.yml"

tests/functional/modules/test_zos_job_query_func.py
ketankelkar:
```suggestion
# Copyright (c) IBM Corporation 2019, 2026
```

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -359,4 +374,83 @@ def test_zos_job_id_query_short_ids_with_wilcard_func(ansible_zos_module):
         assert rc.get("msg") is not None
         assert rc.get("msg_code") == "0000"
         assert rc.get("code") == 0
-        assert rc.get("msg_txt") == "CC"
+
+
+def test_zos_job_query_with_null_and_wildcard_job_id(ansible_zos_module):
```
ketankelkar: should there be another test for `owner=None`?

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -359,4 +374,83 @@ def test_zos_job_id_query_short_ids_with_wilcard_func(ansible_zos_module):
...
+        assert job.get("svc_class") is None
+        assert job.get("steps") is not None
+
+        rc = job.get("ret_code")
+        assert rc.get("msg") is not None
+        assert rc.get("msg_code") == "0000"
+        assert rc.get("code") == 0
+
+    # Assert length of results for both queries is the same
+    assert qresults_all_len == qresults_null_len
```
ketankelkar: is it supposed to be the same? Maybe I misunderstood something (very possible), but I thought the prior behavior was _all_ jobs were getting queried and the fix was supposed to only query the _relevant_ jobs by no longer matching on `'*'`. If that's the case, wouldn't `qresults_all < qresults_null`?

Perhaps another way to test the behavior would be to attempt to query a job that's not accessible (per RACF permissions) and then ensure the default behavior doesn't attempt to query it - this would take some implementation efforts no doubt

plugins/modules/zos_job_query.py
```diff
@@ -46,8 +46,7 @@
   owner:
     description:
       - Identifies the owner of the job.
-      - If no owner is set, the default set is 'none' and all jobs will be
-        queried.
+      - If no owner is set, the default set is 'None' and the parameter will not be used for job querying.
```
ketankelkar:
```suggestion
      - If no owner is set, the parameter will not be used for job querying.
```

I'm not sure it's valuable to mention in module docs the "how" behind the implementation

plugins/modules/zos_job_query.py
```diff
@@ -59,6 +58,7 @@
         with `S`, `J`, `T` and are followed by 7 digits.
       - The I(job_id) can contain include multiple wildcards.
       - The asterisk (`*`) wildcard will match zero or more specified characters.
+      - If no job_id is set, the default set is 'None' and the parameter will not be used for job querying.
```
ketankelkar:
```suggestion
      - If no job_id is set, the parameter will not be used for job querying.
```

plugins/modules/zos_job_query.py
ketankelkar: Good instincts on updating docs (:

tests/functional/modules/test_zos_job_query_func.py
hmillersarah: @ketankelkar I agree that adding the test case you mentioned would be a good way to validate that job_id is not querying all if no id is specified - I will work on adding a new test case for that scenario. 

With this test case, I was validating that the issue below was no longer occurring:

In the following example, job_name and owner are specified, while job_id is set to “\*” (originally, not including job_id or setting it to null would also result in `job_id="*"`):
```
  zos_job_query:
    job_id: "*"
    job_name: "INIT"
    owner: "USRT001"
```

Originally, the module would return all jobs with every `job_name` and `owner`, not just the ones passed in the parameters, which meant that `job_id` was taking precedence over all other parameters. With the updates to the module, if there is a `job_name` and `owner` specified then they will take precedence, and `job_id="*"` should virtually act like `job_id=None`.

plugins/module_utils/job.py
```diff
@@ -251,9 +255,10 @@ def job_status(job_id=None, owner=None, job_name=None, dd_name=None):
         {"job_id": job_id, "owner": owner, "job_name": job_name}
     )
 
-    job_id = parsed_args.get("job_id") or "*"
+    job_id = parsed_args.get("job_id") or None
```
surendrababuravella:
```suggestion
    job_id = parsed_args.get("job_id")
```
get handles it and returns None when key is not found

plugins/module_utils/job.py
```diff
@@ -251,9 +255,10 @@ def job_status(job_id=None, owner=None, job_name=None, dd_name=None):
...
+    # Defaults to wildcard
     job_name = parsed_args.get("job_name") or "*"
-    owner = parsed_args.get("owner") or "*"
+    owner = parsed_args.get("owner") or None
```
surendrababuravella:
```suggestion
    owner = parsed_args.get("owner")
```

tests/functional/modules/test_zos_job_query_func.py
hmillersarah: I have added the test case `test_zos_job_query_return_only_authorized_jobs` that verifies only jobs that a user is authorized to view are output when defaults are specified, and I found issues with the current implementation of `owner` that was likely also causing RACF violation issues after I started testing with a user with limited job view permissions. I have now altered the default for owner so that if nothing is passed/None is specified, then only the remote host user will be used for the query instead of defaulting to all users’ jobs being queried. - I've updated the PR description with the changes to the `owner` parameter as well.

hmillersarah:
ZIP file contains screenshots of successful SPS pipeline runs for test files following updates for bugfix #2368: 
[bugfix:2368.zip](https://github.com/user-attachments/files/29139419/bugfix.2368.zip)

plugins/module_utils/job.py
```diff
@@ -335,7 +340,7 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, sysin=Fal
         If no job status is found it will return a ret_code diction with
         parameter 'msg_txt" = "The job could not be found.
     """
-    if job_id == "*":
+    if job_id == "*" or job_id is None:
```
surendrababuravella: This code change doesn't add anything extra. Previously if job_id is None then job_id_temp is None only.

plugins/module_utils/job.py
```diff
@@ -163,7 +163,11 @@ def _job_not_found(job_id, owner, job_name, dd_name):
     """
     # Note that the text in the msg_txt is used in test cases and thus sensitive to change
     jobs = []
-    if job_id != '*' and job_name != '*':
+    if job_id is None and job_name is None:
+        job_not_found_msg = "job_name and job_id"
+    elif job_id is None:
+        job_not_found_msg = "with the name {0}".format(job_name.upper())
+    elif job_id != '*' and job_name != '*':
```
surendrababuravella: Are you able to test and produce these error messages?

plugins/module_utils/job.py
```diff
@@ -344,23 +349,20 @@ def _get_job_status(job_id="*", owner="*", job_name="*", dd_name=None, sysin=Fal
     final_entries = []
 
     # In ZOAU>= 1.3.0, include_extended has to be set to true so we get the program name for a job.
-    # Observation shows the job_name parameter is not being used, so we will drop that
-
-    # expanding > 1.3.0 of zoau, to include all params
-    entries = jobs.fetch_multiple(job_id=job_id_temp, job_owner=owner, include_extended=True)
+    entries = jobs.fetch_multiple(job_id=job_id_temp, job_name=job_name, job_owner=owner, include_extended=True)
```
surendrababuravella: zoautil_py.exceptions.JobFetchException will be thrown by this code when there are no matching jobs as per search criteria. We are not catching exception here. So '_job_not_found' will never executed.

plugins/modules/zos_job_query.py
```diff
@@ -438,23 +441,25 @@ def query_jobs(job_name, job_id, owner):
...
-        jobs = job_status(owner=owner, job_name=job_name)
-    else:
-        jobs = job_status(job_name=job_name)
-    if not jobs:
-        raise RuntimeError("List FAILED! no such job was found.")
+
+    try:
+        # Owner defaults to current user if none is specified
+        if owner is None:
+            current_user = os.environ.get('USER') or os.environ.get('LOGNAME')
```
surendrababuravella: there are possibilities that ENV vars might not be available or might have set with custom values instead of exact USER. Then current_user will be 'None' or some custom value. We should add this limitaion in Documentation. 

We can use below code which will do better job in fetching USER name.
```
import getpass
current_user = getpass.getuser()
```

hmillersarah: Updated default owner functionality and documentation

plugins/module_utils/job.py
hmillersarah: Updated to catch exception so _job_not_found will execute

plugins/module_utils/job.py
hmillersarah: Missed re-verifying these checks, but have now tested and produced error messages from _job_not_found following fix below to catch zoautil_py JobFetchException.

plugins/module_utils/job.py
hmillersarah: Updated

tests/helpers/users.py
```diff
@@ -922,6 +928,74 @@ def _create_user_zos_limited_hlq(self) -> None:
...
+        command.write(f"tsocmd SETROPTS RACLIST\\({saf_class}\\) REFRESH;")
+        command.write(f"echo SETROPTS RC=$?;")
+
+        cmd = f"{command.getvalue()}"
+        results_stdout_lines = self._connect(self._remote_host, self._model_user, cmd)
+
+        try:
+            # Evaluate the results
+            rdefine_rc = [v for v in results_stdout_lines if f"RDEFINE RC=" in v][0].split('=')[1].strip() or None
+            if not rdefine_rc or int(rdefine_rc[0]) > 4:
```
yrana17: The check on rdefine_rc only evaluates the first character via [0]. Because of this, a multi-digit code like 12 would pass as 1 without triggering the try/except block. My suggestion is to evaluate the whole return code integer (int(rdefine_rc) > 4) to prevent this.
```suggestion
            if not rdefine_rc or int(rdefine_rc) > 4:
```

tests/functional/modules/test_zos_job_query_func.py
```diff
@@ -41,10 +42,37 @@ def get_job_id(hosts, len_id=9):
...
+        Connection to host machine
+    """
+    results = hosts.all.shell(cmd="jls")
+    for result in results.contacted.values():
+        all_jobs = result.get("stdout_lines")
+    for job_n_info in all_jobs:
+        job = job_n_info.split()
+        return job
+
+def create_temp_dir(hosts, data_set_name, temp_path):
```
yrana17: Since this function handles creating both the directory and the PDS member, the current name create_directory is a bit too narrow. can we rename it to something more descriptive.

tests/helpers/users.py
```diff
@@ -922,6 +928,74 @@ def _create_user_zos_limited_hlq(self) -> None:
...
+                err_details = f"rdefine {saf_class} {saf_profile}"
+                err_msg = f"Unable to {err_details} for managed user [{self._managed_racf_user}], review output {results_stdout_lines}."
+                raise Exception(err_msg)
+
+            permit_rc = [v for v in results_stdout_lines if f"PERMIT RC=" in v][0].split('=')[1].strip() or None
+            if not permit_rc or int(permit_rc[0]) > 4:
```
yrana17: Same as the rdefine_rc check—we should evaluate the whole integer here to make sure multi-digit return codes aren't missed by just checking the first character.

tests/helpers/users.py
```diff
@@ -922,6 +928,74 @@ def _create_user_zos_limited_hlq(self) -> None:
...
+                err_details = f"permit {saf_profile} class {saf_class}"
+                err_msg = f"Unable to {err_details} for managed user [{self._managed_racf_user}], review output {results_stdout_lines}."
+                raise Exception(err_msg)
+
+            setropts_rc = [v for v in results_stdout_lines if f"SETROPTS RC=" in v][0].split('=')[1].strip() or None
+            if not setropts_rc or int(setropts_rc[0]) > 4:
```
yrana17: Same as the rdefine_rc check

yrana17 (approved):
Thanks for incorporating suggestions. 
LGTM.

ketankelkar (approved):
Users querying by name and owner will no longer get back jobs they didn't ask for or risk hitting RACF violations, thanks to tighter defaults (None instead of "*") and all three parameters now being forwarded together to ZOAU. Thanks for the thorough test coverage across wildcard combinations, exact ID queries, and restricted-user scenarios. 🚀

